PR #80, opened 2026-04-11
fix(session): persist search results in context delta

lifeodyssey:
Card 1 of BUG-03 fix.

## Summary by CodeRabbit

* **New Features**
  * Session context now preserves the most recent successful search result (last search data) and includes it when reconstructing context; presence of this data can prevent an otherwise early-exit. Malformed or missing search data is ignored.

* **Tests**
  * Added unit tests covering extraction, inclusion, omission, and reconstruction behaviors for last-search data in session context.

### DIFF
--- a/backend/interfaces/session_facade.py
+++ b/backend/interfaces/session_facade.py
@@ -104,14 +104,15 @@ def build_session_summary(state: dict[str, object]) -> dict[str, object]:
     }
 
 
-def build_context_block(
-    session_state: dict[str, object],
-    user_memory: dict[str, object] | None = None,
-) -> dict[str, object] | None:
-    """Derive a context block from session history and cross-session memory."""
-    raw_interactions = session_state.get("interactions")
-    interactions = raw_interactions if isinstance(raw_interactions, list) else []
-    summary = as_str_or_none(session_state.get("summary"))
+def _extract_from_interactions(
+    interactions: list[object],
+) -> tuple[str | None, str | None, str | None, dict[str, object] | None, list[str]]:
+    """Walk interactions in reverse and extract context fields.
+
+    Returns:
+        (current_bangumi_id, current_anime_title, last_location,
+         last_search_data, visited_bangumi_ids)
+    """
     current_bangumi_id: str | None = None
     current_anime_title: str | None = None
     last_location: str | None = None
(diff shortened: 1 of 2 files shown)


lifeodyssey (commented):
## Linus Code Review — PR #80

### Verdict: request_changes

---

### Findings

**P1 — Correctness gap: last_search_data silently dropped when bangumi_id resolution fails**

`backend/interfaces/session_facade.py`, line ~161 (early-return guard in `build_context_block`)

The early-return guard returns `None` when `current_bangumi_id`, `last_location`, `visited_bangumi_ids`, and `summary` are all falsy. If a session has only a `search_bangumi` step where `resolve_anime` did not run (or failed), `last_search_data` can be populated while all four guard conditions evaluate to false. In that case `build_context_block` returns `None` and `last_search_data` is silently dropped with no signal to the caller. No test covers this edge case. Add a test: search_bangumi succeeds with data but no bangumi_id resolved — verify `build_context_block` returns a block containing `last_search_data` rather than `None`. The guard condition needs `or last_search_data is not None` to make the early return safe.

**P1 — Method length: build_context_block exceeds 10-line rule**

`backend/interfaces/session_facade.py`, line 107

The function body is approximately 60 effective lines before this PR and grows by 5 more lines here. The 1-10-50 rule caps methods at 10 lines. This was pre-existing, but this PR adds to it without addressing it. Suggest extracting the interaction-walk loop body into `_extract_from_interactions(interactions) -> tuple[...]` and the user_memory walk into `_merge_user_memory(user_memory, ...) -> tuple[...]`.

**P1 — Method length: extract_context_delta exceeds 10-line rule**

`backend/interfaces/session_facade.py`, line 179

Similarly approximately 45 lines pre-PR, growing by 4 lines. Pre-existing, worsened by this PR. Suggest extracting the two loops into `_extract_from_resolve_anime(step_results)` and `_extract_from_search_bangumi(plan, step_results)`.

**P2 — Variable declared away from usage point**

`backend/interfaces/session_facade.py`, line ~118 (new `last_search_data` variable in diff)

`last_search_data: dict[str, object] | None = None` is declared in the initialization block alongside unrelated variables, several lines before it is first read or written. Per clean code standards, variables should be declared near their usage point. Move the declaration to just before the loop block where it is populated, or immediately before the `if last_search_data is None:` guard inside the loop.

**P2 — Test data construction via hand-written dicts instead of factory-boy**

`backend/tests/unit/test_session_facade.py`, all new test classes

The testing strategy mandates `factory-boy` (`PipelineResultFactory`, `PlanStepFactory`) instead of scattered hand-written dict/object construction. All new tests construct objects manually via `_make_plan` and inline `StepResult(...)` kwargs. This creates maintenance overhead if model fields change. Migrate to factories per `docs/testing-strategy.md`.

**P2 — TDD order not verifiable from diff**

The implementation diff appears before the test diff in the PR. Cannot confirm tests were written first. Not blocking, but flagged per TDD review requirement.

---

### Quality Ratchet

```
ac_total: 5
ac_with_test: 5
missing: []
```

All 5 acceptance criteria have corresponding unit tests in the diff.

| AC | Test |
|----|------|
| search_bangumi success -> last_search_data in delta | test_search_bangumi_success_includes_last_search_data |
| build_context_block reconstructs last_search_data from most recent | test_reconstructs_last_search_data_from_most_recent_interaction + test_uses_most_recent_interaction_with_search_data |
| No search step -> no last_search_data key | test_no_search_bangumi_step_has_no_last_search_data |
| No interactions -> returns None | test_no_interactions_returns_none |
| Malformed last_search_data -> ignored gracefully | test_malformed_last_search_data_ignored_gracefully |

---

### What is clean

- No `Any` type introduced. All new annotations use `dict[str, object]` and `str | None`, consistent with project rules.
- `isinstance(raw_search, dict)` guard for malformed input is correct defensive narrowing.
- Conditional append `if last_search_data is not None: block["last_search_data"] = last_search_data` avoids polluting the context block with `None` keys — correct pattern.
- Docstring on `extract_context_delta` updated to reflect new behavior.
- 27 unit tests pass with no failures.

PR #81, opened 2026-04-11
fix(pipeline): forward clarify events to SSE stream

lifeodyssey:
Card 4 of BUG-03 fix.

## Summary by CodeRabbit

* **New Features**
  * Pipeline now detects and handles clarification requests during agent runs; intent set to "clarify" and final status to "needs_clarification" when triggered
  * Callbacks invoked for clarification events, providing question and option details

* **Tests**
  * Added tests validating clarify event emission, callback behavior, and resulting pipeline intent/status
  * Ensured pipeline handles clarify events without callbacks present

* **Chores**
  * Added development/testing tooling to project configuration

### DIFF
--- a/backend/agents/pipeline.py
+++ b/backend/agents/pipeline.py
@@ -20,6 +20,31 @@ logger = structlog.get_logger(__name__)
 
 MAX_REACT_STEPS = 8
 
+# Intent priority: lower number = higher precedence (route > search > fallback).
+_INTENT_PRIORITY: dict[str, int] = {
+    "plan_route": 0,
+    "plan_selected": 1,
+    "search_nearby": 2,
+    "search_bangumi": 3,
+    "answer_question": 4,
+    "clarify": 5,
+    "greet_user": 6,
+    "resolve_anime": 7,
+}
+
+
+def _infer_intent(step_results: list[StepResult], priority_map: dict[str, int]) -> str:
+    """Return the highest-priority tool name from successful step results."""
+    intent = "answer_question"
+    best_priority = priority_map.get("answer_question", 99)
+    for sr in step_results:
+        if sr.success:
+            p = priority_map.get(sr.tool, 99)
+            if p < best_priority:
+                intent = sr.tool
+                best_priority = p
+    return intent
+
 
 @dataclass
 class ReactStepEvent:
@@ -204,14 +229,17 @@ async def run_pipeline(
                 event.tool, event.status, event.data, event.thought, event.observation
             )
 
-        if on_step is not None and event.type == "clarify":
-            await on_step(
-                "clarify",
-                "needs_clarification",
-                event.data,
-                event.thought,
-                event.message,
-            )
+        if event.type == "clarify":
+            clarify_fired = True
+            final_message = event.message
+            if on_step is not None:
+                await on_step(
+                    "clarify",
+                    "needs_clarification",
+                    event.data,
+                    event.thought,
+                    event.message,
+                )
 
         if event.step_result is not None:
             all_step_results.append(event.step_result)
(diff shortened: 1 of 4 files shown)


lifeodyssey (commented):
REVIEW FINDINGS

## Test Run

The tests in the PR branch were not available in the worktree (current branch is `main`, which has the pre-PR version of `test_pipeline_react.py` with 143 lines). The 3 pre-existing tests passed. Coverage failure (24% < 70%) is a pre-existing repo-wide issue, not introduced by this PR.

---

## SOLID / 1-10-50 Analysis

### pipeline.py — `run_pipeline`

**P1 — Single Responsibility / 1-10-50 violation (non-blocking but notable)**

The `clarify_fired` branch duplicates the status-setting logic that already follows it. Specifically:

```python
# Block A — on_step call for clarify (lines ~207-215)
if on_step is not None and event.type == "clarify":
    await on_step(...)

# Block B — clarify_fired flag set (lines ~222-224)
if event.type == "clarify":
    clarify_fired = True
    final_message = event.message
```

These two `if event.type == "clarify"` checks are separate, causing two scans of the same condition in the same loop body. They should be collapsed into one branch. This is a readability and 1-10-50 concern: the loop body now has 4 independent `if` blocks on `event.type`, making it harder to reason about order of operations.

**P2 — Else usage (minor)**

```python
if clarify_fired:
    intent = "clarify"
else:
    intent = "answer_question"
    best_priority = ...
    for sr in all_step_results:
        ...
```

The else block is long (6 lines). The review standard says "early return, no else." The equivalent early-assignment pattern would be cleaner:

```python
intent = "clarify" if clarify_fired else _infer_intent(all_step_results, _INTENT_PRIORITY)
```

or extract `_infer_intent` as a private helper. As-is, the else block is acceptable functionally but violates the style guideline.

**P2 — `_INTENT_PRIORITY` dict defined inside function body**

`_INTENT_PRIORITY` is a module-level constant being rebuilt on every `run_pipeline` call. This was pre-existing, but the PR extends the dict with a `"clarify": 5` entry. It should be a module-level `_INTENT_PRIORITY` constant. Not introduced by this PR but the PR touches this dict directly.

---

## Naming Convention

All new symbols follow verb-first or noun conventions correctly:
- `clarify_fired` — acceptable boolean flag name
- `_fake_react_loop_clarify` / `_fake_react_loop_done` — verb-first, clear intent
- `mock_db` fixture — acceptable

No naming violations.

---

## No `Any` Type Check

The new test helper signatures use `**kwargs: object` — correct. No `Any` introduced. Pass.

---

## Quality Ratchet — AC Coverage

| AC | Description | Test in diff | Test name |
|----|-------------|--------------|-----------|
| AC1 | react_loop yields clarify event → on_step called with tool="clarify" + question/options data | YES | `test_run_pipeline_clarify_calls_on_step_with_clarify_tool` |
| AC2 | PipelineResult for clarify has intent="clarify", status="needs_clarification" | YES | `test_run_pipeline_clarify_sets_intent_and_status` |
| AC3 | No clarify event → pipeline unchanged | YES | `test_run_pipeline_no_clarify_unchanged` |
| AC4 | on_step is None + clarify fires → no crash | YES | `test_run_pipeline_clarify_no_on_step_no_crash` |

Additionally, `test_react_loop_clarify_yields_clarify_event` covers the react_loop layer specifically.

**All 4 ACs have corresponding tests. Quality ratchet: 4/4.**

---

## Clarify Handling — Pipeline Flow Correctness

The PR correctly guards the existing pipeline:
- Non-clarify paths: `clarify_fired` remains `False`, intent inference and status logic unchanged.
- Clarify path: `intent="clarify"`, `status="needs_clarification"`, `on_step` called before `step_result` collection (which is correct since clarify returns before appending a step_result).

One subtle concern worth flagging:

**P1 — `on_step` called for clarify BEFORE `step_result` is collected, but `success` field in `final_output` reads `all_step_results[-1].success`**

When clarify fires, `all_step_results` is empty (clarify returns immediately from react_loop). So `final_output["success"]` will be `False` (`bool([] and ...)` = `False`). This is arguably correct behaviour — there is no successful data step — but it is implicit and undocumented. A comment would prevent future confusion.

**P2 — `_fake_react_loop_clarify` / `_fake_react_loop_done` are module-level async generators used as `side_effect` on a `MagicMock`**

The `MagicMock(side_effect=_fake_react_loop_clarify)` pattern is correct for replacing an async generator factory. This is non-obvious but valid.

---

## pyproject.toml change

Adding `pytest`, `pytest-asyncio`, `pytest-cov`, `ruff` to `[project.optional-dependencies] dev` is a correctness improvement — these were likely already installed transitively. No concerns.

---

## Summary

The implementation is functionally correct and all 4 ACs are tested. The two P1 findings are style/maintainability issues that do not affect correctness but should be addressed before merge:

1. Collapse the two separate `if event.type == "clarify"` checks in the loop body into one.
2. Extract the intent-inference else-block into a helper or restructure to eliminate the `else`.

PR #192, opened 2026-06-23
fix(ci): atlas-action subaction path + pipeline filter

lifeodyssey:
CD staging first-run found: root atlas-action is a Docker action; SHA-pinning it fails because its Dockerfile COPYs a prebuilt binary absent from the git tree. Switch to the migrate/apply subaction (node24, no Docker build). cd-staging affected filter now also covers _deploy-component.yml + cd-staging.yml so pipeline fixes re-trigger catalog.

Generated with Claude Code

## Summary by CodeRabbit

* **Chores**
  * Optimized internal deployment workflow configuration to streamline component deployment and catalog update processes.


lifeodyssey:
lgtm

PR #328, opened 2026-07-13
feat(edge): local JWT verification + Neon Auth dual-issuer readiness (no cutover)

lifeodyssey:
## What

Edge auth infrastructure slice (F9 + Neon dual-issuer readiness, **no cutover**):

1. **Supabase JWT verification is now local.** `worker/auth.ts` verifies user JWTs against the project JWKS (`jose` `createRemoteJWKSet`, cached per isolate, refetch-on-unknown-kid) instead of a network round-trip to `${SUPABASE_URL}/auth/v1/user` on every authed `/v1/*` request.
2. **Neon Auth (Better Auth) issuer path added, default OFF.** Flag-gated EdDSA verification with kid/iss/aud/exp checks. Tokens are routed by characteristics (header `alg` / `iss` claim): Neon tokens verify against the Neon JWKS, Supabase tokens against the Supabase JWKS — both issuers coexist while the flag is on.
3. `sk_` API-key path: untouched (byte-identical).

## F9 premise verification (done live before implementation, values not committed)

The F9 assumption "Supabase supports asymmetric JWT" was marked [unverified]. Verified against the production project:

- `GET $SUPABASE_URL/auth/v1/.well-known/jwks.json` → HTTP 200 with one active **ES256 (P-256 EC)** signing key.
- A freshly minted user access token has header `{"alg":"ES256","kid":"<matches the JWKS key>"}` and payload `iss = $SUPABASE_URL/auth/v1`, `aud = "authenticated"`, `sub = <user UUID>` (identical to the `id` the old `/auth/v1/user` endpoint returned), ~1h expiry.

**Conclusion: the project signs with asymmetric ES256 today → Plan 1 (local JWKS verification) implemented.** The HS256-shared-secret fallback plan was not needed. HS256 is deliberately excluded from the accepted algorithms (key-confusion guard); legacy HS256 service/anon API keys were never valid on this path.

## Default behavior: zero change

- With today's production env (no `NEON_AUTH_*` vars set), the Neon path is unreachable code — flag defaults to off in code, no config change needed to merge safely.
- `/v1/*` accept/reject semantics, 401 body, and forwarded `X-User-Id`/`X-User-Type` header contract are unchanged. `userId` comes from the token `sub`, which equals the previous `/auth/v1/user` `id`.
- One known tradeoff to disclose (inherent to local verification, industry standard for edge JWT checks): server-side revocation (user deleted/banned, session signed out) is no longer caught mid-lifetime; a revoked token stays valid up to its remaining `exp` (≤1h). Previously the per-request introspection caught it immediately.

## Design notes

- Routing rule: a non-`sk_` bearer token goes to the Neon verifier iff `NEON_AUTH_ENABLED === "true"` AND `NEON_AUTH_JWKS_URL` AND `NEON_AUTH_ISSUER` are set AND (unverified header `alg === "EdDSA"` OR unverified `iss === NEON_AUTH_ISSUER`). Everything else goes to the Supabase verifier. Cross-verification is impossible by construction (per-issuer algorithm allowlists + separate JWKS).
- Supabase verify: `iss = ${SUPABASE_URL}/auth/v1`, `aud = "authenticated"`, algorithms `["ES256","RS256"]`.
- Neon verify: `iss = aud = NEON_AUTH_ISSUER`, algorithms `["EdDSA"]` (kid-resolved via JWKS; 15-min tokens covered by the standard exp check). Shape per `docs/ops/auth-migration-neon.md` §5.
- JWKS fetches go through the existing injected-fetch seam (`jose` `customFetch`), so tests stub the JWKS endpoints with locally generated keypairs — zero real network in the suite.
- `AuthEnv` no longer declares `SUPABASE_ANON_KEY` (only the removed introspection call used it). The Worker secret itself is NOT removed — other consumers (frontend) still use it.

## Tests (16 → 30, all green)

| File | Cases | Covers |
|---|---|---|
| `worker/entry.test.ts` | 11 (unchanged) | routing/header-stripping regression |
| `worker/auth.test.ts` | 9 | no-header, sk_ valid/unknown (regression), valid ES256 local verify (asserts JWKS fetched and `/auth/v1/user` NOT called), expired, wrong iss, wrong signing key, HS256 rejected, garbage token |
| `worker/auth-neon.test.ts` | 10 (new) | flag absent/false (Neon JWKS never fetched), valid EdDSA accept, wrong iss, wrong aud, expired, wrong key, dual-issuer coexistence (each token verified only against its own JWKS), missing JWKS URL no-throw, sk_ regression with flag on |

All token/JWKS fixtures are locally generated per test (`generateKeyPair`/`SignJWT`), unique fake `.example.test` hosts per case, zero real hostnames/ids/secrets anywhere.

## Gates

- `pnpm run test:worker`: **30/30 pass** (Node 24).
- `tsc --noEmit --strict` (es2022 / bundler resolution / workers-types + node): **0 errors**.
- Suppression/`any` grep across changed files: clean (no `any`, no `eslint-disable`, no `@ts-*` suppressions).
- eslint: the repo defines no lint surface for `worker/` (frontend flat config ignores files outside its base path; no typescript-eslint in the workspace). CI parity gate for this directory is `test:worker`.

## Integration notes (for the wrangler.toml owner — no toml change required to merge)

This PR intentionally does not touch `wrangler.toml`. The code treats missing vars as "Neon path off", so merging is safe with zero config changes. At cutover (S0.6/S0.8) add to the Worker env (values operator-side only, never committed):

- `NEON_AUTH_ENABLED = "true"` — flips the dual-issuer routing on
- `NEON_AUTH_JWKS_URL` — `<neon auth base URL>/.well-known/jwks.json`
- `NEON_AUTH_ISSUER` — the Neon Auth base URL (checked as both `iss` and `aud`)

Dependency changes at workspace root: `jose` ^6.2.3 (runtime), `@cloudflare/workers-types` + `@types/node` (dev, for the tsc gate).

## Deferred to cutover (S0.6/S0.8 — out of scope here)

- Set the three `NEON_AUTH_*` vars and flip the flag.
- apps/web login UI on the Better Auth client SDK; retire `scripts/qa_auth.py` / `qa_login.sh` with the Supabase login path.
- Supabase retirement checklist in `docs/ops/auth-migration-neon.md` §5.

### DIFF
--- a/docs/ops/cloudflare-hardening.md
+++ b/docs/ops/cloudflare-hardening.md
@@ -36,7 +36,7 @@ Two credential types, both validated at the Worker edge:
 
 | Credential | Format | Validation |
 |---|---|---|
-| Human JWT | `Bearer <supabase_jwt>` | `validateJwt()` calls Supabase `/auth/v1/user` with `SUPABASE_ANON_KEY` |
+| Human JWT | `Bearer <supabase_jwt>` | `authenticate()` verifies the signature locally against the Supabase JWKS (ES256/RS256 via jose `createRemoteJWKSet`), checking issuer/audience/exp — no `/auth/v1/user` round-trip. Flag-gated Neon Auth (EdDSA) issuer is off by default. |
 | Agent API key | `Bearer sk_<hex>` | `validateApiKey()` SHA-256 hashes the key, looks up `api_keys` table via Supabase REST with `SUPABASE_SERVICE_ROLE_KEY` |
 
 On success the Worker sets `X-User-Id` and `X-User-Type`, deletes the `Authorization` header, and forwards to the container. The container never sees raw bearer tokens.
@@ -45,9 +45,9 @@ On success the Worker sets `X-User-Id` and `X-User-Type`, deletes the `Authoriza
 
 | Variable | Boundary | Notes |
 |---|---|---|
-| `SUPABASE_URL` | Worker-only | Used for JWT validation and API-key lookup |
-| `SUPABASE_ANON_KEY` | Worker-only | Passed as `apikey` header to Supabase auth |
+| `SUPABASE_URL` | Worker-only | JWKS fetch (`/auth/v1/.well-known/jwks.json`) for local JWT verification + `api_keys` lookup |
 | `SUPABASE_SERVICE_ROLE_KEY` | Worker-only | Used for `api_keys` table lookup |
+| `NEON_AUTH_ENABLED` / `NEON_AUTH_JWKS_URL` / `NEON_AUTH_ISSUER` | Worker-only (optional) | Dual-issuer readiness — Neon Auth EdDSA JWKS verification; absent or `false` ⇒ Neon path off (default) |
 | `SUPABASE_DB_URL` | Container-only | Direct Postgres connection for asyncpg |
 | `GEMINI_API_KEY` | Container-only | LLM provider credential |
 | `ANITABI_API_URL` | Container-only | Pilgrimage data API |
@@ -62,7 +62,7 @@ The full container env allowlist is defined in `worker/worker.js` as `CONTAINER_
 ## Current Trust Boundary
 
 - Browser and API clients talk only to the Worker hostname
-- Worker-only auth secrets stay at the edge: `SUPABASE_URL`, `SUPABASE_ANON_KEY`, `SUPABASE_SERVICE_ROLE_KEY`
+- Worker-only auth secrets stay at the edge: `SUPABASE_URL`, `SUPABASE_SERVICE_ROLE_KEY` (+ optional `NEON_AUTH_*`); the edge JWT path verifies against the public Supabase JWKS, so no `SUPABASE_ANON_KEY` is needed there
 - Container runtime receives only its explicit allowlist from `worker/worker.js`
 - Backend auth trust starts from `X-User-Id` and `X-User-Type`, not from raw bearer tokens
 
--- a/docs/ops/deployment.md
+++ b/docs/ops/deployment.md
@@ -42,18 +42,19 @@ The deployment target stays intentionally thin. The Worker owns routing and edge
 | Layer | Responsibility | Secrets/config it should see |
 |---|---|---|
 | Frontend build | Static export only | `NEXT_PUBLIC_SUPABASE_URL`, `NEXT_PUBLIC_SUPABASE_ANON_KEY` |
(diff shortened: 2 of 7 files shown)


lifeodyssey:
## Review changes addressed (commit b2de1e7) — verification logic untouched

`worker/auth.ts` is byte-identical to the approved version (`git diff HEAD~1 -- worker/auth.ts` empty). Only tests + docs changed.

### 1. Fake HS256 test → real (`worker/auth.test.ts`)
The old HS256 case stubbed the JWKS endpoint with HTTP 500, so it rejected for network reasons and would pass even with the algorithm allowlist deleted. Now it serves a **valid EC JWKS** (kid-matched), so the rejection is attributable to the guard, not a fetch failure.

Caveat worth recording: an HS256-via-JWKS test can never be made to go red on allowlist removal — jose's `createRemoteJWKSet` structurally refuses symmetric keys (`JWKSNoMatchingKey`), so HS256 is rejected regardless of the `algorithms` option. That's a stronger guarantee than the allowlist, but it means the HS256 case alone doesn't *prove* the allowlist is load-bearing. So I added a test that does:

### 1b. New: EdDSA algorithm-confusion test that genuinely guards the allowlist
`"EdDSA token is rejected by the Supabase algorithm allowlist"` — an EdDSA token whose signature **is** verifiable against a served OKP JWKS, rejected **only** because EdDSA isn't in `["ES256","RS256"]`. Empirically confirmed: deleting `algorithms: ["ES256","RS256"]` from `verifySupabase` turns this test **red** (jose then resolves the OKP key and accepts the token). Restored immediately; `auth.ts` unchanged in the diff.

### 2. Missing Supabase wrong-audience negative → added
`"wrong Supabase audience is rejected"` (valid ES256 signature + correct issuer, `aud` wrong → `{ok:false}`). Empirically confirmed: deleting `audience: "authenticated"` from `verifySupabase` turns this test **red**. `esFixture` gained an `aud` param (default `"authenticated"`), so existing call sites are unaffected.

Worker tests: **30 → 32, all green.** `tsc --strict` clean.

### 3. Doc drift → synced
`docs/ops/deployment.md` and `docs/ops/cloudflare-hardening.md` no longer describe the edge JWT path as `validateJwt()` calling `/auth/v1/user` with `SUPABASE_ANON_KEY`. Both now document local `createRemoteJWKSet` verification (ES256/RS256, iss/aud/exp, no round-trip), the Neon dual-issuer flag, and that `SUPABASE_ANON_KEY` is no longer consumed at the edge. File-name reference corrected `worker/worker.js` → `worker/auth.ts` in the auth-flow section.

Empirical-proof runs were throwaway (temp-edit `auth.ts` → run → `git checkout worker/auth.ts`); final `git diff HEAD -- worker/auth.ts` is empty.

Issue #289, opened 2026-07-07
S2.8 — workers/users service + route save/list oRPC enabler


lifeodyssey:
## Releasable statement
Stands up a new `workers/users` TypeScript service (mirroring the existing `workers/catalog` pattern) so users can save a route from chat and see it later in My Route, with anonymous-session routes automatically reattached to their account after login. Route data lands in Neon via oRPC, with `apps/web`'s Supabase client used only to obtain an auth token.

## AC summary
- AC count: 6
- Test type distribution: unit=2, integration=4, browser=0, eval=0, api=0, manual/ops=0

## Dependencies
S1.7 (the moment that triggers a save); S0.9's migrations documentation (reference for the schema-migration tool chain).

## Spec source
`docs/superpowers/specs/2026-07-06-frontend-rebuild/iter-2.md` (commit 6fc13b8)

### S2.8 `workers/users` service build-out + route save/list oRPC enabler (SD-2 final)

**User story**: As a user, I want to be able to save a route from chat and then see it in マイルート afterward, and I want the results of my anonymous actions to automatically attach to my account once I log in.

**Design basis**: no visual mockup; `user-journey.md`: "anonymous-state routes are held against the anonymous session and reattached to the account after login."

**Backend enabler (SD-2/SD-3② final decision, not provisional)**:
- Build a new TS service `workers/users` (fully mirroring `workers/catalog`'s existing pattern: Hono + oRPC + Drizzle — do not invent a new pattern).
- On the Neon side, create the table via the SD-1 tool chain (Drizzle schema → atlas-provider-drizzle → `atlas migrate`): a `routes` table moves into Neon (with `user_id`, `title`, `status TEXT CHECK (status IN ('draft','saved','completed'))`, `saved_at`, `updated_at`).
- oRPC routes at `/v1/users/routes` (list/save/delete), with the contract going into `packages/contract` (a new file, e.g. `src/users-contract.ts`).
- Auth: requests carry the Supabase-issued JWT as a bearer token; `workers/users` validates it and takes `sub` to resolve the Neon-side `user_id`; `apps/web`'s `supabase-js` is used only to obtain this JWT (`getSession()`) — it does not query any data table.
- Anonymous route claiming: after login, `apps/web` calls `/v1/users/routes/claim` (or an equivalent endpoint), and the server reassigns the `user_id` of routes tied to the anonymous session to the logged-in user.
[...]

PR #330, opened 2026-07-13
feat(ci): per-PR preview environments — CF preview URLs + Neon branch-per-PR + auto-teardown

lifeodyssey:
## Summary

Per-PR preview environments v2 (Refs #305): label a PR `preview` → CI creates a Neon branch `preview/pr-<N>` (atlas-migrated), uploads **routeless** Cloudflare Worker versions for `workers/catalog` + `apps/web`, and posts one self-updating PR comment with stable preview URLs. Closing/merging the PR deletes the Neon branch automatically.

New standalone workflow `.github/workflows/preview.yml` — `ci.yml` / `deploy.yml` / `_deploy-component.yml` (#324) untouched.

## Design

- **Trigger / opt-in gate**: `pull_request: [labeled, synchronize, reopened, closed]`; the preview job runs only when the PR carries the `preview` label AND is same-repo (fork guard). `opened` is deliberately not in the trigger list — unlabeled PRs never start a run.
- **Isolation (why a dedicated `env.preview`)**: top-level worker names equal production (`catalog`, `animichi-web`). A `versions secret put` against those would be inherited by the next real deploy — preview therefore gets dedicated routeless workers `catalog-preview` / `animichi-web-preview` (`workers_dev = false`, `preview_urls = true`), so staging/production secret chains are never touched.
- **Per-version DB wiring**: `versions secret put DATABASE_URL` (Neon pooled URL via stdin) then `versions upload --preview-alias pr-<N>` — the uploaded version snapshots the secret; the alias gives a stable URL across pushes. Cross-PR races on secret inheritance are prevented by a global job concurrency group (`preview-cf-upload`, queued not cancelled).
- **Idempotency**: Neon `create-branch-action` reuses an existing `preview/pr-<N>`; atlas applies only pending migrations; the PR comment is upserted by hidden marker (`<!-- animichi-preview-status -->`); re-running any event converges.
- **Teardown**: on `closed` (merged or not, labeled or not — always sweeps), with **name-guard assertions**: branch must match `preview/pr-<digits>` (numeric suffix re-checked) plus an explicit denylist (`main/master/production/staging/development`) before any delete; existence is verified against the Neon API first (API failures fail the job loudly instead of silently skipping the delete).
- **Secret hygiene**: connection strings exist only in step `env` + stdin, are `add-mask`ed defensively, and never appear in logs, files, or artifacts. `NEON_PROJECT_ID` is read from a repo **variable** (fail-fast preflight if unset) — nothing hardcoded.
[...]

PR #331, opened 2026-07-13
S2.8: workers/users service — route save/list oRPC + Neon Auth JWT (enabler)

lifeodyssey:
S2.8 enabler: stands up the `workers/users` TypeScript service (mirroring `workers/catalog`) with route **save/list** oRPC + **Neon Auth JWT verification**, wired end-to-end: contract → Neon migration → service → edge routing → CI/deploy lanes.

Refs #289

## AC coverage

| AC (issue #289) | Type | Status | Test |
|---|---|---|---|
| Save → visible in list via `/v1/users/routes` | integration | ✅ (worker-wire level, injected DB) | `workers/users/test/worker.worker.test.ts` "saves then lists a route for the same JWT subject" |
| Empty list returns `[]` without erroring | unit | ✅ | `workers/users/test/routes-api.worker.test.ts` "lists an empty store" |
| Cross-user access rejected 403 by the service (not silently at Neon) | integration | ✅ | wire: "serializes the defined ownership error for a cross-user update" + handler: "throws ROUTE_NOT_OWNED" |
| Anonymous-session claim endpoint | integration | ⏳ deferred (see Leftovers) | — |
| Auth boundary: no valid JWT → flat 401, never anonymous | unit | ✅ | `workers/users/test/auth.worker.test.ts` (10 cases: missing/empty/garbage/expired/wrong-iss/wrong-aud/foreign-key/valid/healthz/unconfigured) |
| CI/deploy wiring: `ci-users` runs on `workers/users/**`; deploy step ordered before the root Worker | integration | ✅ (this PR's CI run exercises `ci-users`) | `.github/workflows/ci.yml` + `deploy.yml` |

Quality ratchet: every shipped AC has a test in this diff (`ac_total == ac_with_test` for the 5 in-scope ACs).

## What's in here

- **`packages/contract/src/users-contract.ts`** (new; catalog surface untouched): `RouteStatus` / `UserRoute` / `SaveRouteInput` / `ListRoutesResult`, users error registry (`ROUTE_NOT_FOUND` 404, `ROUTE_NOT_OWNED` 403) + `pickUsersErrors`, oRPC `GET|POST /v1/users/routes`. `emit-openapi` now also emits a committed, drift-guarded `users-openapi.json`; `openapi.json` (catalog) is byte-identical.
- **`db/migrations/20260713000001_user_routes.sql`**: evolves the existing `routes` table (old Python-agent shape) — adds `user_id`, `title`, `status` (`draft|saved|completed` CHECK), `saved_at`, `updated_at` (+trigger, +index). `session_id` retained on purpose: it is exactly what the future claim flow needs. `atlas.sum` regenerated (atlas v0.30.0, `migrate validate` clean).
- **`workers/users/`**: Hono + oRPC + Drizzle-typing-only/raw-`sql`-queries, per catalog. JWT: jose `createRemoteJWKSet` against `env.NEON_AUTH_JWKS_URL` (cached per URL), EdDSA-only, `iss` == `aud` == JWKS base; failure → flat 401 with the edge's unauthorized envelope. `[env.staging]`=`users-staging`, `[env.production]`=`users`, routeless (service-binding only). 23 workerd tests; coverage 78/70/86/80 vs 60/60/60/50 floors.
- **Edge**: `worker/app.ts` routes `/v1/users/*` to the `USERS` service binding registered *before* the generic `/v1/*` handler; Authorization passes through untouched (users verifies itself — deliberately different trust model from the container path). Root `wrangler.toml` gains the `USERS` binding for top-level/production/staging.
- **CI/deploy**: `ci-users` lane (`_ts-ci.yml`), users deploy jobs serialized **after catalog** (shared atlas dir never applied concurrently) and **before the root Worker** in the manual `deploy.yml`. `_deploy-component.yml` gains an optional `NEON_AUTH_JWKS_URL` secret (additive; existing callers unaffected).

## Review-caught bug worth knowing about

Interpolating a JS array straight into drizzle's `sql` template expands it into a tuple — `VALUES ($1, $2, ($3, $4), $5)`, and an empty array renders invalid `()`. Against real Postgres the INSERT/UPDATE would fail. Fixed by binding arrays as one parameter: `${sql.param(arr)}::text[]` (documented in `workers/users/AGENTS.md`).

## 🚨 OWNER ACTION required before this reaches main

**Set the GitHub Actions repo secret `NEON_AUTH_JWKS_URL`** (value = `$NEON_AUTH_BASE_URL/.well-known/jwks.json`, see `docs/ops/auth-migration-neon.md`). Without it, `deploy-users-staging` will fail on the first main push. `NEON_DATABASE_URL` already exists (catalog uses it).

## Coordination

This PR **owns root `wrangler.toml` + `worker/app.ts`/`entry.ts` edits for this batch** (per lead assignment); `worker/auth.ts` is deliberately untouched (owned by the auth-cutover lead). If a `[vars]`/binding conflict lands, this PR's version of those files wins.

## Gates (all run locally, Node 24)

| Gate | Result |
|---|---|
| `workers/users`: tsc / eslint(strict, 0 warn) / `test:worker` | ✅ / ✅ / 23 passed, thresholds met |
| `workers/catalog`: tsc / eslint / `test:worker` (regression) | ✅ / ✅ / 109 passed |
| root `pnpm run test:worker` (edge) | 18 passed (2 new `/v1/users/*` cases) |
| `packages/contract`: tsc / `emit:openapi` idempotent (2 runs, no diff; catalog spec byte-identical) | ✅ |
| `actionlint` (all workflows) | ✅ |
| `atlas migrate validate` | ✅ |

## Leftovers (follow-up stories)

- **Claim endpoint** (`/v1/users/routes/claim`) + anonymous-session write path: deferred with the S1.7 save-trigger + apps/web login integration; the schema (retained `session_id`) and the auth middleware are already claim-ready.
- **Delete** route oRPC (in the spec's list/save/delete trio) — not needed by the current UI slice.
- **`apps/web` client wrappers** (`src/lib/data/routes.ts`, `claimAnonymousRoutes.ts`) — land with the web slice that consumes them.
- **Real-DB spike coverage**: neon-http array/timestamptz serialization is exercised only via fakes here; a `pg`-driver spike (catalog-style) should verify against real Postgres before further build-out.
- **Agent-side consumption**: the Python agent does not (and per SD-2 must not) gain data endpoints; if the agent ever needs saved-route context it consumes this service's contract in a later story.

### DIFF
--- a/.github/workflows/_deploy-component.yml
+++ b/.github/workflows/_deploy-component.yml
@@ -27,6 +27,8 @@ on:
         required: false
       R2_SECRET_ACCESS_KEY:
         required: false
+      NEON_AUTH_JWKS_URL:
+        required: false
 jobs:
   deploy:
     name: Deploy
@@ -79,6 +81,7 @@ jobs:
           secrets: ${{ inputs.worker_secrets }}
         env:
           DATABASE_URL: ${{ secrets.NEON_DATABASE_URL }}
+          NEON_AUTH_JWKS_URL: ${{ secrets.NEON_AUTH_JWKS_URL }}
       - name: Smoke
         env:
           COMPONENT: ${{ inputs.component }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       catalog: ${{ steps.f.outputs.catalog }}
+      users: ${{ steps.f.outputs.users }}
       agent: ${{ steps.f.outputs.agent }}
       frontend: ${{ steps.f.outputs.frontend }}
       web: ${{ steps.f.outputs.web }}
@@ -37,6 +38,7 @@ jobs:
           filters: |
             contract: ['packages/contract/**']
             catalog: ['workers/catalog/**', 'packages/contract/**']
+            users: ['workers/users/**', 'packages/contract/**']
             agent: ['apps/agent/**', 'packages/contract/**']
             frontend: ['frontend/**']
             web: ['apps/web/**']
@@ -62,6 +64,14 @@ jobs:
     with:
       component: catalog
 
+  ci-users:
+    name: Users CI
+    needs: changes
+    if: ${{ github.event_name != 'pull_request' || needs.changes.outputs.users == 'true' }}
+    uses: ./.github/workflows/_ts-ci.yml
+    with:
+      component: users
+
   contract-openapi-drift:
     name: Contract OpenAPI drift
     runs-on: ubuntu-latest
@@ -77,8 +87,8 @@ jobs:
         working-directory: packages/contract
       - name: Verify committed OpenAPI is current
         run: |
-          git add -A packages/contract/openapi.json
-          git diff --cached --exit-code -- packages/contract/openapi.json
+          git add -A packages/contract/openapi.json packages/contract/users-openapi.json
+          git diff --cached --exit-code -- packages/contract/openapi.json packages/contract/users-openapi.json
 
   ci-frontend:
     name: Frontend CI
@@ -237,9 +247,30 @@ jobs:
       CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
     concurrency: deploy-web-staging
 
+  deploy-users-staging:
+    name: Deploy users staging
+    needs: [deploy-staging, ci-users]
+    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+    uses: ./.github/workflows/_deploy-component.yml
+    with:
+      component: users
+      environment: staging
+      pulumi_stack: staging
+      working_directory: workers/users
+      run_pulumi: false
+      worker_secrets: |
+        DATABASE_URL
+        NEON_AUTH_JWKS_URL
+    secrets:
+      NEON_DATABASE_URL: ${{ secrets.NEON_DATABASE_URL }}
+      CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+      CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+      NEON_AUTH_JWKS_URL: ${{ secrets.NEON_AUTH_JWKS_URL }}
+    concurrency: deploy-users-staging
+
   post-staging:
     name: Post-deploy (staging)
-    needs: [deploy-staging, deploy-web-staging]
+    needs: [deploy-staging, deploy-web-staging, deploy-users-staging]
     uses: ./.github/workflows/_post-deploy-test.yml
     with:
       environment: staging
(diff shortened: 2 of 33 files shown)


lifeodyssey:
## Review changes addressed (Codex xhigh REQUEST_CHANGES) — pushed `2bef84d` + `93a8e48`

CI green (18/18). Point-by-point:

### ✅ Required 1 — P1 non-atomic UPDATE / TOCTOU (`workers/users/src/api/routes.ts`)

The UPDATE now carries the ownership predicate inline and can't feed `undefined` to the mapper:

```sql
UPDATE routes SET title = $1, point_ids = $2::text[], status = $3,
  saved_at = CASE WHEN $4 = 'draft' THEN NULL ELSE COALESCE(saved_at, NOW()) END
WHERE id = $5 AND user_id = $6          -- ← was: WHERE id = $5
RETURNING ...
```

- **Atomic**: a racing ownership change between `assertOwner` and the write can no longer be written through — the predicate is in the single UPDATE statement.
- **Empty-rows guard**: new `updatedRoute(rows, id)` throws `routeNotOwned(id)` (403) when the UPDATE matches nothing, instead of `toUserRoute(undefined)` → 500.
- `assertOwner` kept — it still gives the precise 404 (no such id) vs 403 (wrong owner) distinction for the non-race path.

**Tests** (`test/routes-api.worker.test.ts`, new `describe("atomic route updates")`):
1. *"throws ROUTE_NOT_OWNED when an owned update loses the race"* — inline executor where `SELECT user_id` passes (`user-a`) but `UPDATE` returns `[]` (simulated concurrent reassignment) → asserts **403, not 500**.
2. *"includes user_id in the atomic update predicate"* — renders the captured UPDATE via `PgDialect.sqlToQuery` and asserts `.sql` contains `user_id` **and** `.params` contains `"user-a"` — proves the predicate is in the statement, not only the pre-check.

The shared fake (`test/helpers.ts`) UPDATE branch now matches on both `id` **and** `user_id` (param positions `[title, point_ids, status, status, id, userId]`), so it faithfully enforces the atomic predicate.

### ✅ Required 2 — P2 workers.dev bypass (`workers/users/wrangler.toml`)

`workers_dev = false` on all three envs (top-level, `[env.staging]`, `[env.production]`). The routeless service is now reachable **only** via the root Worker `USERS` binding — no public `users.<subdomain>.workers.dev` host. Matches the PR body's "routeless" claim and #330's style.

### ✅ Cheap 3 — point-id length cap (`packages/contract/src/users-contract.ts`)

`SaveRouteInput.point_ids`: `z.array(z.string().max(128)).max(500)` (per-item cap vs amplification). `users-openapi.json` regenerated (idempotent; catalog `openapi.json` byte-identical).

### ✅ Cheap 4 — legacy NULL title (`toUserRoute`)

Mapper coerces a `NULL`/non-string `title` to `""` instead of throwing, so a claimed legacy agent row (post-S2.9 + claim flow) can't 500 the whole list. `UserRoute.title` stays `z.string()`; `""` satisfies it.

### Deferred per coordinator (not in this PR)
delete + claim endpoints → follow-up card (PR body already flags them deferred); edge #328 vs users JWT-verify drift → later shared-package extraction.

### Gates (local, Node 24)
users tsc / eslint(0 warn) / **25** tests (was 23; +2 atomic-update, coverage 80/70/87/82 vs 60/60/60/50) · catalog 109 · root edge 18 · emit-openapi byte-idempotent · actionlint · secret scan clean.

PR #334, opened 2026-07-13
fix(eval): repair LocaleMatch measurement + query-first reply-language policy

lifeodyssey:
## What

`LocaleMatch 48.6%` (the #1 B-段 target from #304) was mostly the measuring stick lying, not the agent. This PR fixes the measurement, aligns the metric's ground truth with the SD-17 ③ policy, strengthens the agent's locale directive, and applies a baseline erratum (v1.2) — same offline-recompute method as the v1.1 RouteOrderCorrect erratum.

## Root cause (three layers)

**1. Detector bug (dominant — 73% of failures were false).** `detect_language` used first-match single-char search: any kana → `ja`, else any han → `zh`. Pilgrimage replies necessarily quote Japanese proper nouns, so correct zh/en prose was misread:

> 《你的名字》（**君の名は。**）的圣地主要分布在东京新宿·四谷地区…  ← perfect Chinese, judged "ja"

Of 288 scored failures: **211 false fails** (zh 110/110 — ALL zh "failures" were artifacts; en 101) plus **9 false passes** (real drift hidden by quoted kana). The bizarre `en→zh ×51` in the original report was English prose whose only CJK was kanji-only place names.

**2. Wrong expectation semantics.** The evaluator expected the browser `locale`, but the dataset's own J-family is literally `path: "query_lang_ne_locale"` and SD-17 ③ says current-turn text first, locale fallback. Evaluator and agent instruction now share one policy function — metric semantics and product semantics can no longer diverge (the IntentMatch lesson, structurally prevented).

**3. The real defect (now visible): en 66.5%.** DeepSeek answers **English users in Chinese** (56×) or Japanese (11×) — its native-language drift, the opposite of the "everything collapses to Japanese" headline. zh (97.8%) and ja (94.9%) are healthy.

## Changes

- **`agent/utils/language.py` (new)** — `detect_language`: strips quoted spans (「」『』《》【】（）()[]）, markdown table rows, and bold spans (the proper-noun carriers), then classifies remaining prose by script ratios (kana ≥ 2 → ja; han share ≥ 0.30 → zh; else en). `resolve_reply_language(query, locale)`: query-first policy — pure-latin → en; mixed pasted-title+foreign-words → locale; simplified-only han (你说圣选篮…) → zh; kana → ja; shinjitai-only han (気駅発楽…) → ja; undecidable → locale.
- **`evaluators.py`** — `LocaleMatch` expects `resolve_reply_language(query, locale)`, detects with the fixed detector.
- **`pilgrimage_agent.py`** — injected directive names the resolved target directly ("Respond in English. Proper nouns … may stay in their original script"), replacing the soft phrasing DeepSeek ignored.
- **`public_api.py`** — old detector removed; the translation gate now imports the fixed one. Production side effect: the gate stops firing **spurious LLM translation calls** on correct replies that merely quote Japanese place names (most zh/en responses previously took a pointless translate round-trip).
- **Baseline erratum v1.2** — offline recompute over the stored run's 560 messages: `locale_match` **48.6% → 85.7%** (213 flips to pass, 5 to fail). Zero LLM spend.
- **Tests** — new `test_language.py` (22 cases, samples mirror real misclassified outputs from the baseline run); pilgrimage-agent/evaluator-component tests updated for the new directive + semantics.

## Calibration evidence (all offline, zero LLM cost)

Detector + policy calibrated against all 560 stored messages and 647 queries; every residual failure was eyeballed and confirmed to be genuine wrong-language prose:

| | old metric | corrected metric |
|---|---|---|
| overall | 48.6% | **85.7%** |
| ja | 100% (9 false passes) | 94.9% |
| zh | 35.8% | 97.8% |
| en | 15.6% | 66.5% ← real target |

Remaining true failures: en→zh 56, en→ja 11, ja→zh 9, zh→ja 4.

## Gates

- `pytest` unit: **943 passed**, coverage 84.15% (floor 82%)
- `mypy` strict: 107 files clean
- `ruff check` + `format`: clean

## Validation route for the agent-side directive

The measurement fix is fully validated offline (above). The **directive change** (en drift) needs a live eval run to confirm — lands with the next trajectory-tier run (DeepSeek ~¥10, or free on Agnes if the provider probe pans out). The updated baseline keeps the regression gate honest either way.

## Adjacent findings (out of scope, not fixed here)

- `K3_ja_001` (`selected_empty_ids`): empty selection did not error gracefully — routed into a full zh search response (dispatch truthiness bug?).
- `L3_ja_001` (`empty_input`): empty query produced a hallucinated route plan in zh.
- The production translation gate targets browser `locale`; arguably it should target `resolve_reply_language(request.text, locale)` — product decision, deferred.

### DIFF
--- a/apps/agent/agent/agents/pilgrimage_agent.py
+++ b/apps/agent/agent/agents/pilgrimage_agent.py
@@ -32,6 +32,7 @@ from agent.agents.runtime_models import (
     RouteResponseModel,
     SearchResponseModel,
 )
+from agent.utils.language import resolve_reply_language
 
 COMPACT_THRESHOLD = 40  # ~5 turns × 8 messages/turn
 _KEEP_RECENT = 8  # Keep latest turn fully uncompressed
@@ -309,11 +310,15 @@ def _inject_session_context(ctx: RunContext[RuntimeDeps]) -> str:
     """Inject locale enforcement and current session state for multi-turn."""
     parts: list[str] = []
 
-    # Locale enforcement — respond in the user's query language, with
-    # browser locale as fallback when query language is unclear
-    lang = _LOCALE_NAMES.get(ctx.deps.locale, "Japanese")
+    # Locale enforcement — query language first, browser locale as fallback
+    # (SD-17 (3)). A direct order beats "the language the user writes in":
+    # DeepSeek drifts to Chinese for English users under the soft phrasing.
+    target = resolve_reply_language(ctx.deps.query, ctx.deps.locale)
+    lang = _LOCALE_NAMES.get(target, "Japanese")
     parts.append(
-        f"Respond in the language the user writes in. If unclear, default to {lang}."
+        f"Respond in {lang}. Proper nouns (anime titles, station and place "
+        f"names) may stay in their original script, but write all prose "
+        f"in {lang}."
     )
 
     state = ctx.deps.tool_state
--- a/apps/agent/agent/interfaces/public_api.py
+++ b/apps/agent/agent/interfaces/public_api.py
@@ -7,7 +7,6 @@ session management in ``session_facade``, and persistence in ``persistence``.
 from __future__ import annotations
 
 import asyncio
-import re
 from time import perf_counter
 from typing import cast
 from uuid import uuid4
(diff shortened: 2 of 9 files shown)


lifeodyssey:
Superseded / obsolete. The core fix — replacing the first-match `detect_language` bug with count-based script detection + the query-first `resolve_reply_language` policy — is already in `feat/frontend-rebuild` (see `agent/utils/language.py` `_script_counts`/`resolve_reply_language`, wired into `evaluators.py` LocaleMatch ground truth). This PR is 297 commits behind base and predates the #354 evaluator rewrite that absorbed it. Verified 2026-07-21: the held-out suite's LocaleMatch is measured by the count-based detector. Closing; any residual agent-directive delta should be re-filed fresh against current code.

PR #349, opened 2026-07-15
feat(agent): agent redesign umbrella — kill eval thrash (Phase 0 + 1a–1d land here)

lifeodyssey:
## What this PR is

**Phase 0 (SECURITY) of the agent redesign** — the first, safe, additive slice of a phased effort to kill the eval-thrash root cause. It ships two P0/HIGH security fixes + a dead-code cleanup, and includes the full redesign spec that the later phases implement.

**Not for merge yet — opened for owner review.** This is the first of a phased set (0 → 1a → 1b → 1c → 1d); see *Phasing* below.

## Why (the context)

A full MiMo eval surfaced pathological **thrashing**: trivial single-anime queries loop **27–50 model requests** and trip the usage limit. Root-caused (via a 6-batch best-practice audit, each batch dual-reviewed) as a broken **tool contract** — a resolver that fabricates ambiguity + a `clarify`-as-control-flow loop — **not** model weakness. The audit was synthesized into a redesign spec (`docs/superpowers/specs/2026-07-16-agent-redesign-spec.md`, included here), twice dual-reviewed (sonnet + Codex xhigh) and lead-verified. The audit also found two **security** issues that were pulled forward to this Phase 0 because they are independent of the (larger, later) atomic tool-contract switch.

## What Phase 0 does + how

**1. Model-alias allowlist — closes a credential-leak vector (§4g / M1).**
A caller-supplied `request.model` was passed **raw** to `Agent.run()`, so `openai:mimo@https://evil.example` routed to an attacker endpoint and `_resolve_api_key` shipped the generic compat key there. Now every non-`None` `model` override resolves through a **server-owned `MODEL_ALIASES`** allowlist (`default | deepseek | mimo` → fixed `{model spec, base_url, credential}`) **before** `run()`. Unknown alias / any `@`|scheme|URL|whitespace/non-`[a-z0-9_-]` string → typed `ModelAliasError` → **HTTP 400 `invalid_model_alias`** (never 500, never pass-through). Enforced on **all** boundary paths — sync `/v1/runtime`, SSE `/v1/runtime/stream` (validated upfront before the stream commits), and the `selected_point_ids` path. `model=None` and trusted internal `Model` objects unchanged.

**2. Translation sub-agent hardening (§4i / S1).**
The translation sub-agent used a **raw** DuckDuckGo tool (unsanitized web → prompt-injection) and neither shared usage nor inherited the run model. Now: DDG results are wrapped via `wrap_untrusted_web_results` (instruction-as-DATA delimiting); both child `.run()` sites pass `usage=ctx.usage` (tokens count toward the parent limit) + `model=ctx.model`; both catch `FallbackExceptionGroup`; DDG failures degrade to empty results (no 500).

**3. Delete dead `sql_agent`.**
Removed `sql_agent.py` + test, `SQLAgent`/`SQLResult` exports, `ResolvedLocation`, and **5 `# noqa: S608`** suppressions (zero remaining references) — a direct-DB LLM path that violated the catalog-only creed. Docs updated (`ARCHITECTURE.md`) per the same-change rule.

## How it was built + verified

Audit → spec (2× dual review) → **Codex wrote all code** → **lead independent verification** (project gate commands, not self-invented) → implementation dual review (Codex found 2 real contract gaps sonnet's approve missed; lead arbitrated — no leak, but SSE + selected-route returned 200 not 400) → **Codex fixed** → re-verified with real end-to-end HTTP tests.

- `make lint` + `make typecheck` (107 source files) + `make test` (**1,114 passed, 86.69%**) — all green.
- 29 new security tests, incl. real-app HTTP tests proving bad aliases → **400 + the agent sink is never awaited** on sync/SSE/selected paths, and injection text lands **inside** the untrusted-DATA block (not the prompt prefix).
- **Red line intact:** no eval dataset / baseline / `_INSTRUCTIONS` / serialized-`AgentResult` changes; no new suppressions.

## Phasing (this is Phase 0 of 5)

| Phase | Scope | Status |
|---|---|---|
| **0** | Security (this PR) | ✅ ready for review |
| 1a | Deterministic catalog resolver + points-by-work-id (TS) | spec ready |
| 1b | Typed session-state registry + persistence + RuntimeStage | spec ready |
| 1c | **The atomic tool-contract switch** (kills the thrash) | spec ready — has owner-gated eval decisions |
| 1d | Graceful usage limits + direct thrash gates | spec ready |

## Owner decisions (flagged, not blocking this PR)

1. **Branch base / stacking.** This PR targets `feat/pydantic-native-n1` (where the redesign is grounded), which is itself PR #347 (unmerged) → `feat/frontend-rebuild` → `main`. So this stacks 3 deep. Options: keep stacking on #347; or land #347 first then rebase; or make the redesign its own umbrella. Retarget freely.
2. **Phase 1c eval strategy (later).** The atomic switch removes `output.data`; lead verification found the frozen `data_keys` baselines are **not** auto-neutral (500+311 cases carry non-empty `expected_data_keys` sourced from `output.data`) → §7 needs a SessionState→legacy-key compatibility projection before 1c, and any `expected_stage` flip needs your sign-off. Documented for the pre-1c touch-up.


lifeodyssey:
## ✅ Phase 1a landed — deterministic catalog resolver (additive)

Second phase of the redesign. **Additive TypeScript** in `workers/catalog` + `packages/contract`; the existing `search` endpoint is untouched and **nothing consumes these yet** (the agent wiring is Phase 1c). This is the *data-layer* fix for the thrash root cause: the model will branch on a deterministic typed `outcome` instead of judging ambiguity itself.

### What was added
- **`POST /catalog/resolve`** → a discriminated `ResolveOutcome`: `resolved | needs_disambiguation | not_found | upstream_unavailable`.
  - **HIT path** (alias index): `GROUP BY work_id` dedup — this closes the *franchise false-ambiguity bug* (multiple alias/source rows for one work can no longer look like "multiple candidates"). Top-priority dominance → `resolved`; a genuine tie at the top → `needs_disambiguation`, ranked by `(priority DESC, points_count DESC NULLS LAST, bangumi_id ASC)`, capped at 6.
  - **MISS path**: fetch N=8 Bangumi subjects (reusing the shared parser), normalized-exact-match threshold, else relevance-head resolve. Deliberately **no** "2+ subjects → clarify" (that would re-explode the clarify partition).
  - Enrichment derives `points_count` via `COUNT(points.id)` in the query (no schema migration); `year` from `air_date`. Typed upstream 5xx/network/bad-JSON → `upstream_unavailable` (distinct from `not_found`).
- **`POST /catalog/points-by-work-id`** → returns `SearchResult` by reusing the existing `hitResult → pointsForWork → selectPoints` (no free-text re-search).

### How it was built + verified
Codex wrote all code → lead independent verification → **dual review** (sonnet *approve* + Codex *request-changes*; core confirmed faithful to §4a) → Codex fixed → re-verified.

The review earned its keep — it caught, and this branch fixes, **two latent 500 crash paths** that would have gone live at Phase 1c (a malformed Bangumi subject crashing the whole resolve; an orphan alias 500-ing the common resolved path), a **NULLS-LAST comparator `NaN` bug** that silently dropped the tiebreak (sonnet verified it in Node), plus blank-query validation, a real Postgres integration suite, and a DRY cap. Independent gate re-run also caught an unhandled-rejection the subagent's own run missed — now fixed (malformed upstream JSON → `upstream_unavailable`).

- `lint:oxlint` + `typecheck` + contract `typecheck` + `test:worker` all green — **28 files, 210 tests, 91% coverage**, no unhandled rejection. DB-backed `.spike` integration tests run in CI.
- `search` endpoint, `apps/agent`, eval datasets + baselines **untouched**.

### Next
Phase 1b (typed session-state registry + persistence + RuntimeStage mapping) → then the owner-gated **1c** atomic switch. All land on this branch → this stays one PR.

lifeodyssey:
## ✅ Phase 1b landed — session-state scaffolding + RuntimeStage mapping (additive, eval-neutral)

Third phase. **Additive Python with COMPAT READS** — current behavior + the frozen eval baselines are byte-identical. This lays the typed carrier + persistence + intent-mapping the **Phase 1c atomic switch** flips onto; nothing consumes it yet.

### What was added
- **Typed `SessionState`** (`agents/session_state.py`): `CurrentAnime`, `SearchPayloadState`, `RoutePayloadState`, `OrderedCandidate`, `PendingClarification`, `GeocodeStaging`; `ResultRef`/`RouteRef` registries with `MAX_REFS=8` LRU + a typed stale-ref sentinel; all `extra="forbid"`. Attaches as `deps.tool_state.session`, **excluded** from the legacy `tool_state` serialization (keeps SSE/eval byte-compatible). Inert in 1b (the tools populate it in 1c).
- **Versioned persistence** (`session_facade`): a `session_state_v2` delta that round-trips, with safe fallback for old/malformed snapshots. The **sticky `pending_clarify` bug fix is deliberately deferred to 1c** (it changes multi-turn behavior → it belongs with the eval sign-off, not in an eval-neutral phase).
- **Server-owned `RuntimeStage` map** (`animichi_runner` + `agent_result`): `AgentResult.intent` becomes a stored field; new `session_state`/`status`/`success_override` fields default neutrally. The mapped intent **equals** the legacy `output.intent` for every output type — guarded by a parametrized equivalence test — so `intent`/UI/eval are unchanged.

### Dual review + fixes
Codex wrote all code → lead independent verification → dual review (**sonnet approve** + **Codex request-changes**; both confirmed eval-neutral, sonnet verifying the `is_empty()` neutrality linchpin against the installed Pydantic).

The adversarial pass caught, and this branch fixes, real **persistence-scaffolding correctness** issues that would have bitten at 1c: a snapshot-resurrection bug (a malformed newest state scanning back to a *stale older* one), a **JSONB-unsafe LRU** (recency rode on Python dict key order, which PostgreSQL `jsonb` reorders — now explicit ordered ref lists, with a test that reorders the serialized keys and proves the true-oldest ref is still evicted), and a non-recursive strict boundary (nested unknown fields silently dropped → now strict session-local snapshot models).

- `make lint` + `make typecheck` (108 files) + `make test` (**1153 passed, 87.13%**) green.
- **Eval-neutral confirmed by both reviewers** — no `tests/eval`, baseline, `_INSTRUCTIONS`, output-model, `response_builder`, or catalog changes; the sticky OR is byte-identical to base.

### Next — Phase 1c is the owner-gated core
1c is the ONE atomic switch (new tool contracts wired to the 1a endpoints, terminal clarify, compact outputs, the validator rewrite, instructions, and migrating every consumer). It **removes `output.data`**, which flips the frozen `data_keys` eval metric for ~811 cases → it needs the SessionState→legacy-key compatibility projection and the owner's sign-off on any `expected_stage` change. Pausing here for that.

lifeodyssey:
## ✅ Phase 1c landed — the atomic switch (thrash killed) + eval confirmation

The core of the redesign. One atomic switch removes `output.intent`/`output.data`/the `clarify`+`greet`+`qa` **function tools** at once and flips the agent to the typed-outcome contract: the model now **branches on deterministic outcomes computed in the data layer** — it never judges ambiguity and never loops. Net **−2,757 lines** (the redesign simplifies).

Also folded in: the loose name-similar clarify + over-clarify guards (the Phase-1a resolver amendment), multi-select (`execute_multi_selection` with the T1–T6 terminal matrix) + place selection, the normalized `route_anime` join table, and the D3-aware eval harness.

### Dual review → typed provenance fix
Both reviewers (sonnet + Codex xhigh) confirmed the **core is sound** (validator complete, total clarify transition table holds, the model cannot loop, T1–T6 semantics correct, every consumer migrated, `/v1/chat` on the unified boundary). They caught **2 real P1s the green gate missed** — both cross-turn bugs invisible because every test starts from a fresh SessionState: a non-success terminal (T4/T5/T6/error) projecting a **prior turn's** route/search, and the validator authorizing an output against a stale registry entry. Fixed with a **typed `TurnProvenance`** (Produced/Rejected Search/Route value objects) so the response builder projects — and the validator authorizes — **only this turn's exact ref**; cross-turn regression tests (seed a prior route + fresh pending → assert no stale projection) now guard the class.

- Agent gate: `make lint` + `make typecheck` (106 files) + `make test` (**1,045 passed, 86.5%**) green. Catalog gate (city touched TS): oxlint + typecheck + `test:worker` (216) green. Untrusted-tool-output invariant kept verbatim; no new suppressions.

### 📊 Eval — thrash killed, quality up (50-case MiMo subset, owner-confirmed)

The datasets were re-baselined to the new contract (168 clarify expectations `question/options/status → reason+candidates`, + 5 D3 selection cases). A 50-case MiMo trajectory run vs the old 643-case baseline:

| Metric | Old | New | Δ |
|---|---|---|---|
| **tool calls / case** | **27–50 (thrash)** | **median 2 · mean 2.6 · max 17** | 🔥 collapsed |
| **usage-limit blow-ups** | present (hit the 27 cap) | **0** | 🔥 |
| tool_f1 | 0.763 | 0.829 | +6.6pt |
| route_order_correct | 0.768 | 0.840 | +7.2pt |
| locale_match | 0.540 | 0.780 | +24pt |
| nonempty_results | 0.846 | 1.000 | +15pt |
| tool_recall | 0.800 | 0.870 | +7pt |

Even the worst new case (17 tool calls) sits below the old thrash floor (27), and **no case hit the usage limit**. Honest caveats: `step_efficiency` only nudged (0.811→0.823 — it's a ratio-to-minimum; a small efficiency tail remains — one 17-call outlier + ~20% over the official max-tool-calls threshold), and the N2 `*_official` metrics have no old baseline to compare. Owner confirmed the effect on the subset; the full 655-case run + formal baseline write is deferred.

### Next
Phase 1d (graceful usage limits + a distinct `PartialResponseModel`) → capabilities/model/sub-agent cleanups (Phase 2–4) → the deferred `logfire[asyncpg]` extra + the Logfire `production` prompt publish. All land on this branch → this stays one PR.

lifeodyssey:
## ✅ Phase 1d landed — graceful UsageLimits + `PartialResponseModel` + direct thrash gates (`bf044fd`)

Closes the runner's degradation seams and locks in the thrash fix, without disturbing the 1c core.

### Graceful degradation (no more 500s on the limit)
`UsageLimitExceeded` and any plain-string model output now return a **distinct, server-only `PartialResponseModel`** (stage/status `"partial"`, runner-owned `success=False`) instead of raising. Key properties:
- **The model can never emit it** — `PartialResponseModel` is excluded from the model output union (`RuntimeStageOutput` / `_output_types()`); only the runner's `_partial_result` constructs it.
- **Projects only the current turn** — it reuses the 1c typed `TurnProvenance`, so a partial shows the search/route it produced *this* turn and **never leaks a stale registry ref** (`last_result_ref`/LRU are never consulted).
- **Persists + streams like a completed turn** — the assistant message + any provenance-supplied route are written to history (so a partial route survives a reload), and SSE completes as a normal `data-response`, not an error frame. `success=False` is retained for analytics only.

### Direct thrash red-line gates — report-only by default (unblocks the re-baseline)
Deterministic per-case gates over **model-initiated** steps only: `requests≤12 · tool-calls≤6 · repeated-identical=0 · p95≤6` (p95 excludes zero-request deterministic/selection cases so they can't dilute the ceiling). They lock in the 1c thrash fix against regression.

**They default to REPORT-ONLY** (`DIRECT_GATE_ENFORCE` unset). This is deliberate and matters for the re-baseline (task #22): the gates **surface** the numbers on the full eval but are **not merge-blocking**, and a bootstrap run still writes the baseline. Otherwise the known post-1c efficiency tail (a 17-tool-call outlier + ~20% over tool≤6) would make the full 655-case eval refuse to establish the new baseline — deadlocking the owner sign-off. Per spec §7 (*calibration precedes gating*), you flip `DIRECT_GATE_ENFORCE=1` **after** confirming calibrated thresholds.

### Dual review → 9 reconciled fixes
Both reviewers (sonnet + Codex xhigh) **cleared the 1d core** (union exclusion, graceful catch, `usage` accounting, current-turn provenance, `model_initiated` filter, test genuineness) and both returned **request-changes on the glue** — exactly what a green gate can't catch. Reconciled into 9 fixes:

| # | Fix | Source |
|---|---|---|
| P1 | Direct gate defaulted to report-only (the `DIRECT_GATE_ENFORCE` flag) — was hard-blocking the §7 re-baseline | sonnet |
| P2 | A partial turn now persists (assistant message + provenance-supplied route) — was dropped as an ordinary failure | Codex |
| P2 | p95 excludes zero-request cases — deterministic selection cases were diluting the ceiling | sonnet |
| P2 | Graceful `UsageLimitExceeded` no longer logs a Logfire **ERROR** (was inflating the error-rate dashboard on a 200 outcome) | sonnet |
| P2 | Capped runs surface the direct metrics (report-only) instead of total silence | sonnet |
| P3 | Strengthened the stale-ref test to fresh-**and**-stale-in-one-turn (search + route) — proves current wins over stale | both |
| P3 | Persistence tests now assert the actual `insert_message` + `save_route` DB writes | Codex |
| P3 | Spec §4h corrected — it literally said "project `last_result_ref`", which is the 1c stale-ref bug; now requires current-turn provenance | sonnet |
| P3 | `_partial_message` unknown-locale fallback → `ja` (matches the app-wide default) | sonnet |

I verified independently (not just the green gate): read the P1/persistence/observability diffs, confirmed the new tests genuinely assert the DB writes, and cleared Codex's tool-recording finding as a non-issue (tools record their typed outcome unconditionally + the request gate backstops).

**Deferred (out of 1d scope, noted for task #22 / N2):** a cross-dataset aggregate gate (v3 + runtime-journey + the 15 regressed-MiMo union) and errored-case trajectory coverage — eval-architecture expansions, not a 1d bug.

- Gate: `make lint` + `make typecheck` (107 files) + `make test` (**1063 passed, 86.56%**) green. No new suppressions.

### Next
Phase 2–4 cleanups (capabilities / model layer / sub-agents) → deferred `logfire[asyncpg]` extra + the Logfire `production` prompt publish → full 655-case MiMo eval + `DIRECT_GATE_ENFORCE` calibration (task #22). All on this branch — still one PR.

lifeodyssey:
# 🎯 Agent redesign — the thrash-kill arc (Phases 0→1d) is complete

This is the capstone summary of the whole redesign now landed on this branch: **what** we changed, **the effect** it had, and **how** we built it. The thrash-fix core (1a→1d) is done and eval-confirmed; Phases 2–4 (modernization cleanups) come next on this same PR.

---

## 1. What we did — and why

### The problem
A MiMo trajectory eval surfaced pathological **thrash**: trivial single-anime queries ("秒速5厘米的取景地") looped **27–50 model requests** per turn and routinely hit the usage limit (a 500). A 6-batch best-practice audit (each dual-reviewed) traced the root cause to a **broken tool contract**, *not* a weak model:
- `resolve_anime` **fabricated ambiguity** (the catalog physically returns one work) and **offloaded a "is this query short/vague?" judgment** onto the model.
- `clarify` was implemented as a **`ModelRetry` control-flow loop** — the model could re-enter it indefinitely.

So the model wasn't misbehaving; the harness was *asking* it to loop. The fix had to be **agentic** (no hardcoded loop-breakers), following pydantic-ai official best practice.

### The redesign
One umbrella, built in reviewable phases:

| Phase | What changed |
|---|---|
| **0** | Security hardening of the selection/SSE boundary (dispatch-by-reason, upfront validation). |
| **1a** | Catalog resolver returns **typed discriminated outcomes** (`ResolveResolved/Ambiguous/NotFound/UpstreamDown`) + a **loose name-similar** clarify predicate (deterministic substring containment, CJK-aware) — the model never judges ambiguity. |
| **1b** | Typed `SessionState` registry scaffolding (server-owned refs for search/route payloads). |
| **1c** | **The atomic switch** — removed `output.intent` / `output.data` / the `clarify`+`greet`+`qa` function tools in one move and flipped the agent to a **typed-outcome contract**: the model branches on deterministic outcomes computed in the data layer and cannot loop. Plus multi-select (`execute_multi_selection`, terminal matrix T1–T6), place selection, a normalized `route_anime` join table, and a D3-aware eval harness. **Net −2,757 lines.** |
| **1d** | Graceful `UsageLimitExceeded` → a distinct server-only **`PartialResponseModel`** (no more 500s; partial results persist + render), + **direct thrash red-line gates** (req≤12 / tool≤6 / repeat=0 / p95≤6) defaulted to report-only. |

The load-bearing idea: **decisions move from the model to the data layer.** Tools return typed outcomes; a server-owned validator authorizes only what *this turn* actually produced (typed `TurnProvenance`); the model's job shrinks to picking a branch. There is no `ModelRetry` control-flow loop left to thrash in.

---

## 2. The effect — eval, before vs after

50-case MiMo trajectory subset (owner-confirmed; full 655-case run is task #22), against the pre-redesign baseline:

| Metric | Before | After | Δ |
|---|---|---|---|
| **tool calls / case** | **27–50 (thrash)** | **median 2 · mean 2.6 · max 17** | 🔥 collapsed |
| **usage-limit blow-ups** | present (hit the cap) | **0** | 🔥 |
| tool_f1 | 0.763 | 0.829 | +6.6pt |
| route_order_correct | 0.768 | 0.840 | +7.2pt |
| tool_recall | 0.800 | 0.870 | +7pt |
| locale_match | 0.540 | 0.780 | +24pt |
| nonempty_results | 0.846 | 1.000 | +15pt |

**Thrash is eliminated** and quality rose across the board — even the *worst* new case (17 calls) sits below the old thrash *floor* (27), and no case hit the limit. Honest caveats: `step_efficiency` only nudged (0.811→0.823 — a small tail: the 17-call outlier + ~20% of cases over the strict tool≤6 line), and the full 655-case baseline write is deferred to task #22.

---

## 3. How we did it — the method

Every phase ran the same disciplined loop, and it repeatedly paid off:

1. **Audit → spec → dual review.** A 6-batch contract audit produced a redesign spec (v2.3), dual-reviewed (Fable + Codex) including a re-review of the owner's three decisions (re-baseline the eval, loose clarify, multi-select — explicitly *not* CodeMode).
2. **Policy B implementation.** Codex (gpt-5.6-sol, xhigh) writes all product code; the lead does orchestration, git, and **independent verification** — always re-running the project's own gate (`make lint` + `make typecheck` + `make test`), never trusting a "tests green" claim.
3. **Adversarial dual review every phase.** sonnet + Codex xhigh review each phase's diff. **This caught real bugs the green gate could not** — e.g. in 1c, two cross-turn stale-ref P1s (invisible because every test starts from a fresh session), fixed with typed `TurnProvenance`; in 1d, an eval-gate that would have **deadlocked the owner re-baseline** by refusing to write a baseline while the known efficiency tail exists, plus a graceful partial that still logged a Logfire ERROR. The lead reconciles the two reviews — de-duplicating overlaps, de-escalating over-scoped findings, and disproving non-issues — before Codex applies the fixes.
4. **One design call worth flagging:** the direct thrash gates ship **report-only** (`DIRECT_GATE_ENFORCE` env, default off). They *surface* the per-case numbers on the full eval but don't block, so the re-baseline (task #22) can establish a baseline; you flip enforcement on **after** calibrating thresholds — per spec §7 (*calibration precedes gating*).

The through-line: the green gate proves the code *runs*; adversarial dual review + independent verification prove it's *correct*. Each phase's per-phase comment above documents its specifics.

---

## 4. What's next
- **Phase 2–4** (this PR): capabilities cleanup → model layer → sub-agents/guardrails — modernization hygiene, not the thrash fix.
- **Deferred follow-ups:** the `logfire[asyncpg]` extra declaration; publishing the local instructions to the Logfire `production` managed prompt; the full 655-case MiMo eval + `DIRECT_GATE_ENFORCE` calibration (task #22).
- Still **one PR** — this stays the single umbrella for your review.

lifeodyssey:
## ✅ Phase 2 landed — capability-composition cleanup (`33474b7`)

Modernization hygiene, not the thrash fix. Net **~−440 lines**.

### What changed
- **Deleted the legacy composition path + the `ANIMICHI_MODERN_COMPOSITION` rollback flag.** `build_animichi_agent()` is now no-arg with a single native history-compaction path and an extracted `_modern_capabilities()`. The legacy hand-rolled compaction helpers (`_compact_tool_results`, `_sliding_window`, …) are gone; `_summarize_*` stays (native compaction depends on it).
- **Removed `OverflowingToolOutput`.** Post-1c the catalog tools return only `row_count` + an opaque `result_ref` to the model (rows live in `SessionState`), so a large catalog tool return is physically impossible — the overflow guard was dead weight.
- **Web tools are now eager** (dropped `defer_loading` + the `ToolSearch` capability): a 2-tool web menu doesn't need on-demand loading.
- Removed the dead `preview` branch in `_extract_anime_title`.

### ⚠️ Spec drift — this phase was reconstructed from the real code, not implemented verbatim
The spec's §6 Phase 2 bullet named several things that **don't exist** in the current tree (I mapped it before dispatching): there is no CJK tokenizer in compaction, no `max_tokens` sentinel constant, and no `_model_summary` function. Rather than invent code to match stale names, I scoped the *actual* cleanup from a fresh read and marked the phantom items as no-ops. The `SlidingWindow.max_tokens` was **kept** (the installed harness validation requires it) with a comment; the "candidates-via-registry" idea was **assessed and rejected** — `pending_clarification` is current mutable state that could mis-associate candidates with a historical turn, so per-return preservation is safer.

### Dual review → the real find
sonnet **approved**; Codex xhigh **requested changes**. Both independently flagged the same latent bug, and sonnet's deeper trace calibrated its severity:

- **The history-compaction summarizers were parsing a RETIRED shape.** `_candidate_summary` / `_summarize_resolve` / `_extract_anime_title` still keyed on `{ambiguous, candidates, metadata, title}` — but post-1c the tools emit discriminated outcomes `{outcome: "needs_disambiguation", candidate_ids, anime_title}`. So the ordered-candidate preservation silently no-op'd and titles dropped, and **the tests fabricated the retired shape, so they passed while proving nothing.** A latent bug carried over from 1c (dormant in practice because an ambiguous return is small enough to survive verbatim, but a landmine for long conversations with many candidates). **Fixed:** the summarizers now parse the current outcome shape and preserve `candidate_ids` verbatim + ordered; the fixtures are rebuilt from real `ResolveAmbiguous/ResolveResolved/SearchOk.model_dump()` output so the tests prove production behavior.
- **`web_search` description narrowed to QA-only.** Its docstring listed "look up a pilgrimage location" with a "宇治駅 anime pilgrimage spots" example — harmless when the tool was hidden behind `ToolSearch`, but now that it's eager it could invite catalog mis-routing. It now explicitly steers pilgrimage discovery to `search_nearby`/`search_bangumi`. The behavioral routing check (does an eager menu raise web calls on catalog queries?) is deferred to the batched eval (task #22).

Codex's third finding (unbounded model-facing strings) was **traced clean** by sonnet and dropped — every model-facing catalog return is a compact typed outcome (refs + counts + short IDs), and our catalog titles are bounded; no schema cap warranted.

- Gate: `make lint` + `make typecheck` (107) + `make test` (**1049 passed, 86.59%**) green. Test count dropped from 1063 (legacy-path tests removed) with coverage steady — no modern behavior lost its only assertion.

### Next
Phase 3 (model layer: alias-registry ergonomics, `trust_env`, timeout/fallback, shared lifespan client, profile-based thinking-disable) → Phase 4 (sub-agents/guardrails). Still one PR.

lifeodyssey:
## ✅ Phase 3 landed — model-layer engineering + **MiMo→DeepSeek prod switch** (`1dfa361`)

> ## 🔑 OWNER ACTION REQUIRED before deploy
> Set two **GitHub repo secrets** — the deploy will fail without them (the worker now hard-requires both, and `deploy.yml` provisions them from these secrets):
> - **`MIMO_API_KEY`** — the new prod **primary** (MiMo)
> - **`DEEPSEEK_API_KEY`** — the **fallback** (was a manual CF secret; now synced via `deploy.yml`)

### The production model switched to MiMo (owner decision)
Prod now runs **`FallbackModel[ primary = MiMo mimo-v2.5, fallback = DeepSeek v4-flash ]`** — MiMo is ~68% cheaper and the redesign was validated on it; DeepSeek stays as a proven fallback that fires only on MiMo failure (no cost hit). The default request path, the CF worker credential forwarding + hard-require, `validate_api_keys`, `.env.example`, `wrangler.toml`, and the deploy docs are all wired to this. **MiMo sends NO thinking-disable param** — its acceptance of DeepSeek's `extra_body` is unverified; verify against the MiMo API before enabling it (a 400 would force every call to fall back to DeepSeek). DeepSeek keeps thinking disabled.

### Model-layer engineering (§4g)
- **Credentials unified** through `get_settings()` (one `CredentialRef` policy, no direct `os.environ`; added `mimo_api_key`) — env var names unchanged.
- **Shared `trust_env=True` httpx client** owned by the FastAPI lifespan — now including the `model=None` hot path (both reviewers flagged the earlier half-wiring; the hot path no longer runs on an unclosed import-time client).
- **Timeout/fallback inversion fixed** — `model_attempt_timeout=45 < agent_deadline=100` (validator-enforced), `max_retries=0` under a FallbackModel, and the primary==fallback guard now compares the effective `(provider_kind, name, base_url)`. Failover previously could *never* fire inside the deadline (120s attempt > 90s wall); now it does.
- **Profile-based thinking-disable** (by model, not DNS host) — fixes the prior inconsistency where the mimo alias got none.
- **Import-time-validated immutable alias registry** raising a typed `ModelAliasRegistryError` (was lazy + a bare `ValueError` → uncaught 500); `_FALLBACK_MODEL → _DEFAULT_MODEL_SPEC`.

### The two things you asked about
- **Client identity to the providers:** every LLM call now carries **`X-App-Client: animichi Prod|Staging|Dev`** (by `app_env`), via the OpenAI SDK `default_headers` (the httpx User-Agent is overridden by the SDK, so a custom header is the mechanism that actually reaches MiMo/DeepSeek). Logfire already tags `environment=app_env` + `service=animichi-runtime`.
- **The deploy drift had no test — now it does.** The worker required `DEEPSEEK_API_KEY` while `deploy.yml` provisioned `GEMINI_API_KEY`; individually tested, their *alignment* was not. Added a **deploy-consistency test** asserting every `CONTAINER_REQUIRED_KEY` is forwarded AND provisioned in `deploy.yml`'s Wrangler `secrets:` block. **Gold-standard verified**: it FAILS on the pre-fix workflow (`Extra items: MIMO_API_KEY, DEEPSEEK_API_KEY`). (worker/entry.ts had routing/auth tests but not this; `infra/` has Pulumi config but no unit test.)

### Review + verification
Two dual-review rounds (sonnet + Codex xhigh). sonnet's final verdict: **approve, no deploy-breaker** — a missing key is a loud startup crash, not a silent 401; failover fires on real `ReadTimeout → APITimeoutError → ModelAPIError`; the **30+ test-file changes are mechanical adaptations with net +10 asserts, zero suppression, zero weakened behavioral coverage**. All P2/P3 items were then fixed here. I independently re-ran the gate and gold-standard-verified the deploy-consistency test.

- Gate: `make lint` + `make typecheck` (107) + `make test` (**1082 passed, 87.36%**) green.

### Next
Phase 4 (sub-agents / guardrails: translation catalog-ify, `sql_agent` delete, `web_trust` consolidation). Still one PR.

lifeodyssey:
## ✅ Phase 4 landed — sub-agents / guardrails (`d878704`) — **the redesign umbrella is complete**

The final phase: translation catalog-ify + guardrail consolidation. (`sql_agent` was already deleted on this branch.)

### Translation catalog-ify (§4i)
Anime-title Chinese localization now resolves through the **catalog contract** (`catalog.resolve` → `title_cn`) instead of a legacy DB→Bangumi-API→web-search cascade. English titles, Japanese place names, and general text go through a **tool-less `Agent[None, str]`** — the web-search tool, the DDG plumbing, and `translation_bangumi.py` are deleted. Two correctness wins:
- **`source`/`confidence` are app-assigned** (catalog=1.0 / llm=0.6 / untranslated only for blank/exception) — the model can no longer fabricate a `"web_search"` source (an adversarial test proves it).
- **Catalog is consulted ONLY for `kind="anime_title"`** — a Chinese place name (宇治駅, 秋葉原) never hits the anime resolver, so `秋葉原` can't be mis-resolved to the anime `秋葉原電脳組`. (Conscious tradeoff: the old 萌娘百科/Wikipedia web fallback for long-tail titles not yet in the catalog is gone — to be measured by the translation eval.)

### Guardrails (§4i/§4f)
Consolidated the three injection sites: deleted the redundant `public_api` scan; `detect_prompt_injection` stays a **log signal**; and the input guard moved from `InputGuard(replace)` to a **runner preflight** — it runs on the user text *before* the model, **always logs** the injection signal (restores the always-on signal the default-off `InputGuard` had dropped), and **blocks with a typed server-only `BlockedResponseModel` (zero tokens)** only when `ANIMICHI_INPUT_GUARD=1` (default off, per §9).

### Dual review caught a real P1
Both reviewers (sonnet + Codex xhigh) flagged that the *first* block implementation added **`str` to the model's `output_type`** so the block message would pass validation — which **re-opened plain-text model output**, violating the 1c typed-outcome contract (a correct QA answer emitted as prose would have been silently converted into a false "processing limit" partial), and the block message itself got mangled into a `"partial"`. Fixed: **`str` removed** — I built the agent and confirmed its output schema is `ToolOutputSchema` with **`allows_text=False`** (the model physically cannot emit plain text) — and the block is delivered as the server-only `BlockedResponseModel` via the preflight, keeping the contract intact.

- Gate: `make lint` + `make typecheck` (106) + `make test` (**1061 passed, 87.23%**) green. No suppressions.

---

## 🏁 The redesign umbrella (#349) is now complete — 0 / 1a / 1b / 1c / 1d / 2 / 3 / 4
The thrash-fix arc (1a–1d) is eval-confirmed (median 2 tool calls vs 27–50, quality up across the board); Phases 2–4 modernized capabilities, the model layer + **MiMo→DeepSeek prod switch**, and the sub-agents/guardrails. Every phase was independently gate-verified + dual-reviewed, and each review caught real bugs the green gate could not. **Still one PR, for your review.**

**Before merge/deploy:** (1) set the GitHub secrets **`MIMO_API_KEY`** + **`DEEPSEEK_API_KEY`** (Phase 3); (2) the full 655-case MiMo eval + `DIRECT_GATE_ENFORCE` calibration + the new-baseline write are deferred to task #22; (3) verify MiMo accepts (or keep disabled) the thinking-param before enabling it.

lifeodyssey:
## 🔧 Follow-up (`79d3500`): prod default → **MiMo-only** (DeepSeek fallback temporarily disabled)

Amends the Phase-3 mimo+deepseek default. On verifying the keys before deploy, the **DeepSeek account returned 402 Insufficient Balance** — the key authenticates but has no credit, so the fallback couldn't actually fire (a MiMo failure would have been total failure, not a graceful fall-back). Per the owner ("暂时就只mimo"), the prod default is now **MiMo-only**: `fallback_agent_model=""` → `get_default_model()` returns a bare MiMo primary (no `FallbackModel`, retries=0 — verified on the built model). The DeepSeek key + worker/deploy wiring stay **provisioned and dormant**, so re-enabling is a one-line flip (`fallback_agent_model` → `deepseek:deepseek-v4-flash`) after a recharge.

**Secrets status:** `MIMO_API_KEY` is set (the current, valid `sk-` key — MiMo authenticates + has balance). `DEEPSEEK_API_KEY` remains provisioned (valid key, no balance). `validate_required_env` now hard-requires only `MIMO_API_KEY` for the MiMo-only default.

**Also verified:** MiMo accepts `extra_body={"thinking":{"type":"disabled"}}` (HTTP 200, not 400) — so enabling thinking-disable is technically safe, but it's kept OFF because the redesign eval ran with MiMo's default (thinking on); disabling it is a cost↔quality tradeoff to measure in the #22 eval, not a free flip.

Gate: lint + typecheck (106) + test (1062, 87.23%) green.

lifeodyssey:
# 🏁 Final gate — full 655 eval + umbrella dual review → **merge-ready** (`5d03cca`)

## Full 655-case MiMo eval (the "定心丸")
Ran the complete re-baselined set against MiMo (~21 min, 15/655 errored ≈2.3%, direct thrash gates report-only → all passed):

| Metric | Old baseline (n=643) | **Full n=655** | Δ |
|---|---|---|---|
| **request p95** | **27–50 (thrash)** | **7** | 🔥 thrash killed |
| tool_recall | 0.800 | 0.855 | +5.5 |
| tool_f1 | 0.763 | 0.817 | +5.4 |
| route_order_correct | 0.768 | 0.798 | +3.0 |
| **locale_match** | 0.540 | 0.739 | **+20** |
| step_efficiency | 0.811 | 0.802 | −0.9 (flat) |
| nonempty_results | 0.846 | 0.769 | −7.7* |

\* the nonempty evaluator was rewritten in the re-baseline (reads `source_ref` now) — not apples-to-apples with the old contract; the 15 errored cases also drag it. The N2 `*_official` metrics (tool_correctness 52%, trajectory_match 69%) have no old baseline. Honest read: **thrash is dead on the full set** (p95 7, near-zero repeats), core quality up (tool_f1 +5.4, locale +20), with a real tail (nonempty + official) for future work. The 655 baseline was **not committed** — it awaits your §7 sign-off.

## Umbrella dual review (Fable 5 + Codex 5.6-sol xhigh)
Both independently **request-changes** and **converged on ONE blocker** — everything else verified sound end-to-end (both confirmed: the 1c typed-outcome contract holds — output schema is 5 `ToolOutput`s, `allows_text=False`, no reintroduced `ModelRetry` loop, Partial/Blocked server-only; current-turn provenance; the MiMo-only deploy chain; the security posture; the migration).

**The blocker (fixed in `5d03cca`):** the redesign's own new typed terminals returned **HTTP 500** on `POST /v1/runtime` — two seams that never met. (1) `run_nearby_search` recorded its clarify outcomes as *failed* steps while `run_resolve` recorded them as success, so place-clarify finished `success=false` → 500 **and lost its assistant message**; (2) `_http_status_for_response` didn't know `partial`/`blocked`/`empty`/`too_large`/`needs_clarification`/`invalid_selection`. Fixed: typed deterministic outcomes now record `success=True` (provenance unchanged); graceful terminals → 200, `invalid_selection` → 400; **new `/v1/runtime` HTTP-layer tests pin each** (there was zero coverage). Also folded in: T4/T6 no longer fabricate a catalog-failure error, `blocked` turns persist their refusal, the eval default is now MiMo (was the dead DeepSeek), and the runtime docs are corrected. Verified: `allows_text=False` still (no `str` re-added), gate green (1071 tests, 87.29%).

## ✅ Merge-readiness
After the blocker fix, both reviewers' assessment is **ready to merge + deploy** — once you (a) **review/merge**, and (b) confirm the `MIMO_API_KEY` GitHub secret (already set) + note prod is MiMo-only (DeepSeek fallback dormant, 402 no balance).

### Follow-up cards (NOT blockers — for after merge)
- **Needs your decision:** ingest-on-demand is unreachable from the new agent path (an un-catalogued new-season work returns "no points" forever, and the catalog no longer grows from user demand) — restore background `ingestWork` on resolve-MISS, or sign off that the catalog grows only via offline ingest.
- CI main promotion deploys only the catalog Worker, not the root MiMo container (staging can't pre-validate the switch); SessionState snapshot ~20× amplification in retained interactions; + minor P3s (localhost dev-key exemption, dead `_partial_result` param, instructions "empty" case, clarify prior-turn pending, `/v1/chat` origin fields, catalog `APIError` aborts a turn, migration expand/contract window).
- Deferred (acknowledged): the 655 baseline write + `DIRECT_GATE_ENFORCE` calibration, the MiMo thinking-param eval, the `logfire[asyncpg]` extra. CI integration suite must be green before merge (not runnable in the sandbox).

lifeodyssey:
## ✅ Owner-approved follow-ups landed (`3eed9b6`) — demand-driven ingest + session envelope

### A. The catalog grows from user demand again (spec §4a's original tiered design, wired onto the new path)
`pointsByWorkId` with zero published rows for a known work now returns an **L1 preview with `partial:true`** and schedules the full `ingestWork` in the background — so the first user asking about an un-catalogued new-season anime sees points immediately instead of "no points forever". Flow: guard → **atomic claim** (single `INSERT..ON CONFLICT..WHERE` singleflight on the existing `ingest_jobs` table — running 15min / failure 1h / genuinely-empty `not_found` 7d / stale-claim reclaim; **no migration needed**) → published re-read (heals the TOCTOU) → preview (**claim winner only** — a burst of N cold requests costs ONE Anitabi call, not N) → `waitUntil(ingest)`. This activates the `partial` contract that was plumbed end-to-end but dormant since 1c.

### B. Session snapshot: one envelope copy instead of ~20
`session_state_v2` moved from *every retained interaction* to **one top-level envelope key** (write amplification on heavy `plan_multi` sessions: up to 20 copies × 8 refs × 500 rows → 1 copy). Read is envelope-first with a legacy fallback; a malformed envelope never resurrects stale state; interactions are pure history again.

### Dual review → 9 reconciled fixes (each reviewer caught a P1 the other missed)
- **sonnet P1**: the agent didn't handle the newly-reachable `UPSTREAM_UNAVAILABLE` on `points_by_work_id` → a typed `SearchUpstreamDown` outcome now (mirrors resolve), no more whole-turn internal-error.
- **Codex P1**: **preview rows are not routeable** (their IDs aren't published yet) → `plan_route` on a partial ref returns a typed `RoutePendingSync`, and multi-selection's **T7 PARTIAL-SYNC** terminal returns merged results without calling the route endpoint.
- Also: temporary empty states carry `partial:true` (only 7-day `not_found` is definitive — a transient failure no longer reads as "this anime has no pilgrimage spots"); ingest 404 = genuinely-empty; `work_id` constrained `^\d+$` (400, was an undefined 500); **OpenAPI regenerated** (drift CI stays green; second generator run byte-identical); envelope-preservation test; claim-fenced `markDone`/`markFailed`; `JobStore` split per the 50-line rule.
- Both reviewers **cleared the prod `waitUntil` threading** (traced to the real `ExecutionContext` at the worker entry — the sync-fallback full ingest cannot fire in prod) and the singleflight SQL as race-safe.

### Gates + one honest discovery
Agent `make lint`+`typecheck`+`test` (**1080, 87.43%**) green; catalog `oxlint`+`typecheck`+`test:worker` (**231/231, 0 unhandled errors**) green — I re-ran the workerd suite the implementer's sandbox couldn't.

⚠️ **Discovery:** the Docker **spike suite** (real-Postgres SQL tests) has been **un-runnable locally since the Wave-2 `neon-http` driver switch** — the Neon serverless driver physically can't speak to a raw local Postgres (`https://api.0.0.1/sql`), so all 9 failing files (untouched by this change) are pre-existing breakage, not a regression. Real-SQL concurrency coverage (incl. the new stale-reclaim spike test, already written) lands with the **owner-approved Neon-branch test-infra PR** (local + CI test DBs move to Neon branch-per-run, reusing the preview.yml pattern) — which also retires the `supabase start` + Docker dependency.

lifeodyssey:
## 📐 Agent 架构图(v2,`eb9fc9b` 已入库:docs/architecture-diagrams.md)

按 owner 反馈重画:**只画 agent**(5 张)、所有路由边逐字对照 `_INSTRUCTIONS` 指令表与源码、全部 Mermaid 块经 mermaid-cli 机器验证(解析+渲染)。


- [1. The agent and its direct dependencies](#1-the-agent-and-its-direct-dependencies)
- [2. One turn end to end](#2-one-turn-end-to-end)
- [3. The typed tool-outcome contract](#3-the-typed-tool-outcome-contract)
- [4. Model layer](#4-model-layer)
- [5. Session state and multi-turn selection](#5-session-state-and-multi-turn-selection)

---

## 1. The agent and its direct dependencies

The agent is a FastAPI app in a Cloudflare Container. Auth lives at the edge
Worker; the container trusts the forwarded `X-User-Id`.

```mermaid
flowchart TD
    EDGE["Cloudflare Worker - worker/entry.ts<br/>strips Authorization, injects X-User-Id<br/>routes /v1 to the container"]

    subgraph AGENT["Agent service - apps/agent"]
        API["FastAPI 0.139 + uvicorn, Python 3.13<br/>lifespan owns the catalog client and<br/>one shared model httpx client"]
        RUNTIME["RuntimeAPI - public_api.py<br/>dispatch, translation gate, persistence"]
        RUNNER["run_animichi_agent - animichi_runner.py<br/>injection preflight, usage limits,<br/>partial and blocked terminals"]
        PAI["animichi agent - pydantic-ai 2.9.1<br/>6 tools, 5 typed outputs, output validator"]
    end

    CAT["Catalog Worker - workers/catalog<br/>typed oRPC contract - packages/contract<br/>resolve, pointsByWorkId, route, geocode"]
    PG[("Supabase Postgres + PostGIS<br/>sessions, messages, routes + route_anime<br/>via asyncpg")]
    LLM["MiMo mimo-v2.5 - prod primary<br/>OpenAI SDK, header X-App-Client<br/>DeepSeek fallback dormant"]
    DDG["DuckDuckGo - web_search tool<br/>output wrapped as untrusted"]
    OBS["Logfire<br/>service animichi-runtime,<br/>environment = app_env"]

    EDGE -->|"POST /v1/runtime, /v1/runtime/stream, /v1/chat"| API
    API --> RUNTIME --> RUNNER --> PAI
    PAI -->|"oRPC over HTTP"| CAT
    RUNTIME -->|asyncpg| PG
    PAI --> LLM
    PAI --> DDG
    AGENT --> OBS
```

Endpoints (`interfaces/routes/`): `POST /v1/runtime` (JSON), `POST
/v1/runtime/stream` (SSE step events), `POST /v1/chat` (Vercel AI protocol
adapter over the same RuntimeAPI), plus read-only conversations/routes/bangumi
routes and `/healthz`.

---

## 2. One turn end to end

Dispatch order and the graceful terminals, exactly as `public_api.py` and
`animichi_runner.py` implement them.

```mermaid
sequenceDiagram
    autonumber
    participant C as Client
    participant API as RuntimeAPI
    participant R as Runner
    participant A as Agent model loop
    participant T as Tools
    participant RB as ResponseBuilder
    participant DB as Persistence

    C->>API: text, locale, session_id, optional selection fields
    API->>API: hydrate SessionState from the envelope key
    alt selected_point_ids present
        API->>API: execute_selected_route - no model
    else selected_candidate_ids present
        API->>API: dispatch by pending reason - no model<br/>anime_ambiguity to execute_multi_selection<br/>place_ambiguity to execute_place_selection
    else normal turn
        API->>R: run_animichi_agent
        R->>R: injection preflight - always logged,<br/>blocks only when ANIMICHI_INPUT_GUARD=1
        alt blocked
            R-->>API: BlockedResponseModel - zero model tokens
        else clean
            R->>A: agent.run with trusted context
            loop typed outcome loop - median 2 tool calls
                A->>T: tool call
                T-->>A: discriminated outcome - row_count plus ref,<br/>rows go to the SessionState registry
            end
            A-->>R: ONE typed output - allows_text is False
            R->>R: validate_output - anti-fabrication,<br/>clarify transition table, current-turn refs
            Note over R: UsageLimitExceeded becomes PartialResponseModel
        end
        R-->>API: AgentResult - output, intent, session_state,<br/>steps with TurnProvenance, usage
    end
    API->>RB: agent_result_to_response
    RB->>RB: project data.results and data.route from<br/>THIS turn provenance only - never registry recency
    API->>DB: envelope snapshot, messages, route + route_anime
    API-->>C: 200 for success AND graceful terminals -<br/>clarify, partial, blocked, empty, too_large.<br/>400 invalid_selection. 500 only for real faults.
```

---

## 3. The typed tool-outcome contract

The thrash fix: tools return **discriminated outcomes** computed in the data
layer; the model only branches on them. The routing edges below quote the
agent's actual instruction table (`animichi_agent.py`, `_INSTRUCTIONS`).
`str` is not in the output union, so a run cannot end in prose.

```mermaid
flowchart LR
    subgraph TOOLS["Tools - pydantic-ai Tool with timeouts"]
        RA["resolve_anime"]
        SB["search_bangumi"]
        SN["search_nearby"]
        PR["plan_route"]
        WS["web_search - QA prose only,<br/>wrapped untrusted, source tiers"]
        TT["translate_anime_title -<br/>catalog title_cn, tool-less LLM"]
    end

    subgraph OUTCOMES["Discriminated outcomes - tool_outcomes.py"]
        RES["Resolve: resolved,<br/>needs_disambiguation,<br/>not_found, upstream_unavailable"]
        SEA["Search: ok with partial flag,<br/>empty with partial flag,<br/>upstream_unavailable"]
        NEA["Nearby: ok, empty,<br/>place_ambiguity, place_unresolved,<br/>missing_location"]
        ROU["Route: ok, empty,<br/>stale_ref, pending_sync"]
    end

    subgraph OUT["Model outputs - exactly 5, no str"]
        SR["search_response"]
        RR["route_response"]
        CR["clarify_response - terminal,<br/>must echo the exact pending<br/>candidate set"]
        GR["greeting_response"]
        QR["qa_response"]
    end

    subgraph SRV["Server-only outputs - model can never emit"]
        PART["PartialResponseModel -<br/>usage limit reached"]
        BLK["BlockedResponseModel -<br/>injection preflight"]
    end

    RA --> RES
    SB --> SEA
    SN --> NEA
    PR --> ROU

    RES -->|resolved| SB
    RES -->|needs_disambiguation, not_found| CR
    RES -->|upstream_unavailable| QR
    SEA -->|ok| SR
    SEA -->|"empty - partial true means still syncing,<br/>never assert no points exist"| SR
    SEA -->|upstream_unavailable| QR
    NEA -->|"ok, empty"| SR
    NEA -->|"place_ambiguity, place_unresolved,<br/>missing_location"| CR
    ROU -->|ok| RR
    ROU -->|"stale_ref - re-run the search"| SB
    ROU -->|"pending_sync - catalog still syncing"| SR
```

The validator enforces the contract from the other side: a `search_response`
or `route_response` is rejected unless THIS turn recorded a matching Produced
step; a `clarify_response` must match the pending reason and candidate ids
exactly. `route empty` has no instructed output route - the validator simply
refuses a `route_response` for it.

---

## 4. Model layer

Verified values from `config/settings.py`, `config/model_aliases.py`,
`agents/base.py`.

```mermaid
flowchart TD
    REQ["caller model string"] --> GATE{"resolve_model_alias<br/>fullmatch a-z 0-9 _ - BEFORE lookup"}
    GATE -->|unknown or URL-like| ERR["400 ModelAliasError"]
    GATE -->|default| DEF["get_default_model"]
    GATE -->|mimo or deepseek| REG["MODEL_ALIASES registry -<br/>immutable, validated at import,<br/>duplicate effective models rejected"]

    DEF --> PRIM["PRIMARY openai mimo-v2.5<br/>base_url api.xiaomimimo.com/v1<br/>credential MIMO_API_KEY via get_settings<br/>thinking param OFF, max_retries 0"]
    DEF -.->|"fallback_agent_model is empty -<br/>MiMo-only, DeepSeek dormant"| FBK["FallbackModel mimo then deepseek -<br/>re-enable is one env line"]

    subgraph TRANSPORT["Transport"]
        CL["one shared httpx AsyncClient<br/>trust_env true, owned by the FastAPI<br/>lifespan - including the model=None path"]
        HD["default header<br/>X-App-Client animichi Prod, Staging or Dev"]
        TM["model_attempt_timeout 45s under<br/>agent_deadline 100s -<br/>validator enforces two attempts fit"]
    end

    PRIM --> CL
    REG --> CL
    CL --> HD
    CL --> TM
```

Credential routing is one policy: `xiaomimimo.com` uses `MIMO_API_KEY`,
`deepseek.com` uses `DEEPSEEK_API_KEY`, anything else uses
`OPENAI_COMPAT_API_KEY` - all read through `get_settings()`.
`validate_required_env` hard-fails at startup on whatever the resolved
default and fallback actually need.

---

## 5. Session state and multi-turn selection

```mermaid
flowchart TD
    subgraph TURN["During a turn"]
        REGY["SessionState registry<br/>search_results maps ResultRef to rows and partial flag<br/>routes maps RouteRef to ordered points<br/>LRU-tracked, evictable refs"]
        PEND["PendingClarification<br/>reason, candidate_ids,<br/>ordered_candidates, revision"]
        PROV["TurnProvenance from steps -<br/>Produced or Rejected, Search or Route -<br/>what THIS turn actually did"]
    end

    PROJ["response_builder projects data.results<br/>and data.route from provenance refs only -<br/>a prior turn ref can never leak"]
    REGY --> PROJ
    PROV --> PROJ

    subgraph SAVE["Persisted per turn"]
        ENV["session envelope - ONE key<br/>session_state_v2 full snapshot,<br/>interactions stay pure history"]
        MSGS["messages - user and assistant,<br/>including clarify, partial and blocked turns"]
        RTBL["routes table + route_anime join table"]
    end

    REGY --> ENV
    PROJ --> MSGS
    PROJ --> RTBL

    subgraph NEXTTURN["Next request"]
        HYD["hydrate - envelope first,<br/>legacy interaction scan as fallback"]
        Q1{"selected_point_ids?"}
        Q2{"selected_candidate_ids<br/>plus clarification_id?"}
        SELR["execute_selected_route - no model"]
        DISP["dispatch by pending reason - no model<br/>anime_ambiguity to execute_multi_selection<br/>place_ambiguity to execute_place_selection"]
        NORM["normal agent turn"]
    end

    ENV --> HYD --> Q1
    Q1 -->|yes| SELR
    Q1 -->|no| Q2
    Q2 -->|"yes - revision must match"| DISP
    Q2 -->|no| NORM

    DISP --> TERM["multi-selection terminals - selection.py<br/>merged route on success -<br/>empty, error, partial, too_large otherwise -<br/>typed, never a crash, pending preserved<br/>on non-success"]
```

Multi-selection semantics (`selection.py`): fetch all selected works in
parallel, merge and dedupe by point id in selection order; any partial source
short-circuits to a `partial` terminal **before** the route call (preview
point ids are never routed); over 500 points or over 50 clusters returns
`too_large` without calling the route endpoint; all-empty keeps the pending
clarification so the user can pick again.

PR #430, opened 2026-07-26
refactor(agent): centralize tool lifecycle on official pydantic-ai events

lifeodyssey:
## Problem

Per-tool progress (`StepEvent`) and history (`StepRecord`) were pushed **manually by each tool**, through four parallel mechanisms. Two consequences:

- **`web_search` and `translate_anime_title` participated in none of them** — invisible to both `AgentResult.steps` and the SSE progress stream.
- **Tool-return injection detection covered only `web_search`** (1 of ~8 tools), and was log-only.

Nothing structurally enforced participation: a new tool silently fell through, exactly as those two did.

## Approach

Pass `event_stream_handler=` to the **existing** `run()` call and translate the framework’s own `FunctionToolCallEvent` / `FunctionToolResultEvent` into the existing `StepEvent`/`StepRecord` contracts, then delete the four hand-rolled emitters.

Deliberately **not** `run_stream_events()` (it only wraps `run(event_stream_handler=…)` in a background task + context manager) and **not** the `ProcessEventStream` capability (agent-wide reusable plumbing; this needs a request-scoped translator bound to one `RuntimeDeps`). `run()` already returns the typed `AgentRunResult` the runner consumes, and its handler is fully drained before each graph node advances — which is what makes records visible to `output_validator` in time. The outer `asyncio.wait_for(...)` deadline and the typed-result assembly are untouched.

New: `agents/tool_event_bridge.py`.

## What official events cannot express — kept as-is

Function-tool events cannot represent the **server-internal geocode substep** or the **server-synthesized terminal clarification** (neither is a model tool execution), nor the two deterministic dispatch paths (`selected_point_ids` / `selected_candidate_ids`) which never touch the agent. All keep their existing mechanisms; those files have a zero diff.

Also lost from official events and reconstructed locally: `provenance` (exact registry, no session-diff heuristics), `thought`/`observation`, validated-vs-raw args.

## Review round (Fable) — three substantive fixes in `2b9073a`

**P2-1 — `success` semantics regression (user-visible).** A tool exception *recovered by the error boundary* was persisting `StepRecord(success=False)`, and since `AgentResult.success = all(step.success)`, the whole run reported failure. Concretely: the catalog blips once on `resolve_anime`, the model recovers via `web_search` and answers correctly — the client would have received `success: false` on a good answer, and `runtime.success` telemetry would skew.

Fixed by mirroring the `RetryPromptPart` treatment (emit an error **lifecycle event**, do not persist a poisoning step), and by **naming the distinction that was previously implicit**: `terminal_failures` → `recovered_exceptions`. A persisted failed step is now reserved for a genuinely terminal failure. Asserted both directions end-to-end: `result.success is True` after a recovered exception, `is False` on real terminal failure.

**P2-2 — silent no-op hid a contract violation.** `register_tool_provenance` skipped silently when `tool_call_id` was missing or non-`str`; lost provenance makes `_valid_search` reject a valid search and retry with no breadcrumb. Now diagnosed, with a test.

**P2-3 — missing regression test for the riskiest ordering case.** Drain-before-advance was proven *across* nodes but not *within* one. Added a test emitting a function tool and an output tool in the **same `ModelResponse`**, asserting provenance is available in time.

## Verification

`make lint` (329 files) · `make typecheck` (mypy strict, 112 files) · `make test` **1305 passed, coverage 88.23%** (floor 82%) — all re-run independently after rebasing onto current base (0 conflicts).

Injection detection is **broadened, not dropped**: the removed `_log_result_injections` is replaced by a central scan covering every tool return.

## ⚠️ Requires an eval baseline refresh after merge

`web_search` and `translate_anime_title` now appear in `model_initiated` steps for the first time. `tests/eval/direct_gates.py` builds trajectories from those steps and `tests/eval/official_evaluators.py` consumes their params, so evaluator inputs change. **No re-baselining was done in this PR** — deliberately, so the shift is reviewed rather than absorbed.

Implemented by Codex; independently re-verified (gates re-run, tests read, four designed-in risks checked against the diff rather than taken on trust).

### DIFF
--- a/apps/agent/agent/agents/agent_result.py
+++ b/apps/agent/agent/agents/agent_result.py
@@ -54,6 +54,7 @@ class RejectedRoute:
 StepProvenance: TypeAlias = (
     ProducedSearch | RejectedSearch | ProducedRoute | RejectedRoute
 )
+StepData: TypeAlias = dict[str, object]
 
 
 @dataclass(frozen=True)
@@ -70,8 +71,8 @@ class StepRecord:
 
     tool: str
     success: bool
-    params: dict[str, object] = field(default_factory=dict)
-    data: dict[str, object] | None = None
+    params: StepData = field(default_factory=dict)
+    data: StepData | None = None
     provenance: StepProvenance | None = None
     error: str | None = None
     model_initiated: bool = True
--- a/apps/agent/agent/agents/animichi_runner.py
+++ b/apps/agent/agent/agents/animichi_runner.py
@@ -44,7 +44,7 @@ from agent.agents.runtime_models import (
     SearchResponseModel,
 )
 from agent.agents.session_state import CurrentAnime, SessionState
-from agent.agents.tool_event_bridge import tool_event_bridge
+from agent.agents.tool_event_bridge import tool_event_bridge_for
 from agent.agents.tool_state import ToolState
 from agent.agents.web_trust import detect_prompt_injection
 from agent.clients.catalog_client import CatalogClientProtocol
@@ -204,7 +204,7 @@ async def run_animichi_agent(
             message_history=message_history or [],
             usage_limits=RUN_USAGE_LIMITS,
             usage=run_usage,
-            event_stream_handler=tool_event_bridge,
+            event_stream_handler=tool_event_bridge_for(resolved_model),
         )
     except (UsageLimitExceeded, UnexpectedModelBehavior) as error:
         return _capped_partial_result(deps, run_usage, error)
--- a/apps/agent/agent/agents/animichi_runner.py
+++ b/apps/agent/agent/agents/animichi_runner.py
@@ -44,6 +44,7 @@ from agent.agents.runtime_models import (
     SearchResponseModel,
 )
 from agent.agents.session_state import CurrentAnime, SessionState
+from agent.agents.tool_event_bridge import tool_event_bridge
 from agent.agents.tool_state import ToolState
 from agent.agents.web_trust import detect_prompt_injection
 from agent.clients.catalog_client import CatalogClientProtocol
@@ -203,6 +204,7 @@ async def run_animichi_agent(
             message_history=message_history or [],
             usage_limits=RUN_USAGE_LIMITS,
             usage=run_usage,
+            event_stream_handler=tool_event_bridge,
         )
     except (UsageLimitExceeded, UnexpectedModelBehavior) as error:
         return _capped_partial_result(deps, run_usage, error)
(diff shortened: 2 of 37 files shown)


lifeodyssey:
## CI triage: one real regression, one class of noise

Two red checks on this PR, and they are **not** the same kind of problem.

### 1. `Python integration (Neon)` — a genuine regression in this PR

```
test_conversation_hydration_contract.py::test_persisted_greeting_response_hydrates_correctly
  assert result.intent == "greet_user"
  AssertionError: assert 'error' == 'greet_user'
  log: animichi_agent_complete  intent=error steps=0
```

This is ours, with high confidence:

- That test's own docstring says **"Uses FunctionModel (no real LLM) + real testcontainer DB"** — it is deterministic, with no model API or network in the loop, so it cannot be flakiness.
- `git log` on the file shows its last two commits came from unrelated PRs — **this branch never touched it**.
- Its sibling in the same file (`test_persisted_search_response_hydrates_correctly`, which *does* involve tool calls) **passed**. Only the **zero-tool-call** path broke.
- Control: PR #433, on the same base and the same CI, has `Python integration (Neon)` **green** — so the environment is fine.

**Diagnosis:** `greet_user` is not a function tool, it is an `output_type` member, so a greeting run emits **no** `FunctionToolCallEvent`/`FunctionToolResultEvent` at all. `intent=error` with `steps=0` means the run raised and the error boundary converted it — i.e. something on the new `event_stream_handler` path most likely raises when a run produces zero tool events. Every other zero-tool-call path (plain QA, a tool-free clarify) is presumably affected the same way.

This is the blind spot the change's own test suite had: all of its coverage is about *how tool events become steps*, and nobody asked what happens when there are **no** events. Fix in progress; it must be a general fix for zero-tool-call runs, not a greeting special case.

Worth noting: the L0 smoke gate — first time it has ever guarded a real behaviour change — is what surfaced this.

### 2. `Agent Eval (L0 smoke)` — 2/80 cases errored, both transport-level

```
A2_zh_003  轻音少女的圣地在哪         → ReadError
A5_en_001  Gundam anime pilgrimage    → ReadTimeout
```

Both are HTTP failures talking to the model provider — unrelated to this diff. The deterministic thrash gate in the same run passed cleanly (every case `repeats=0`, `request_p95=5`).

This exposes a flaw in the gate itself, not in this PR: `EVAL_SMOKE`'s zero-errored rule does not distinguish "the agent broke" from "the network hiccuped", and with 80 live model calls per run the odds of at least one blip are not small. Filed as #434. **Not a blocker here** — but note the irony that the transport noise sat immediately next to the genuine regression above, which is precisely why that distinction is worth building.

### Also standing red
`Security / Semgrep` and `Security / osv-scanner` are the repo's pre-existing failures, unrelated to this diff.

Issue #426, opened 2026-07-26
bug(web): production build hydration-crashes on EVERY route — `__name is not defined` blanks the page (surfaced by the 404 browser test)


lifeodyssey:
## Symptom

The `Web app CI` job's final step — **`Browser test (branded 404 on built worker)`** — fails on every PR touching `apps/web`, while every other step in the same job passes:

```
✓ Build + output-layout integration test
✓ Type check (tsc)
✓ Oxlint (type-aware, strict)
✓ Unit tests (vitest, coverage gate)
✓ Install Playwright browser
X Browser test (branded 404 on built worker)     ← only failure
```

## Reproduced on three unrelated PRs

| PR | Scope of the diff | Touched 404/worker/routing? |
|---|---|---|
| #415 | S1.2 turn-waiting ritual | no |
| #420 | D1-D9 chat fallback states | no |
| #425 | tool step badge i18n | no |

Three PRs with completely disjoint diffs, none touching `e2e/web-404.spec.ts`, the worker, or routing — same step, same failure. It is pre-existing infrastructure breakage, not caused by any of them. Latest instance: [run 30191298126, job 89764927052](https://github.com/lifeodyssey/animichi/actions/runs/30191298126/job/89764927052).

## Why this is worth fixing now, not later

`Web app CI` is not a required check, so PRs merge anyway — which is exactly the problem. **A check that is always red trains everyone to ignore it**, and then a real regression in build/tsc/lint/unit-tests (all of which live in this same job and currently pass) would be indistinguishable from the standing noise.

We just lived through the identical failure mode on the agent side: the L0 smoke eval gate was red on every `agent_behavior` PR for a configuration reason (a missing `SUPABASE_DB_URL` binding, fixed in #424), which made the gate worthless until someone traced why. Same shape, different job.

## What is known vs. not known

**Known:**
- The failing test is `e2e/web-404.spec.ts` — `undefined route renders a branded 404`: `page.goto("/this-route-does-not-exist")`, then asserts `status() === 404`, a `404` heading, and the `Animichi` brand text.
- It runs against the **built worker** (not the dev server), after `Install Playwright browser`.
- All non-browser steps in the job pass, so this is not a build/type/lint/unit regression.

**Not known (needs investigation):**
[...]

PR #431, opened 2026-07-26
fix(web): per-env keepNames guard + retried tool steps instead of error

lifeodyssey:
Two small frontend changes, one PR.

## A — close the hole in the #426 regression guard

The `__name(` assertion in `apps/web/tests/integration/build-output.test.ts` only
bundled the **top-level** wrangler config. `keep_names: false` inherits into every
env today, but a per-env `keep_names: true` (or a per-env `build` key) under
`env.production` would have shipped the hydration crash with CI green.

The bundle assertion is now parameterized over targets **derived from
`wrangler.jsonc`'s own `env` keys**, so a newly added env is covered automatically.
All three named envs ship, so all three are covered:

| target | deployed by |
|---|---|
| top-level | inherited base for all of the below |
| `staging` | `ci.yml` → `_deploy-component.yml` (`environment: staging`) |
| `production` | `ci.yml` → `_deploy-component.yml` (`environment: production`) |
| `preview` | `preview.yml` (`wrangler deploy` / `versions upload --env preview`) |

Runtime is unchanged: `deploy --dry-run --outdir` **is** a dry-run, so the three
standalone dry-run tests were folded into the parameterized one — 4 wrangler
invocations before, 4 after (integration suite: 36s → 25s locally).

**Red-verify:** with `"keep_names": true` temporarily added under `env.production`,
`× dry-runs production without esbuild keepNames wrappers` fails and the other 6
tests pass — i.e. it fails *for that env only*, which the old top-level-only
assertion never would have caught. Sabotage reverted; `wrangler.jsonc` is
unchanged in this diff.

## B — a recoverable retry should not look like a failure

**What a user sees today:** `stepStatus()` maps `output-error` → `"error"`, and
`.chat-step[data-status="error"]` paints the badge with `--color-error-strong`.
`MessageList` renders every tool part it receives, keyed by `toolCallId`. So once
#430 lands, a repeat-guard deflection renders a red "聖地をさがしてるよ…" badge that
stays visible for the rest of the turn — including inside the collapsed settled
footprint — even though the retry succeeded and the answer is fine. Verified
empirically: with supersession disabled, the SSE-level test observes
`['error', 'done']`.

**Decision: render the superseded step as `retried` (muted + line-through), not
error.** Reasoning:

- *Leave as-is* misreports state: the turn succeeded; the only durable signal the
  user keeps is a red badge. Error styling is a scarce channel — spending it on a
  self-healed internal retry trains users to ignore it.
- *Suppress the step entirely* is the other extreme: it hides that the agent did
  extra work (relevant for latency perception and for anyone reading the footprint
  as a trace), and popping a rendered badge out mid-stream causes layout shift.
- *Retried* keeps the trace honest and subordinate. `--color-error-strong` stays
  reserved for failures the user actually has to act on.

Superseded = an errored step whose **tool type** appears again later in the same
turn (a `ModelRetry` re-issues the same tool under a fresh call id). A failure that
is never retried keeps `error`; if every retry also fails, the last one keeps
`error`.

Style uses existing tokens only (`--color-muted-fg`), and since the distinction is
otherwise colour/decoration-only the retried badge carries an `aria-label` with a
localized "retried" suffix (ja/zh/en).

**No dependency on #430.** This is driven entirely from the frontend: the new
`chatStreamRetryHandler` derives the retry part sequence (`tool-output-error` on
the original call id, then `tool-input-start` / `input-available` /
`output-available` on a fresh id) from the **real** recorded `search.sse` capture —
same technique the existing D-state handlers use — because no recording of the
post-#430 shape exists yet. When #430 lands and a real capture is recorded, that
handler can be swapped for the recording with no production change.

**Tests** (all new, all red before the change):
- `tool-steps.test.ts` — supersession matrix (retry running / retry succeeded /
  terminal error / all-retries-failed / different tool / step identity preserved).
- `message-list.test.tsx` — mid-stream `output-error` + in-flight retry →
  `['retried','running']`; error-then-success → `['retried','done']`.
- `chat-page-stream.test.tsx` — full SSE turn: error-then-success leaves no
  `[data-status="error"]` anywhere on a successful answer.
- `chat-design-css.test.ts` — retried uses the muted token; error keeps
  `--color-error-strong`.

## Verification (actual output)

```
pnpm run lint:oxlint      → exit 0, no warnings
pnpm exec tsc --noEmit    → TypeScript: No errors found
pnpm run test             → Test Files 135 passed | Tests 768 passed
                            coverage 99.04 stmts / 95.43 branch / 99.59 funcs / 99.85 lines (floor 90)
pnpm run test:integration → Test Files 1 passed | Tests 7 passed (24.98s)
```

Coverage floors untouched (not lowered). `apps/agent/**` untouched.

### DIFF
--- a/apps/web/src/features/chat/components/MessageList.tsx
+++ b/apps/web/src/features/chat/components/MessageList.tsx
@@ -1,6 +1,7 @@
 import type { ChatStatus, UIMessage } from "ai";
 import type { ChatDict } from "../i18n";
 import { formatElapsed } from "../telemetry";
+import { statusedSteps } from "../tool-steps";
 import { DataPartCard } from "./DataPartCard";
 import { SettledFootprint } from "./SettledFootprint";
 import { ToolStepBadge } from "./ToolStepBadge";
@@ -24,7 +25,9 @@ function partKey(messageId: string, part: Part, index: number): string {
 }
 
 function ToolBadges({ parts, dict }: Readonly<{ parts: readonly ToolPart[]; dict: ChatDict }>) {
-  return parts.map((part) => <ToolStepBadge key={part.toolCallId} type={part.type} state={part.state} dict={dict} />);
+  return statusedSteps(parts).map(({ step, status }) => (
+    <ToolStepBadge key={step.toolCallId} type={step.type} status={status} dict={dict} />
+  ));
 }
 
 type PipelineProps = Readonly<{
--- a/apps/web/src/features/chat/components/ToolStepBadge.tsx
+++ b/apps/web/src/features/chat/components/ToolStepBadge.tsx
@@ -4,19 +4,24 @@ import type { StepStatus } from "../tool-steps";
 
 type Props = Readonly<{ type: string; status: StepStatus; dict: ChatDict }>;
 
-/** The retried style is purely visual; name the state for assistive tech too. */
-function ariaLabel(label: string, status: StepStatus, dict: ChatDict): string | undefined {
-  if (status !== "retried") return undefined;
-  return `${label} · ${dict.toolSteps.retried}`;
+/**
+ * The retried state is otherwise conveyed by colour and `line-through` alone, so it
+ * needs a text channel. `aria-label` is prohibited on generic elements (accname §4.3.1)
+ * and `role="status"` would turn every badge into a live region, so append real text
+ * that only sighted users have hidden from them.
+ */
+function RetriedNote({ status, dict }: Readonly<{ status: StepStatus; dict: ChatDict }>) {
+  if (status !== "retried") return null;
+  return <span className="chat-step__note"> {dict.toolSteps.retried}</span>;
 }
 
 export function ToolStepBadge({ type, status, dict }: Props) {
   const name = type.replace(/^tool-/, "");
   if (HIDDEN_TOOL_STEPS.has(name)) return null;
-  const label = toolStepLabel(dict, name);
   return (
-    <span className="chat-step" data-status={status} data-tool={name} aria-label={ariaLabel(label, status, dict)}>
-      {label}
+    <span className="chat-step" data-status={status} data-tool={name}>
+      {toolStepLabel(dict, name)}
+      <RetriedNote status={status} dict={dict} />
     </span>
   );
 }
--- a/apps/web/src/features/chat/components/ToolStepBadge.tsx
+++ b/apps/web/src/features/chat/components/ToolStepBadge.tsx
@@ -1,23 +1,22 @@
 import { HIDDEN_TOOL_STEPS, toolStepLabel } from "../i18n";
 import type { ChatDict } from "../i18n";
+import type { StepStatus } from "../tool-steps";
 
-export type StepStatus = "done" | "error" | "running";
+type Props = Readonly<{ type: string; status: StepStatus; dict: ChatDict }>;
 
-/** Tool-part `state` machine → badge status (spec S1.1: step badges ← tool parts). */
-export function stepStatus(state: string): StepStatus {
-  if (state === "output-available") return "done";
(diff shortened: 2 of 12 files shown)


lifeodyssey:
## Known limitation of the supersession heuristic (accepted for now, tracked)

Qodo flagged this and I agree — recording it explicitly so it is a known trade-off rather than a hidden assumption.

The rule is in `apps/web/src/features/chat/tool-steps.ts`:

```ts
if (status !== "error") return status;
return later.some((next) => next.type === step.type) ? "retried" : "error";
```

An errored step is demoted to `retried` whenever **any** later step of the same tool type appears in the turn — the heuristic does not (and cannot, from the frontend alone) distinguish a `ModelRetry` re-issue from a legitimate second call of the same tool.

**Concrete misclassification:** `search_bangumi` genuinely fails (catalog outage) → the agent changes strategy and calls `search_bangumi` again with different arguments → the first, real failure renders muted "retried" instead of `error`.

**Why it is acceptable to land as-is:**
- It only affects badge styling, never correctness or the answer.
- The failure mode is *under*-alarming a mid-turn step, in a turn that has other visible signals; a turn that ends badly still shows `error` on the last step (nothing later supersedes it).
- The alternative available today — leaving every retried step red — spends the scarce alarm channel on a self-healed internal event on turns that succeed, which is the worse default.

**Upgrade path (after #430 lands):** the backend closes a retried tool part as `tool-output-error` and re-issues under a fresh call id, so it *knows* the difference. Emitting a first-class `superseded` / `retried` signal on that part would let the UI drop the heuristic entirely. Filing this against the frontend follow-up rather than blocking here, since #430 is still in review and this PR was deliberately built not to depend on it.

lifeodyssey:
Both P3s fixed in b26a06cb (third commit, kept separate so the revert story stays clean).

**1. `output-denied` is no longer demotable.** I took the exclude option rather than arguing the agent can never re-issue a denied tool — I cannot prove that from the frontend, and the failure mode is asymmetric: if the assumption ever breaks, the UI actively misinforms. `resolve()` now gates on `step.state !== "output-error"` rather than on the mapped status, so only a genuine ModelRetry signal can be superseded; a denial keeps its distinct state whatever follows it. New test `never demotes a denied step, even when the same tool is called again` — red-verified against the previous rule (`expected [ 'retried', 'done' ] to deeply equal [ 'error', 'done' ]`).

**2. `aria-label` dropped.** You are right that the test was certifying a guarantee the browser need not honour. I did not add `role="status"`: it carries an implicit `aria-live="polite"`, so every step badge in every turn would become a live region and the pipeline would announce itself continuously — worse than the bug. `role="img"` (there is precedent in `SceneThumb`) would work but replaces the badge's text with the label, which is a lie for a text-only element. Instead the badge appends **real text** in a visually-hidden `.chat-step__note` span (absolute / 1px / `clip-path: inset(50%)`), so the accessible name is computed from content — no role required, honoured everywhere, and the visible tool name stays first in the name.

Tests now assert the new mechanism, not the old one: `aria-label` is asserted **absent**; the note's text is asserted present and localized across all three locales; and `chat-design-css.test.ts` pins the clip rule so a future `display: none` refactor (which would drop it from the a11y tree) fails.

Untouched as instructed: same-tool-type supersession heuristic, and the two-commit split.

## Gate (actual)
```
pnpm run lint:oxlint      → exit 0, no warnings
pnpm exec tsc --noEmit    → TypeScript: No errors found
pnpm run test             → Test Files 135 passed | Tests 773 passed (was 768)
                            99.04 stmts / 95.43 branch / 99.59 funcs / 99.85 lines (floor 90, untouched)
pnpm run test:integration → Test Files 1 passed | Tests 7 passed (27.40s)
```

Head: `b26a06cb1f58f210051ef6c8fc0499d5016ff21a`

PR #444, opened 2026-07-26
docs(iter5): S1.x sprint handoff + session log entry

lifeodyssey:
Places the 2026-07-26 S1.x handoff under `docs/iterations/iter5/` per `docs/DOCS_POLICY.md` ("iteration artifacts under docs/iterations/"), and appends the matching entry to the existing `progress.md` session log rather than starting a parallel structure.

Docs only — no code, no config, no test changes.

The handoff deliberately does **not** restate what the PRs, issues and commit messages already say. It carries the things that live nowhere else:

- the review loop that surfaced a P1 in every reviewed PR this sprint, and *why* each step matters
- traps confirmed by experience today: the coverage floor being silently overridden by a CLI flag, list-shaped merge conflicts that silently stop running a test suite, gitleaks scanning commit history rather than the final tree, `pnpm build` not running `tsc`
- **four decisions still open for the owner**, each with my recommendation and reasoning — notably that production `ANON_ACCESS_ENABLED` is currently `false` on purpose, and that the Turnstile secret should be rotated

Related: #430 #435 #436 #438 #439 #440 #442 merged this sprint; #437 #441 #443 filed.

Issue #281, opened 2026-07-07
S1.9 — Cloudflare Turnstile bot-abuse protection for anonymous chat


lifeodyssey:
## Releasable statement
Anonymous chat requests must pass a low-friction Cloudflare Turnstile check before reaching the backend, protecting the newly opened anonymous surface from abuse. The check is verified server-side at the edge before any request is forwarded.

## AC summary
- AC count: 4
- Test type distribution: unit=2, integration=2, browser=0, eval=0, api=0, manual/ops=0

## Dependencies
S1.8.

## Spec source
`docs/superpowers/specs/2026-07-06-frontend-rebuild/iter-1.md` (commit 6fc13b8)

### S1.9 Cloudflare Turnstile

**User story**: As a site operator, I want anonymous chat requests to pass a Cloudflare Turnstile check before reaching the container, so the newly opened anonymous surface isn't abused.

**Design basis**: no visual canvas; a G7 supporting mechanism.

**Releasable statement**: an anonymous user completes one low-friction Turnstile check before their first message; the edge Worker verifies the token server-side before forwarding to the container.

**AC**:
- An anonymous user who completes a normal Turnstile check has their message delivered to the agent normally -> integration
- Empty: an anonymous user isn't re-challenged on every message within the same short-lived token window -> unit
- Error: an invalid/expired token is rejected by the edge Worker with a retryable prompt, and is not forwarded to the container -> integration
- i18n: Turnstile-related retry/error copy renders in ja/zh/en -> unit

**Files changed**: `worker/turnstile.ts` (new, the siteverify call), `worker/app.ts` (wiring), `apps/web/src/components/chat/TurnstileGate.tsx`.

**Dependencies**: S1.8.

lifeodyssey:
证据核验通过,关闭 —— 注意本卡范围是"建门",当前状态是 DORMANT(有意如此)。

交付 PR: #436 (feat(#281): Cloudflare Turnstile edge gate (dormant pending #274)),merged 2026-07-26 13:16 UTC 进 feat/frontend-rebuild(非 main)。

核心 AC 抽查(origin/feat/frontend-rebuild 树):
- siteverify 服务端校验 + 短窗缓存:`worker/turnstile.ts`(`createTurnstileGate()`/`guardTurnstile()`)+ `worker/turnstile.test.ts`(AC1-AC3 逐条覆盖:正常校验通过、窗口内不重复挑战、无效/过期 token 403 且不转发容器)
[...]

Issue #274, opened 2026-07-07
S1.8 — Anonymous access, edge rate limiting, and daily-budget circuit breaker


lifeodyssey:
## Releasable statement
Anonymous visitors can complete a full chat planning round-trip without logging in. The edge layer rate-limits requests by anonymous identity, and once a configured global daily cost threshold is exceeded, new anonymous requests are rejected and guided toward login rather than failing silently.

## AC summary
- AC count: 6
- Test type distribution: unit=3, integration=1, browser=1, eval=0, api=1, manual/ops=0

## Dependencies
S1.1; S0.3 (worker CI wiring).

## Spec source
`docs/superpowers/specs/2026-07-06-frontend-rebuild/iter-1.md` (commit 6fc13b8)

### S1.8 Anonymous opt-in + edge rate limiting + a global daily-budget circuit breaker (X4) + the auth-model change (X5)

**User story**: As an anonymous visitor, I want to be able to fully use chat (search/plan/refine) without logging in, with this open surface protected by rate limiting; I want to know that if the global daily cost runs away, the product will gracefully fall back to the login wall instead of burning money without limit or quietly dying.

**Design basis**: `user-journey.md` §3.3, the login-wall section (the scope that stays login-free); inputs G7/X4/X5.

**Releasable statement**: any anonymous visitor can complete one full chat planning round-trip without being asked to log in (until they tap 「保存する」); the edge Worker rate-limits requests by anonymous identity; when the global daily cost (configured via an env value) is exceeded, new anonymous chat requests are rejected and guided toward login, rather than failing silently.

[...]

PR #448, opened 2026-07-28
fix: 401 an expired/invalid JWT instead of falling through to the anonymous handler

### LINKED ISSUE #441: Expired/invalid JWT falls through to the anonymous handler instead of returning 401
## Problem

`worker/app.ts` (the `/v1/*` handler, ~lines 196-203) treats **every** authentication failure as "this caller is anonymous":

```ts
const auth = await authenticate(c.req.raw, c.env, c.executionCtx);
if (auth.ok) {
  return forwardV1(c.env, c.req.raw, { userId: auth.userId, userType: auth.userType });
}
const anonymous = isAnonymousV1(pathname)
  ? await handleAnonymousV1(c.env, c.req.raw, Date.now())
  : null;
if (anonymous !== null) return anonymous;
return c.json({ error: { code: "unauthorized", ... } }, 401);
```

`authenticate` returns `ok: false` both when there is no credential at all *and* when a
credential was presented but is expired, malformed, or signed by the wrong key. Since
#274 opened `/v1/chat` to anonymous callers, the second case now silently falls through
to `handleAnonymousV1` instead of returning 401.

## Failure scenario

A logged-in user's access token expires mid-session. Their next `/v1/chat` turn still
carries the stale `Authorization: Bearer …` header. Instead of the 401 the client was
built to handle, the edge mints/reuses an `aid` cookie and serves the turn as
`anon_<hex>`:

- the turn is charged to the **anonymous** daily-dollar pool (`daily_usage` scope `anon`),
  so a signed-in user's spend can trip the breaker for every anonymous visitor
- the turn is metered under the **anonymous** burst limiter rather than the user's
- the user's own history/session ownership is not applied

When a `session_id` is attached, `validate_session_owner` catches the mismatch and
raises 404. **On a fresh chat there is no signal at all** — the turn just succeeds as a
different identity, and the client never learns the token needs refreshing, so it never
enters the 401 → D8 → refresh path that exists precisely for this.

## Candidate designs

1. **401 on a malformed/expired bearer.** Distinguish "no credential" from "bad
   credential" in the `AuthResult` (e.g. `reason: "absent" | "invalid"`) and only fall
   through to the anonymous handler for `absent`. An `invalid` credential returns 401,
   restoring the refresh path. Most faithful to intent; requires threading a reason out
   of `authenticate`.
2. **Anonymous fallback only when no `Authorization` header is present at all.** Cheaper:
   gate `handleAnonymousV1` on `c.req.header("Authorization") === undefined` (plus the
   equivalent cookie/credential checks). No change to `authenticate`'s contract, but it
   is a proxy for the real condition and would 401 a request that carries an unrelated
   `Authorization` header.

Either way the anonymous path must stay reachable for genuinely credential-less
visitors, and `/v1/chat`'s status codes must not otherwise change — #281's Turnstile gate
and the web client's `D8`/`D11` classification both depend on the current contract.

## Origin

Raised as **P3-7** in the Fable review of #438 (card #274). Recorded, not fixed, in that
PR — behaviour is unchanged there.

lifeodyssey:
Fixes #441

## Root cause

`authenticate()` collapsed two different outcomes into one `{ ok: false }`: "no credential was
presented" and "a credential was presented and failed to verify". Since #274 opened `/v1/chat` to
anonymous callers, `worker/app.ts` treated both as *anonymous*, so an expired/malformed/wrong-key
bearer token was silently demoted to a freshly minted `anon_<hex>` identity — charged to the
anonymous daily-dollar pool, metered under the anonymous burst limiter, and never surfacing the
401 the web client needs to enter its refresh path.

## Change

**Edge (`worker/`)** — candidate design 1 from the issue. `AuthResult`'s failure arm now carries
`reason: "absent" | "invalid"`, and the `/v1/*` handler only falls through to `handleAnonymousV1`
on `"absent"`. A non-Bearer `Authorization` scheme stays `"absent"` so unrelated headers do not
start 401ing (the downside the issue flagged for candidate 2). Everything else about the contract
is untouched: anonymous access, the burst limiter, the budget breaker, and #281's dormant
Turnstile gate all keep their current status codes.

**Container ingress (`apps/agent`, authoritative)** — defense in depth. The edge strips
`Authorization` before stamping an identity, so an `X-User-Type: anonymous` arriving *with* a
credential can only mean the credential failed and the request was demoted anyway, or the
container was reached directly. `_get_trusted_auth_context` now 401s that combination with the
typed `authentication_error` code instead of serving the turn on the anonymous meter.

## Tests

New `worker/authFallthrough.test.ts` (10 tests, registered in `test:worker` and `worker/tsconfig.json`):

- expired bearer JWT → `reason: "invalid"`
- bearer JWT signed by an untrusted key → `"invalid"`
- malformed bearer token → `"invalid"`
- unknown `sk_` API key → `"invalid"`
- absent `Authorization` header → `"absent"`
- non-Bearer scheme → `"absent"`
- `Bearer` with no token → `"absent"` (header values arrive trimmed)
- invalid credential → 401, container never reached, no `Set-Cookie`
- invalid credential 401 body carries `error.code == "unauthorized"`
- absent credential → still reaches the container as `X-User-Type: anonymous`

New `apps/agent/agent/tests/unit/test_ingress_anonymous_credential.py` (5 tests): anonymous stamp +
credential → 401 / runtime never invoked / `authentication_error` code; credential-free anonymous
turn still 200; logged-in stamp unaffected by a lingering credential.

Existing `worker/auth.test.ts`, `auth-neon.test.ts`, `anonymous.test.ts`, `entry.test.ts` updated to
the new `AuthResult` shape.

## Mutation verification

Every new guard was mutated and the covering test went red, then the mutation was reverted:

| # | Mutation | Result |
|---|---|---|
| M1 | drop `auth.reason === "absent" &&` in `app.ts` | 2 failed |
| M2 | missing header returns `INVALID` instead of `ABSENT` | 3 failed |
| M3 | `verifySupabase` failure returns `ABSENT` | 2 failed |
| M4 | unknown API key returns `ABSENT` | 1 failed |
| M5 | neuter the ingress `if` (`if False and …`) | red |
| M6 | drop the `credential is not None` half | red |
| M7 | drop the `user_type == ANONYMOUS_USER_TYPE` half | red |

Restored tree: 10/10 worker fallthrough tests and 5/5 ingress tests pass.

## Gates

- `pnpm run test:worker` — 107 passed, 0 failed
- `tsc --noEmit -p worker/tsconfig.json` — clean; `oxlint worker/` reports nothing in the new file
- `make lint` (ruff check + format) — clean
- `make typecheck` (mypy strict) — no issues in 119 source files
- `apps/agent` unit tests — **1402 passed**, coverage **88.51 %** (floor 87 %)

No suppressions added; no `Any`; new helpers are within 1-10-50; both new test files are under 200 lines.

---

# Review round 1 (second seat, `request_changes`) — addressed

## P1-1 · case-sensitive `Bearer` check (real, additional bypass)

Confirmed empirically before fixing — `bearer <expired jwt>`, `BEARER …`, `BeArEr …` and
`Bearer<TAB>…` all returned `absent` and were silently handed a fresh anonymous identity, i.e. the
very class #441 closes, still open through a different door.

RFC 7235 §2.1 makes the auth-scheme token case-insensitive, so scheme detection is now
`/^bearer[ \t]+/i` with the token sliced by the match length. Deliberate boundaries: a non-bearer
scheme (`Basic`) and a scheme that merely *starts* with the word (`Bearerish`) stay `absent`, so the
"unrelated Authorization header" false-401 the issue warned about is still avoided. HTAB is accepted
as a separator because a client sending `Bearer<TAB><jwt>` is unambiguously presenting a bearer
credential.

New `worker/authScheme.test.ts` — 7 tests: lowercase, uppercase, mixed case, TAB separator, multiple
spaces, `Bearerish` (stays absent), whitespace-only token (stays absent). All 5 bypass cases were
verified red before the fix.

## P1-2 · CI was not actually running this PR's suites — **fixed, and green**

Root cause: `dorny/paths-filter` resolves a PR's changed files through the GitHub API
(`listFiles`), which requires `pull-requests: read`. `ci.yml` declares a workflow-level
`permissions: contents: read`, which grants only Contents + Metadata, so on this private repo the
call returned `Resource not accessible by integration`; the `changes` job failed and **every**
downstream lane reported `skipping`. Not transient — reproduced across a rerun and two fresh pushes.

Fix: an explicit `permissions: {contents: read, pull-requests: read}` on the `changes` job.
Verified on run `30370192030`:

| Job | Before | After |
|---|---|---|
| `changes` | failure | **success** |
| `Edge worker tests` | skipping | **success** |
| `Agent CI / Python CI` | skipping | **success** |

`Contract OpenAPI drift` still skips — correctly, this PR touches no contract.

**Separately: three security jobs (`gitleaks`, `osv-scanner`, `Semgrep`) fail with the same
`Resource not accessible by integration`, on every PR including the docs-only #444.** Same class,
but `_security.yml` is a *reusable* workflow whose permissions cannot exceed the caller's, so it
needs a coordinated two-file change validated on a throwaway branch — out of scope here and tracked
in **#455**.

## P2-1 · the invalid path is no longer silent

Both tiers now emit a structured, credential-free record: the edge logs
`{"event":"edge_auth_invalid_credential","path":…}` and the container logs
`anonymous_credential_rejected`. Neither records the token, the header, or the identity. Both are
covered by a test, including an assertion that the credential does not appear in the log entry.

## P2-2 · `ABSENT` / `INVALID` are `Object.freeze`d

They are isolate-wide singletons shared by every request; a stray mutation would poison the verdict
for all of them.

## P2-3 · JWKS-outage availability trade-off — documented, follow-up opened

A JWKS fetch failure is currently indistinguishable from a rejected credential: both return
`INVALID` → hard 401. During a JWKS outage a client whose D8 path clears the token and refreshes
would loop, because the fresh token is equally unverifiable.

This PR keeps failing **closed** deliberately — failing open is what #441 is about, and splitting
the two requires reworking the jose error handling and choosing a response for the unverifiable
case (503 + `Retry-After`?), which is a design decision rather than a bug fix. Tracked in **#452**.

## P3 · addressed / deferred

- The empty-token guard is restored as an explicit `if (!token) return ABSENT;`, and the comment no
  longer over-claims what header trimming does. It is genuinely reachable (`Bearer \x0B` survives
  HTTP trimming, which strips only SP/HTAB, but not JS `trim()`) and is covered by a test.
- The two 401 codes (`unauthorized` at the edge vs `authentication_error` in the container) are a
  public contract change that needs the web client updated in the same PR — tracked in **#453**.

## Mutation verification, round 2

| # | Mutation | Result |
|---|---|---|
| M8 | revert scheme regex to case-sensitive `/^Bearer[ ]/` | 4 red |
| M9 | drop the token-less `ABSENT` guard | 1 red |
| M10 | drop the edge `logInvalidCredential` call | 1 red |
| M11 | drop the container `_logger.warning` | 1 red |

Restored tree: 18/18 worker auth tests green.

## Gates, round 2

- CI: `Edge worker tests` **success**, `Agent CI / Python CI` **success** (run `30370192030`)
- local `pnpm run test:worker` — **115 passed**
- `tsc --noEmit -p worker/tsconfig.json` — clean; `oxlint worker/` adds no new finding (the single
  `auth.ts` template-literal error is pre-existing on the base commit)
- `actionlint .github/workflows/ci.yml` — clean
- `make lint` / `make typecheck` — clean
- agent unit tests — **1403 passed**, coverage 88.5 % (floor 87 %)

Follow-ups opened: **#452** (JWKS availability), **#453** (401 code unification), **#455** (CI security lane).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,15 @@ jobs:
 
   changes:
     runs-on: ubuntu-latest
+    # dorny/paths-filter resolves a pull_request's changed files through the
+    # GitHub API (listFiles), which needs `pull-requests: read`. The workflow-level
+    # `permissions: contents: read` grants only Contents+Metadata, so on a private
+    # repo the call returns "Resource not accessible by integration"; this job then
+    # fails and EVERY downstream lane (`Agent CI`, `Edge worker tests`, `Contract
+    # OpenAPI drift`, …) reports `skipping` — a green-looking PR with no suite run.
+    permissions:
+      contents: read
+      pull-requests: read
     outputs:
       catalog: ${{ steps.f.outputs.catalog }}
       users: ${{ steps.f.outputs.users }}
--- a/apps/agent/agent/interfaces/routes/_deps.py
+++ b/apps/agent/agent/interfaces/routes/_deps.py
@@ -104,8 +104,12 @@ def _reject_credentialed_anonymous(
     Serving it anonymously would meter and rate-limit the turn under the wrong
     identity and hide the expiry from a client built to refresh on 401.
     """
-    if user_type == ANONYMOUS_USER_TYPE and credential is not None:
-        raise HTTPException(status_code=401, detail="Valid credentials required.")
+    if user_type != ANONYMOUS_USER_TYPE or credential is None:
+        return
+    # #441 surfaced only through anomalous anonymous spend; its inverse must not
+    # be equally invisible. The credential itself is never recorded.
+    _logger.warning("anonymous_credential_rejected")
+    raise HTTPException(status_code=401, detail="Valid credentials required.")
 
 
 def _get_trusted_auth_context(
--- a/apps/agent/agent/interfaces/routes/_deps.py
+++ b/apps/agent/agent/interfaces/routes/_deps.py
@@ -19,6 +19,7 @@ from agent.infrastructure.session import SessionStore, create_session_store
 from agent.infrastructure.supabase.client import SupabaseClient
 from agent.interfaces.public_api import PublicAPIResponse, RuntimeAPI
 from agent.interfaces.schemas import GRACEFUL_TERMINAL_STATUSES
+from agent.interfaces.usage_metering import ANONYMOUS_USER_TYPE
 
 if TYPE_CHECKING:
     import logfire
@@ -92,13 +93,31 @@ def _normalize_optional_header(value: str | None) -> str | None:
     return text or None
 
 
+def _reject_credentialed_anonymous(
+    user_type: str | None, credential: str | None
+) -> None:
+    """Refuse an anonymous stamp that arrived with a credential (issue #441).
+
+    The edge strips ``Authorization`` before stamping an identity, so the pair
+    can only co-occur when a presented credential failed to verify and the
+    request was demoted anyway, or when the container was reached directly.
+    Serving it anonymously would meter and rate-limit the turn under the wrong
+    identity and hide the expiry from a client built to refresh on 401.
+    """
+    if user_type == ANONYMOUS_USER_TYPE and credential is not None:
+        raise HTTPException(status_code=401, detail="Valid credentials required.")
+
+
 def _get_trusted_auth_context(
     x_user_id: Annotated[str | None, Header(alias="X-User-Id")] = None,
     x_user_type: Annotated[str | None, Header(alias="X-User-Type")] = None,
+    authorization: Annotated[str | None, Header(alias="Authorization")] = None,
 ) -> TrustedAuthContext:
+    user_type = _normalize_optional_header(x_user_type)
+    _reject_credentialed_anonymous(user_type, _normalize_optional_header(authorization))
     return TrustedAuthContext(
         user_id=_normalize_optional_header(x_user_id),
-        user_type=_normalize_optional_header(x_user_type),
+        user_type=user_type,
     )
 
 
--- a/apps/agent/agent/tests/unit/test_ingress_anonymous_credential.py
+++ b/apps/agent/agent/tests/unit/test_ingress_anonymous_credential.py
@@ -73,6 +73,19 @@ async def test_the_rejection_carries_the_authentication_error_code() -> None:
     assert response.json()["error"]["code"] == "authentication_error"
 
 
+async def test_the_rejection_is_logged_without_the_credential() -> None:
+    from structlog import testing
+
+    app, _ = _app()
+    with testing.capture_logs() as captured:
+        await _post(app, {**ANON_HEADERS, "Authorization": STALE_JWT})
+    events = [
+        entry for entry in captured if entry["event"] == "anonymous_credential_rejected"
+    ]
+    assert len(events) == 1
+    assert STALE_JWT not in str(events[0])
+
+
 async def test_a_credential_free_anonymous_turn_still_succeeds() -> None:
     app, runtime = _app()
     response = await _post(app, ANON_HEADERS)
--- a/apps/agent/agent/tests/unit/test_ingress_anonymous_credential.py
+++ b/apps/agent/agent/tests/unit/test_ingress_anonymous_credential.py
@@ -1,0 +1,87 @@
+"""Container ingress refuses an anonymous stamp carrying a credential (#441).
+
+The edge strips ``Authorization`` before it stamps an identity, so the two can
+only arrive together when the edge fell through on a bad credential or was
+bypassed. Either way the authoritative tier must 401 rather than serve the turn
+as a fresh anonymous visitor on the anonymous meter.
+"""
+
+from __future__ import annotations
+
+from unittest.mock import AsyncMock, MagicMock
+
+import httpx
+from fastapi import FastAPI
+
+from agent.interfaces.public_api import RuntimeAPI
+from agent.interfaces.schemas import PublicAPIResponse
+from agent.tests.unit.conftest_fastapi import async_client, build_app, build_stub_db
+
+ANON_HEADERS = {
+    "X-User-Id": "anon_0123456789abcdef0123456789abcdef",
+    "X-User-Type": "anonymous",
+}
+HUMAN_HEADERS = {"X-User-Id": "user-1", "X-User-Type": "human"}
+STALE_JWT = "Bearer eyJhbGciOiJFUzI1NiJ9.stale.signature"
+
+
+def _body() -> dict[str, object]:
+    return {
+        "messages": [
+            {"id": "u1", "role": "user", "parts": [{"type": "text", "text": "京吹"}]}
+        ]
+    }
+
+
+def _app() -> tuple[FastAPI, MagicMock]:
+    db = build_stub_db()
+    db.usage = MagicMock()
+    db.usage.accumulate_usage = AsyncMock(return_value=None)
+    db.usage.total_cost_usd = AsyncMock(return_value=0.0)
+    runtime = MagicMock(spec=RuntimeAPI)
+    runtime.handle = AsyncMock(
+        return_value=PublicAPIResponse(
+            success=True, status="ok", intent="plan_route", message="ルートだよ。"
+        )
+    )
+    runtime.validate_session_owner = AsyncMock(return_value=None)
+    runtime._db = db
+    app, _ = build_app(runtime_api=runtime, db=db)
+    return app, runtime
+
+
+async def _post(app: FastAPI, headers: dict[str, str]) -> httpx.Response:
+    async with async_client(app) as client:
+        return await client.post("/v1/chat", json=_body(), headers=headers)
+
+
+async def test_an_anonymous_stamp_with_a_credential_is_rejected() -> None:
+    app, _ = _app()
+    response = await _post(app, {**ANON_HEADERS, "Authorization": STALE_JWT})
+    assert response.status_code == 401
+
+
+async def test_the_rejected_turn_never_reaches_the_runtime() -> None:
+    app, runtime = _app()
+    await _post(app, {**ANON_HEADERS, "Authorization": STALE_JWT})
+    assert runtime.handle.await_count == 0
+
+
+async def test_the_rejection_carries_the_authentication_error_code() -> None:
+    app, _ = _app()
+    response = await _post(app, {**ANON_HEADERS, "Authorization": STALE_JWT})
+    assert response.json()["error"]["code"] == "authentication_error"
+
+
+async def test_a_credential_free_anonymous_turn_still_succeeds() -> None:
+    app, runtime = _app()
+    response = await _post(app, ANON_HEADERS)
+    assert response.status_code == 200
+    assert runtime.handle.await_count == 1
+
+
+async def test_a_logged_in_stamp_is_unaffected_by_a_lingering_credential() -> None:
+    app, runtime = _app()
+    response = await _post(app, {**HUMAN_HEADERS, "Authorization": STALE_JWT})
+    assert response.status_code == 200
+    assert runtime.handle.await_count == 1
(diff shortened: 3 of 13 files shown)


lifeodyssey (commented):
## Reviewer (Fable seat) — independent review of #448

**Verdict: approve** (P2/P3 items are follow-ups or pre-existing infra; the core fix is sound, well-tested, and I reproduced every gate and two mutations independently in a detached worktree.)

### Security semantics — reverse enumeration of anonymous-reachable paths

Every path into `handleAnonymousV1` was walked; only the edge `/v1/*` non-public branch reaches it, and only on `reason === "absent"`. `AuthResult` is a closed discriminated union — every failure return in `auth.ts` (jwt-verify catch, non-string `sub`, unknown/errored API key) is `INVALID`; the only `ABSENT` returns are missing header and non-`Bearer ` prefix. No third state exists.

- **Dead-branch removal (empty `Bearer` token) is correct.** Fetch header-value normalization strips leading/trailing HTTP whitespace, so any value passing `startsWith("Bearer ")` yields a non-empty token after `slice(7).trim()`; a bare/whitespace `Bearer` normalizes to `Bearer` and lands in ABSENT. The `bearer("   ")` test pins this. Even in a hypothetical non-normalizing runtime, an empty token would flow to `verifyJwt` → INVALID → 401, which fails safe (401, not anonymous).
- **Cookie path is clean.** The `aid` cookie is a worker-minted HMAC identity, not a credential; no auth token rides cookies to the worker (Better Auth cookie is exchanged client-side in `authSession.ts`). No demotion path via cookies.
- **Container backstop holds via the edge.** `forwardV1` deletes `Authorization` whenever it stamps any identity (including anonymous), so a real anonymous user can never be falsely 401ed by `_reject_credentialed_anonymous`. All identity-consuming routes (`chat.py` via `_require_trusted_user`, `runtime.py`, `bangumi.py`, `feedback.py`) go through `_get_trusted_auth_context` — no bypass reads `X-User-Type` directly. Direct-to-container with a forged anonymous stamp and *no* credential is still trusted — that is the pre-existing trust model, unchanged and out of scope.

### Client impact (D8 / refresh)

- Classifier (`apps/web/src/lib/chat/errorClassifier.ts`) keys D8 on **status** (401/403), not error code — so the edge's `unauthorized` and the container's `authentication_error` both land in D8. D8's resume calls `clearAuthToken()` then recovers (`use-stream-recovery.ts:60`) — exactly the refresh path this fix is meant to trigger.
- Anonymous users cannot be mis-401ed: `authHeaders()` attaches `Authorization` only when a token exists; signed-out requests carry none → `absent` → anonymous handler as before.

### Findings

**P2**
1. `worker/auth.ts:149` — RFC 7235 auth schemes are case-insensitive, but `startsWith("Bearer ")` is exact-case. A client sending `bearer <expired-jwt>` is classified `absent` and still silently demoted to anonymous on `/v1/chat` — a residual instance of the #441 bug class for non-canonical clients. No privilege is gained (anonymous is obtainable credential-free), and this repo's own clients send canonical `Bearer `, so impact is UX-only; still, a case-insensitive scheme match would close it. Follow-up acceptable.
2. **CI never ran the gating suites on this PR**: the `changes` paths-filter job failed with "Resource not accessible by integration", so Edge worker tests / Agent CI were all SKIPPED, and gitleaks/osv/semgrep are red. The last two runs on `feat/frontend-rebuild` are also red — pre-existing infra, not introduced here — but please get a green (or at least executed) CI run before merge. I re-ran the gates locally as compensation (below).

**P3**
3. Multiple `Authorization` headers combine via `Headers.get()` (`"Basic x, Bearer y"` → not `Bearer `-prefixed → `absent` → anonymous). Same class as P2-1, more contrived; reverse order yields an invalid token → 401 (safe).
4. Error-code divergence for the same logical condition: edge 401 body uses `code: "unauthorized"`, container 401 uses `authentication_error`. D8 works either way (status-keyed), but one condition ↔ two wire codes is a contract smell worth unifying later.
5. `STALE_JWT` fixture in `test_ingress_anonymous_credential.py` begins with a real base64 JWT header — repo convention prefers zero-entropy fixture strings, and gitleaks' jwt rule may flag it (couldn't verify locally; the security job is already red on base).

### Independent gate re-run (detached worktree at PR head `0ddedfea`)

- `pnpm run test:worker`: **107/107 pass**; `tsc --noEmit -p worker/tsconfig.json`: clean
- `make lint`: clean; `make typecheck`: clean (119 files)
- Agent unit: **1402 passed**, coverage **88.51%** ≥ 87% floor

### Mutation spot-check (2 of the implementer's 7, independently reproduced)

- **M1** (drop `auth.reason === "absent" &&` in `app.ts`): 2 worker tests fail ✔ (matches self-report)
- **M6** (drop `credential is not None` in `_deps.py`): ingress test goes red (credential-free anonymous turn 401s) ✔

Both reverted; worktree verified clean, then removed.

### Standards

No suppressions in the diff, no `Any`, new functions within 1-10-50, both new test files < 200 lines, assertions are strong (`deepEqual` on full result shapes, container-not-reached, no `Set-Cookie`, positive controls on both tiers).

lifeodyssey (commented):
## Reviewer (Fable seat) — re-review of the fix round (HEAD 0bb64229)

**Final verdict: approve.** All delta items independently verified.

### Scheme regex boundary analysis (`worker/auth.ts` `BEARER_SCHEME = /^bearer[ \t]+/i`)

- **Bearerish-class schemes stay correctly absent**: the class requires at least one SP/HTAB immediately after `bearer`, so `Bearerish`, `Bearer2`, bare `Bearer`/`bearer` all fail to match → `absent`. Tested (`a scheme that merely starts with bearer is not our scheme`).
- **Case folding**: `bearer`/`BEARER`/`BeArEr` now report `invalid` on a bad token — the P2-1 residual downgrade is closed. Anchored, character-class-only regex — no ReDoS surface.
- **HTAB tradeoff**: RFC 7235's ABNF is strictly `1*SP`; accepting `[ \t]+` is liberal, but over-matching errs toward "credential presented" → 401, the safe failure direction (the dangerous direction is under-matching → silent anonymous). Documented in the comment. Accepted.
- **Restored empty-token guard is correct and reachable**: the `Bearer \x0B` test demonstrates the one class that survives HTTP normalization (SP/HTAB-only trimming) but empties under JS `trim()` — classifying it `absent` is right (nothing was presented) and grants nothing a header-less caller couldn't get. My earlier P3 comment-accuracy point is resolved.
- Remaining known P3 (multi-`Authorization` join → `absent`) unchanged and acceptable as documented.

### Logging — no credential leakage

- Edge: `logInvalidCredential` emits exactly `{event, path}`; the test `deepEqual`s the *entire parsed record*, so any future field addition (token, user id, header) fails the test — strong structural guarantee. The absent-path 401 is deliberately not counted; only `invalid` verdicts are.
- Container: `anonymous_credential_rejected` logged with event name only; test captures structlog and asserts the stale JWT is absent from the record.

### Independent verification (detached worktree at 0bb64229, removed after)

- `pnpm run test:worker`: **115/115** (authScheme suite included); worker `tsc --noEmit`: clean
- `make lint` / `make typecheck`: clean; agent unit: **1403 passed**
- **Independent mutations (beyond the self-reported M8–M11):**
  - drop the `i` flag → **28 tests fail** (case tests + all canonical-Bearer coverage)
  - narrow `[ \t]+` to `[ ]+` → **exactly 1 test fails** (the HTAB test)
  - both reverted, worktree verified clean
- **CI fix confirmed real**: run 30370192030 — `changes` success with `pull-requests: read`, and `Edge worker tests`, `Agent CI / Python CI`, `Python integration (Neon)` all executed and green. Remaining gitleaks/osv/semgrep reds are pre-existing on base, tracked in #455; 401-code unification in #453, JWKS follow-up in #452.

### Standards

`ABSENT`/`INVALID` frozen with rationale; `_reject_credentialed_anonymous` still within 1-10-50 after the early-return inversion; job-scoped CI permission is minimal; no suppressions, no `Any`.

Issue #434, opened 2026-07-26
ci(eval): L0 smoke gate counts transport errors as failures — a model-API hiccup turns the gate red


lifeodyssey:
## Symptom

The L0 smoke eval gate (`EVAL_SMOKE=1`, required on any PR touching `agent_behavior` paths) failed on PR #430 with:

```
2/80 cases errored (EVAL_SMOKE requires zero errors).
```

The two cases:

| Case | Query | Error |
|---|---|---|
| `A2_zh_003` | 轻音少女的圣地在哪 | **ReadError** |
| `A5_en_001` | Gundam anime pilgrimage | **ReadTimeout** |

Both are **HTTP transport failures talking to the model provider** — not agent-logic errors, not thrash, and unrelated to the code under review. The deterministic thrash gate in the same run passed cleanly (every case `repeats=0`, `request_p95=5`).

## Why this matters

`EVAL_SMOKE`'s zero-errored rule currently makes no distinction between:

1. **the agent broke** — an exception escaping tool execution or the run loop, which is exactly what this gate exists to catch; and
2. **the network hiccuped** — a `ReadError` / `ReadTimeout` / 429 / 502 from the provider, which says nothing about the PR.

With 80 live model calls per run, the probability of at least one transport blip is not small. So the gate goes red for reasons uncorrelated with code quality — and a gate that is red for unrelated reasons is the failure mode we have now hit **four separate times today**:

- The same gate was permanently red because of a missing `SUPABASE_DB_URL` binding (fixed in #424).
- `Web app CI` was red for weeks and dismissed three times as "CI noise" — it was a **real production hydration crash** (#426).
- Two unit tests flake only under load, one of them in violation of the repo's own no-timing-assertions rule (#427).
- And now this.

Each time, the cost is not the red itself — it is that **a red we have learned to ignore looks exactly like a red that matters**. On this very PR the same gate *also* surfaced a genuine regression (a zero-tool-call path returning `intent=error`), and the transport noise sat right next to it.

## Suggested direction

Classify errors before counting them:

- **Fail the gate** on agent-originated errors: unhandled exceptions in tools or the run loop, output-validation failures, anything the agent itself produced.
- **Do not fail** on provider transport errors (`httpx.ReadError`, `ReadTimeout`, `ConnectError`, and HTTP 429/502/503 from the model endpoint). Report them prominently in the summary — a rising rate is an operational signal — but they should not gate a code change.
[...]

PR #450, opened 2026-07-28
fix(eval): stop the repeat tripwire from equating unrecorded tool params (#443)

### LINKED ISSUE #443: eval: L0 repeat guard fires intermittently on search_nearby across unrelated branches (C1_en_001 / C1_en_005)
## Symptom

The L0 smoke gate's repeat guard fires on `search_nearby` intermittently, on branches whose diffs cannot plausibly affect agent behaviour. Observed today (2026-07-26):

| Branch | PR | Failure | Diff touches |
|---|---|---|---|
| `feat/274-anonymous-access` | #438 | `C1_en_005: repeated identical tool call: search_nearby` | edge auth, rate limiter, usage metering |
| `feat/eval-stats-293` | #440 | `C1_en_001: repeated identical tool call: search_nearby` | eval gate machinery only |

**#438 passed on a plain re-run of the same commit**, with no code change. So the assertion is not deterministic against the live model.

Note this is a *different* failure mode from the one tracked on PR #430 (`N/80 cases errored`, i.e. cases raising exceptions). Here the count of errored cases is **zero** — every case completed; one of them simply called `search_nearby` twice with identical arguments.

## Why it matters

The smoke gate is a required-ish PR gate. An assertion that fails on unrelated PRs and passes on re-run is the failure mode issue #434 was filed about, one layer up: it trains everyone to re-run rather than read, and once that habit exists a real regression gets re-run away too.

## Two candidate explanations, and how to tell them apart

1. **Live-model non-determinism on a genuinely ambiguous case.** The `C1_*` family is the location/nearby-search path, where a second `search_nearby` after a clarification is arguably reasonable behaviour. If so, the fix belongs in the *assertion* — either the accepted-chain list for these cases admits a repeat, or the repeat guard should key on something narrower than identical arguments.
2. **A real weak spot in the repeat guard.** If the agent is genuinely re-issuing an identical call rather than using the first result, that is thrash and the guard is doing its job — in which case the fix is in the agent, not the assertion.

**How to decide:** the case IDs are known (`C1_en_001`, `C1_en_005`). Run those two cases in isolation, several times, and look at the actual trajectory: does the second `search_nearby` carry identical arguments, and does anything between the two calls justify it? That is a bounded experiment, not a guess.

## Do not

Do not widen the repeat guard just to make CI green. If explanation (2) is right, that removes the only thing catching thrash — and thrash is what the whole #349 agent redesign was about. Decide with the trajectory in hand.

## Related
- #434 — smoke gate counts transport errors as failures (fixed in #440)
- PR #430 — a separate, still-undiagnosed `N/80 errored` failure

lifeodyssey:
Fixes #443. Follow-up: #459.

## Diagnosis — explanation (1), decided with trajectories, not reasoning

I ran `C1_en_001` / `C1_en_005` in isolation against the live MiMo model on the trajectory tier (mock catalog, `NullDatabase`), capturing every executed step *and* the runtime guard's fingerprint ledger: **50 runs** on the first tree, **32 more** on the final tree — **82 runs, 0 repeat-guard hits, 0 errors, 0 unrecorded params** (0/82 → 95% one-sided upper bound **3.6%**; the first batch alone, 0/50, bounds it at 5.8%).

The runtime `before_tool_execute` repeat-guard is intact — not bypassed, no thrash regression.

What the model actually does, in **5/82 runs (6.1%, 95% CI 2.0%–13.7%)**, is retry `search_nearby` with a **different location spelling** after place resolution fails:

```
C1_en_005 (run 22)
  geocode        {"location": "Nishinomiya, Hyogo, Japan"}
  search_nearby  {"location": "Nishinomiya, Hyogo, Japan"}
  geocode        {"location": "西宮市, 兵庫県, 日本"}
  search_nearby  {"location": "西宮市, 兵庫県, 日本"}
  geocode        {"location": "Nishinomiya"}
  search_nearby  {"location": "Nishinomiya"}

C1_en_001 (run 1)
  search_nearby  {"location": "Uji, Kyoto, Japan"}
  search_nearby  {"location": "宇治市, 京都府, 日本"}
  search_nearby  {"location": "34.8862, 135.8002"}
```

Those arguments are plainly **not identical**, and the guard correctly let them through. The gate still failed because of how the calls were *recorded*. On the tree those CI runs used (pre-#430), a rejected nearby search was recorded with its arguments dropped:

```python
# apps/agent/agent/agents/catalog_tools.py @ 53271753^
if not isinstance(resolved, tuple):
    _record(ctx.deps, ToolName.SEARCH_NEARBY.value, {}, resolved.model_dump(), ...)
                                                    ^^ location thrown away
```

`direct_gates.RecordedToolCall` compares recorded `step.params`, so two or three *different* locations all projected onto the same `("search_nearby", "{}")` identity and read as one repeated identical call. That fully explains the report: intermittent (only when the model needs a geocode retry), independent of the branch diff, and passing on a plain re-run.

So this was never thrash and never a broken guard — the tripwire was reading a lossy projection. **The guard is not widened by one inch here.**

## Change

#430 already stopped dropping the arguments on that path. This turns the two surviving `{}` substitutions in the bridge into defense-in-depth that announces itself instead of lying, and gives the resulting signal a consumer:

- `StepRecord` / `ActiveToolCall` carry `params_recorded`. The event bridge sets it `False` for an unprojectable call and for a return whose call event was never seen — both previously became a silent `{}` — and logs a structlog warning (`tool_call_params_unprojectable` / `tool_call_params_unrecorded`) so the loss is visible in production, not only in an eval line.
- `direct_gates` excludes unknown-argument calls from the identical-call comparison (unknown ≠ identical) **and fails the gate on `unrecorded_params > 0`**. On every live path the bridge projects the model's own call arguments, so that count is 0 in a healthy run — anything above 0 is a recording defect, and a blind tripwire must not pass green.
- Identical **recorded** params still trip the wire — including a genuinely no-argument near-me search repeated, which is real thrash.
- The tripwire comment in `direct_gates.py` is narrowed to what the code supports: the guard keys on **validated** args while these records carry **raw** args, so it promises only "identical *recorded* params never execute twice" and cannot see guard breakage hiding in default normalization. Closing that fork is #459.
- `StepRecord` is constructed by keyword at the bridge site (8 fields, 6 optional).

## Gates

- `make lint` pass · `make typecheck` pass (mypy strict, 119 files) · `make test` **1409 passed, coverage 88.52%** (floor 87)
- **L0 smoke, cap 80 = the CI configuration** (`EVAL_SMOKE=1 EVAL_MAX_CASES=80`, live MiMo), run on the final tree — the selection contains the reported case `C1_en_001`:
  ```
  C1_en_001: requests=2 tool_calls=1 repeats=0 unrecorded_params=0
  C1_ja_007: requests=3 tool_calls=1 repeats=0 unrecorded_params=0
  80/80 evaluated, errored=0 → All gates passed.
  ```
  No case in the run reported a nonzero `repeats` or `unrecorded_params`.
- Targeted `C1_en_001` / `C1_en_005` batch on the final tree: 32 runs, **repeats=0, unrecorded_params=0**, including 3 runs that exercised the multi-attempt nearby/geocode retry path that used to false-positive.
- Mutation verification — **10/10 mutants killed**, baseline green:

| Mutant | Result |
|---|---|
| M1 unknown-argument calls back in the repeat comparison | KILLED |
| M2 `finish()` fallback: no warning, marked as recorded | KILLED |
| M3 `ValidationError` fallback: no warning, marked as recorded | KILLED |
| M4 `StepRecord` always claims recorded params | KILLED |
| M5 repeat detection disabled (tripwire removed) | KILLED |
| M6 `unrecorded_params` always counted as 0 | KILLED |
| M7 `unrecorded_params` gate never fails | KILLED |
| M8 `unrecorded_params` gate fails unconditionally | KILLED |
| M9 no production warning on an orphan return | KILLED |
| M10 no production warning on unprojectable args | KILLED |

CI's own `agent-eval-smoke` lane still needs the `changes`-job permission fix from #448 before it can run here.

### DIFF
--- a/apps/agent/agent/agents/agent_result.py
+++ b/apps/agent/agent/agents/agent_result.py
@@ -76,6 +76,10 @@ class StepRecord:
     provenance: StepProvenance | None = None
     error: str | None = None
     model_initiated: bool = True
+    # False when the call's arguments could not be projected (#443). Empty
+    # `params` then means "unknown", not "called with no arguments" — consumers
+    # that compare calls by arguments must not equate two unknowns.
+    params_recorded: bool = True
 
 
 @dataclass
--- a/apps/agent/agent/agents/tool_event_bridge.py
+++ b/apps/agent/agent/agents/tool_event_bridge.py
@@ -40,6 +40,9 @@ class ActiveToolCall:
 
     tool: str
     params: dict[str, JsonValue]
+    # See StepRecord.params_recorded (#443): an unprojectable or never-seen call
+    # carries empty params that must not be read as "no arguments".
+    params_recorded: bool = True
 
 
 @dataclass
@@ -50,11 +53,11 @@ class ToolLifecycleRegistry:
     provenance: dict[str, StepProvenance] = field(default_factory=dict)
     recovered_exceptions: set[str] = field(default_factory=set)
 
-    def start(self, call_id: str, tool: str, params: dict[str, JsonValue]) -> None:
-        self.active[call_id] = ActiveToolCall(tool, params)
+    def start(self, call_id: str, call: ActiveToolCall) -> None:
+        self.active[call_id] = call
 
     def finish(self, call_id: str, tool: str) -> ActiveToolCall:
-        return self.active.pop(call_id, ActiveToolCall(tool, {}))
+        return self.active.pop(call_id, ActiveToolCall(tool, {}, params_recorded=False))
 
     def register_provenance(self, call_id: str, provenance: StepProvenance) -> None:
         self.provenance[call_id] = provenance
@@ -109,23 +112,26 @@ def register_recovered_tool_exception(
     ctx.deps.tool_lifecycle.mark_recovered_exception(call_id)
 
 
-def _call_params(event: FunctionToolCallEvent) -> dict[str, JsonValue]:
+def _call_params(event: FunctionToolCallEvent) -> ActiveToolCall:
     """Never let an odd argument shape kill the run — the siblings all guard too.
 
     Malformed JSON already arrives as {"INVALID_JSON": raw} from pydantic-ai, so
     the model keeps its own retry path; the residual raiser is dict-typed args
     holding non-JSON values. Recording empty params is strictly better than
-    turning a retryable tool call into a terminal run error.
+    turning a retryable tool call into a terminal run error — but it is flagged
+    as unrecorded (#443) so no consumer mistakes the loss for "no arguments".
     """
+    tool = event.part.tool_name
     try:
-        return _OBJECT.validate_python(event.part.args_as_dict())
+        return ActiveToolCall(tool, _OBJECT.validate_python(event.part.args_as_dict()))
     except ValidationError:
-        return {}
+        return ActiveToolCall(tool, {}, params_recorded=False)
 
 
 async def _handle_call(deps: RuntimeDeps, event: FunctionToolCallEvent) -> None:
-    params = _call_params(event)
-    deps.tool_lifecycle.start(event.tool_call_id, event.part.tool_name, params)
+    call = _call_params(event)
+    deps.tool_lifecycle.start(event.tool_call_id, call)
+    params = call.params
     data = cast(StepData, params)
     await _emit(
         deps, StepEvent(event.part.tool_name, event.tool_call_id, "running", data)
@@ -206,7 +212,15 @@ def _record_terminal_return(
     params = cast(StepData, call.params)
     payload = cast(StepData | None, data)
     deps.steps.append(
-        StepRecord(call.tool, success, params, payload, provenance, error)
+        StepRecord(
+            call.tool,
+            success,
+            params,
+            payload,
+            provenance,
+            error,
+            params_recorded=call.params_recorded,
+        )
     )
 
 
--- a/apps/agent/agent/agents/tool_event_bridge.py
+++ b/apps/agent/agent/agents/tool_event_bridge.py
@@ -57,7 +57,11 @@ class ToolLifecycleRegistry:
         self.active[call_id] = call
 
     def finish(self, call_id: str, tool: str) -> ActiveToolCall:
-        return self.active.pop(call_id, ActiveToolCall(tool, {}, params_recorded=False))
+        active = self.active.pop(call_id, None)
+        if active is not None:
+            return active
+        logger.warning("tool_call_params_unrecorded", tool=tool, call_id=call_id)
+        return ActiveToolCall(tool, {}, params_recorded=False)
 
     def register_provenance(self, call_id: str, provenance: StepProvenance) -> None:
         self.provenance[call_id] = provenance
@@ -125,6 +129,9 @@ def _call_params(event: FunctionToolCallEvent) -> ActiveToolCall:
     try:
         return ActiveToolCall(tool, _OBJECT.validate_python(event.part.args_as_dict()))
     except ValidationError:
+        logger.warning(
+            "tool_call_params_unprojectable", tool=tool, call_id=event.tool_call_id
+        )
         return ActiveToolCall(tool, {}, params_recorded=False)
 
 
@@ -213,12 +220,12 @@ def _record_terminal_return(
     payload = cast(StepData | None, data)
     deps.steps.append(
         StepRecord(
-            call.tool,
-            success,
-            params,
-            payload,
-            provenance,
-            error,
+            tool=call.tool,
+            success=success,
+            params=params,
+            data=payload,
+            provenance=provenance,
+            error=error,
             params_recorded=call.params_recorded,
         )
     )
--- a/apps/agent/agent/tests/eval/direct_gates.py
+++ b/apps/agent/agent/tests/eval/direct_gates.py
@@ -15,14 +15,20 @@ from agent.agents.agent_result import AgentResult
 REQUEST_LIMIT = 12
 # Calibrated 2026-07-20: legitimate multi-step disambiguation (sequel/season
 # cases added by #304B) needs 7-8 tool calls; this cap only bounds runaway
-# breadth. NOTE: the repeated-identical detector below reads EXECUTED steps —
-# the runtime repeat-guard deflects identical re-calls before execution, so
-# post-guard this detector is a regression TRIPWIRE for the guard itself
-# (it firing means the guard broke), not a live thrash signal. That inference
-# holds only while recorded params are LOSSLESS: #443 fired on two nearby
-# searches with different locations because the rejected-search recording path
-# dropped its arguments, so both projected to "{}". Calls whose arguments were
-# not recorded are therefore excluded from the comparison, never equated.
+# breadth. NOTE: the repeated-identical detector below reads EXECUTED steps.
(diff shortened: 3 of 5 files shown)


lifeodyssey (commented):
## Fable independent review — PR #450 (fix #443)

**Verdict: APPROVE.** The one question that matters — is this a disguised widening of the repeat guard — answers **no**, on structural grounds, not just on the author's word.

### 1. Is the exclusion a widening? No — verified structurally

The actual thrash defense is the runtime `before_tool_execute` hook (`block_identical_repeat`, `animichi_agent.py:493`), which fingerprints **validated args at execution time**, entirely upstream of and independent from the recording layer this PR touches. A ModelRetry-deflected repeat never executes and never produces a StepRecord (`_handle_retry` appends nothing). The eval tripwire is therefore a second-order regression wire, exactly as its comment says.

For real thrash to escape the tripwire via the new exclusion, both duplicate calls would need to (a) execute — runtime guard already broken — AND (b) independently fail argument projection. I traced every live path to `params_recorded=False`:

- `_call_params` ValidationError: model args arrive as JSON text; I verified installed pydantic-ai 2.11.0 `args_as_dict(raise_if_invalid=False)` returns `{"INVALID_JSON": raw}` for malformed JSON — projectable and faithful (identical raw → identical fingerprint → still trips). The ValidationError branch is only reachable with in-process dict args holding non-JSON Python values — not a live-model shape.
- `finish()` orphan return: a result event whose call event never arrived — a stream anomaly, not something thrash produces (pydantic-ai emits the call event before executing).

So the exclusion boundary covers exactly recording failures and nothing a genuinely repeating model can reach. Genuinely-empty-args repeats still trip (`test_genuinely_empty_arguments_still_trip_the_wire` covers the near-me case).

### 2. Flag propagation completeness — clean

All non-bridge `StepRecord` constructors (`selected_route.py:87,130`, `selection.py:184`, `step_recording.py:16`, `animichi_runner.py:326`) are `model_initiated=False` and filtered out by `TrajectoryCase.from_result` before comparison. The bridge (`_record_terminal_return`) is the sole model-initiated source and propagates the flag. No third silent `{}` source in the compared set. `OfficialArgumentCorrectness` also reads `step.params` but as a scorer, not an identity comparison — an unrecorded step scores low there, which surfaces loss rather than masks it. Fine as-is.

### 3. Diagnostic independently verified in git history

`git show 53271753^:apps/agent/agent/agents/catalog_tools.py` confirms the pre-#430 rejected-nearby path recorded `_record(ctx.deps, ToolName.SEARCH_NEARBY.value, {}, ...)` with `model_initiated=True` (old `step_recording._record` default) — so distinct rejected locations really did collapse onto `("search_nearby","{}")` in the tripwire. The claimed root cause is real, and #430 (which routed recording through the event bridge) removed that path. This PR closes the two residual `{}` fallbacks the bridge itself introduced.

### 4. Gates + mutation — independently re-run

In a fresh worktree of the PR head: `make lint` pass, `make typecheck` pass (119 files), unit suite **1394 passed** excluding the 12 tsx-dependent `test_chat_wire_contract` cases (env-only: no `pnpm install` in my worktree; consistent with the PR's 1406). Two mutants re-verified independently:
- `_comparable_calls` → return all calls: KILLED by `test_unrecorded_params_do_not_manufacture_a_repeat`
- `finish()` fallback → `params_recorded=True`: KILLED by `test_marks_params_unrecorded_when_no_call_event_was_seen`

Test assertions are strong throughout (exact failure-message equality, tuple asserts on `(params, params_recorded)`, exact metrics-line match) — no weak `is not None` patterns.

### Findings

**P2-1 — `unrecorded_params` is print-only; the tripwire can go blind silently.** If projection ever breaks wholesale (e.g. a pydantic-ai upgrade changes the args contract), every call becomes unrecorded, `repeats=0`, and the smoke gate stays green — the exact "loud loss" this PR set out to prevent would be metrics-line-loud only, and #434 taught us metrics lines don't get read. Live runs measured `unrecorded_params=0` across the board, so the gate is free today: fail (or at minimum warn-list) the enforced smoke run when `unrecorded_params > 0`. Recommend a small follow-up, not a change to this PR.

**P2-2 — `args_as_dict` TypeError escape: yes, file a follow-up issue.** `_call_params` catches only `ValidationError`. In pydantic-ai 2.11.0, `args` of a non-str/non-dict type would reach `pydantic_core.from_json` and raise `TypeError`, outside both the library's `except (ValueError, AssertionError)` and this function's guard — a terminal run error from a recording concern. Unreachable from the documented wire contract today, so keeping it out of this PR is the right call; a one-line `except (ValidationError, TypeError)` with `params_recorded=False` plus a test belongs in its own issue rather than an unreviewed widening here.

**P3 — comment accuracy nit** (`direct_gates.py` header): "the rejected-search recording path dropped its arguments, so both projected to '{}'" — worth noting inline that this path was already removed in #430 and the flag now guards the bridge's own fallbacks, so a future reader doesn't go hunting for a live `{}` recording path that no longer exists. Optional.

### Quality Ratchet
Core AC (tripwire only fires on genuinely identical recorded arguments) — covered: distinct-args, unknown-unknown, identical-recorded, genuinely-empty-repeat, metrics visibility (5 unit tests) + 4 bridge propagation tests. ac_with_test == ac_total.

eval_results: test_eval skipped locally (live-model tier; PR includes its own 50-run isolation evidence + L0 smoke pass, and per current policy eval batching is deferred to终局 runs).

lifeodyssey (commented):
## Fable review — round 2 (delta a5dfc01d..8ab0c9db), confirmed

**Final verdict: APPROVE.** The rework strengthens, and in no way widens, the properties I verified in round 1.

**① Hard-fail on unrecorded_params — strictly tighter, not wider.** `_repeat_failures`/`_comparable_calls` are untouched and the exclusion set is unchanged: the same two `params_recorded=False` sources (unprojectable args, orphan return — `finish()` is refactored pop-then-check but semantically identical), no new path sets the flag. The only behavioral delta is that an unrecorded call now fails `direct_thrash_gate` outright (`_unrecorded_failures` wired into `_case_failures`) instead of silently abstaining — this closes my round-1 P2-1 exactly. The two structlog warnings are additive observability and close the production half of it. Re-verified at HEAD: 12/12 touched tests pass; one independent mutant (unrecorded gate disabled via early return) KILLED by `test_unrecorded_params_fail_as_a_recording_defect_not_as_a_repeat`; the new negative test `test_recorded_params_alone_keep_the_recording_gate_silent` guards against over-firing.

**② Comment weakening — accurate.** The narrowed header now states the real one-way limitation (guard fingerprints VALIDATED args, records carry RAW args, so guard breakage hiding in default normalization is invisible to the tripwire) and #459 exists and describes that fork correctly. This is honest scoping, not a behavior change.

**③ Evidence — adequate.** 80/80 L0 at the CI cap on the final tree, all gates passed, `repeats=0 unrecorded_params=0` on every case; 32 targeted C1_en_001/C1_en_005 runs including 3 exercising the multi-attempt nearby/geocode retry path that used to false-positive. Consistent with the mechanism: healthy runs record every param, so the new hard gate costs nothing today and bites exactly when recording degrades.

No new findings. P2-2 (args_as_dict TypeError escape) remains a correctly-deferred follow-up; P3 comment nit is superseded by the rewrite.

PR #449, opened 2026-07-28
chore(catalog): derive spike fixtures from the shared contract (#363)

lifeodyssey:
Closes #363. Follow-up tracked in #456.

## The drift list (audited against current `src` + `packages/contract`)

| # | Fixture | Drifted from | Symptom |
|---|---|---|---|
| 1 | `resolve-api.spike.test.ts` seeds | `PointsByWorkIdInput` (`work_id: /^\d+$/`) | work ids `alpha`/`beta`/`zero` → 400 on `pointsByWorkId` |
| 2 | `geo-query.spike.test.ts` Oarai case | `MAX_RADIUS_M = 50_000` in `src/lib/geo-query.ts` | **not** an FK/seed gap as the issue suspected — the row lands fine (verified with a live direct query); the 200 km request is *clamped* to 50 km, so the ~82 km Oarai point is correctly excluded |
| 3 | `catalog-api.spike.test.ts` alias-miss (not in the issue — found by running the full live suite) | tiered miss path in `src/api/search.ts` | unstubbed → real Bangumi search → live 502 |
| 4 | `catalog-api.spike.test.ts` `afterEach` | `vi.stubGlobal` semantics | `restoreAllMocks()` does not undo `stubGlobal`; fetch stubs leaked between tests and only worked by chaining luck |

**Correction (review P1-1):** the `neonSql` stub the issue describes (`() => Promise.resolve([])`) is **not** on this PR's base. It was fixed by `8c56e225`, an unverified WIP already merged into `feat/frontend-rebuild`; base and this branch are byte-identical on that wiring. The spike's head comment is now written as a forward-looking guard ("do not stub `neonSql` — resolve's raw-SQL path needs a real one") rather than a claim about history a future reader cannot find.

## Approach — stop hand-copying snapshots

Rather than write a fresh snapshot that rots again, seeds and expectations are **derived from `packages/contract`** and the Drizzle schema:

- **`test/fixtures/catalog-seed.ts`** — every fixture value is `.parse()`d through the shared schemas at construction (`PointsByWorkIdInput`, `Latitude`, `Longitude`, `AnimeCandidate`, `ResolveOutcome`). Table and column names come from `getTableName` / `getTableColumns`, so a schema rename breaks the emitted SQL offline.
- **`test/catalog-seed.worker.test.ts`** — that guard runs in the always-on worker pool, so fixture rot is visible in CI *without* Neon credentials (the spikes skip when they are absent, which is how this rotted unseen).
[...]

PR #454, opened 2026-07-28
fix(ci): grant pull-requests:read to the changes job (unblocks all path-gated CI)

lifeodyssey:
Every PR's `changes` job (dorny/paths-filter) fails with `Resource not accessible by integration`, which makes **all path-gated CI lanes skip** — merges have been running blind (observed on #448, #449, and unrelated branches; systemic, not diff-related).

Root cause: the workflow-level `permissions:` block grants only `contents: read`. `dorny/paths-filter` on `pull_request` events lists the PR's changed files via the REST API, which needs `pull-requests: read`. The `changes` job has no job-level permissions block, so it inherits the too-narrow default.

Fix: job-level `permissions` for `changes` only (contents + pull-requests read). No other job touched.

Verification: this PR's own CI run is the test — the `changes` job must go green and the path-gated lanes must actually execute.

### DIFF
--- a/.github/workflows/_security.yml
+++ b/.github/workflows/_security.yml
@@ -20,6 +20,9 @@ jobs:
   gitleaks:
     name: gitleaks (secret scan)
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read  # gitleaks-action lists PR commits via the API; without this it dies with "Resource not accessible by integration"
     steps:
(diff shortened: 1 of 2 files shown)


lifeodyssey:
Superseded: the changes-job fix rides on #448 (identical hunk, already live-validated there — run 30370192030 shows the lanes actually executing). The gitleaks half of this PR triggered the known reusable-workflow startup_failure (job permissions exceeding the caller's grant) and needs the coordinated two-file fix tracked in #455 instead.

Issue #284, opened 2026-07-07
S1.11 — Bring-your-own-key (BYOK) support with secure credential handling


lifeodyssey:
## Releasable statement
Power users can configure their own LLM provider credentials (choice of OpenAI-compatible, Anthropic, or Gemini) for chat, stored client-side only. Subsequent requests use that credential for the main model call instead of the platform default. Credentials are scrubbed from all logging/observability surfaces, and outbound requests are validated to prevent server-side request forgery against internal or otherwise unsafe network destinations.

## AC summary
- AC count: 13
- Test type distribution: unit=4, integration=8, browser=1, eval=0, api=0, manual/ops=0

## Dependencies
S1.1.

## Spec source
`docs/superpowers/specs/2026-07-06-frontend-rebuild/iter-1.md` (commit 6fc13b8)

### S1.11 BYOK (bring your own LLM key, three provider families, finalized under SD-11) + the SSRF egress guard (backfilled from SD-20, now finalized, replacing the original `pending-confirmation` P8 label)

**User story**: As a power user, I want to be able to use the product with my own LLM key (a choice of OpenAI-compatible / Anthropic / Gemini), unconstrained by the free quota, and be confident that this key never leaves my browser except as a passed-through request header, never appears in any log, and can never be abused to hit an internal address I don't want it to hit — including one behind a relay service I've deployed myself.

**Design basis**: no visual canvas; main spec inputs §10 SD-11 + Step5 (SD-20, finalized 2026-07-06, backed by industry research on BYOK + closing out P8) + X3 (finalized).

**Releasable statement**: the chat input area (group G) provides a settings entry point — choose a provider (OpenAI-compatible / Anthropic / Gemini, pick one) + enter a key (+ an optional `base_url` for the OpenAI-compatible family), stored client-side only (`sessionStorage`/in-memory state + a strict CSP, **with no homegrown encryption** — backfilled from SD-20, since frontend encryption is security theater and isn't introduced); subsequent chat requests pass the corresponding provider's credential through as a request header (a request-scoped local variable, released as soon as the function returns, never persisted or stored server-side), and the agent uses it instead of the server-side default key to make the main-loop LLM call (internal helper calls still use the server-side key, D18); the key never appears in any log/trace; outbound requests (especially to a custom `base_url`) go through the strict SSRF egress guard.

**AC**:
[...]

PR #458, opened 2026-07-28
feat(agent): SSRF egress guard for BYOK egress (#284 Task 1)

lifeodyssey:
## Summary

Implements **Task 1** of the #284 BYOK design spec
(`docs/superpowers/specs/2026-07-28-284-byok-design.md`): the SSRF egress guard
that every other BYOK task depends on. Standalone — no BYOK wiring in this PR.

- `agent/infrastructure/egress_guard.py` (new): `validate_base_url()` (pre-flight,
  resolve-once, dual-condition IP acceptance) + `GuardedAsyncTransport` (connect-time
  re-validation, IP-literal pinning with preserved `Host` header and TLS
  `sni_hostname`, 3xx refusal) + `build_guarded_async_client()` (`trust_env=False`,
  no proxy mounts — T13).
- Interpreter floor raised to `>=3.11.10` in `apps/agent/pyproject.toml` (+ `uv.lock`
  regen); `Dockerfile` pinned to `python:3.11.13-slim` instead of the floating
  `3.11-slim` tag. CPython gh-113171 fixed `ipaddress.is_global` for CGNAT ranges in
  3.11.10/3.12.4 — pre-fix interpreters would silently lose the P1-1 protection.

## AC coverage (Task 1, 9 ACs)

| AC | Test | Type |
|---|---|---|
| Happy path (pinned connect, correct `Host` + TLS `sni_hostname`) | `test_egress_ssrf_guard.py::test_happy_path_reaches_stub_with_original_host_and_sni` (real self-signed TLS stub server) | integration |
| Null/empty/boundary shape (None/""/"   "/no-scheme/userinfo/port-22/port-8443/punycode/IPv6-bracketed) | `test_egress_guard_classification.py` (parametrized, each asserted individually) | unit |
| IP-literal `base_url` rejected (127.0.0.1, 10.0.0.5, 169.254.169.254) | `test_ip_literal_base_url_rejected_without_opening_a_socket` | integration |
| P1-1 dual-condition regression (100.100.100.200, 100.64.0.1 CGNAT, NAT64 `64:ff9b::7f00:1`) + direct `is_global` assertion (P2-3/rev5) | `test_dual_condition_regression_addresses_are_denied` + `test_cgnat_is_global_false_asserted_directly` | integration |
| TOCTOU (resolve-once, pin, never re-query) | `test_toctou_connects_only_to_first_resolved_address` | integration |
| Redirect refusal (3xx never followed) | `test_redirect_response_is_refused_not_followed` | integration |
| IPv6 loopback/mapped/mixed record sets | `test_ipv6_special_and_mixed_record_sets_are_rejected` | integration |
| Scheme enforcement (http/file/gopher/ftp rejected pre-resolution) | `test_rejects_disallowed_schemes_before_resolution` | unit |
| T13: `trust_env=False`, no proxy mounts, even with `HTTPS_PROXY`/`ALL_PROXY` set | `test_guarded_client_has_no_proxy_mounts_even_with_proxy_env_set` + `test_guarded_transport_ignores_proxy_env_and_pins_to_resolved_ip` | unit |

`ac_total == ac_with_test` = 9/9.

## Mutation gate evidence (Verification Plan §3 — both mandatory, verified by hand)

1. **Redirect/TOCTOU gate**: commented out `GuardedAsyncTransport`'s connect-time
   re-validation call (the `validate_base_url` + URL rewrite in
   `handle_async_request`) → `test_toctou_connects_only_to_first_resolved_address`
   goes **red** (`assert 'rebind.example' == '8.8.8.8'` — proves the transport was
   actually pinning, not the pre-flight check alone).
2. **Dual-condition gate — `is_global` half**: deleted the `if ip.is_global is not
   True: return False` branch → `test_dual_condition_regression_addresses_are_denied[100.64.0.1]`
   goes **red** (`DID NOT RAISE EgressBlocked`).
3. **Dual-condition gate — deny-flag half**: deleted the
   `is_private/is_loopback/.../is_unspecified` branch → 
   `test_dual_condition_regression_addresses_are_denied[64:ff9b::7f00:1]` goes **red**.

All three restored and full suite re-verified green after each gate.

## Gate results

- `make lint` — pass (ruff check + format)
- `make typecheck` — pass, 120 source files, 0 errors
- `make test` — 1417 passed, coverage 88.28% (floor 87%)
- New files: 35/35 tests passing (`test_egress_guard_classification.py` unit + 
  `test_egress_ssrf_guard.py` integration, using a real self-signed TLS stub server)
- `make test-integration` (full suite): blocked in this sandbox by a pre-existing,
  unrelated environment mismatch (`installed Atlas is 1.2.4; expected 0.30.0` for the
  DB-backed testcontainer fixtures) — confirmed unrelated to this change since the
  new egress-guard integration tests carry no DB dependency and pass standalone.

## Test plan

- [x] All 9 Task 1 ACs have a corresponding test (unit/integration as annotated above)
- [x] Both mutation gates from the Verification Plan run and confirmed red/green
- [x] `make lint && make typecheck && make test` green
- [x] New test files ≤200 lines (TLS stub server extracted to
      `_egress_tls_stub.py` helper, not itself a test module)
- [x] No suppressions (`noqa`/`type: ignore`/etc.) added
- [ ] Not merging — per harness workflow, awaiting dual review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,8 @@
-FROM public.ecr.aws/docker/library/python:3.11-slim AS builder
+# Pinned to a patch tag, not the floating `3.11-slim` — #284 Task 1 raises the
+# interpreter floor to >=3.11.10 (CPython gh-113171 fixed `ipaddress.is_global`
+# classification for CGNAT/metadata ranges in 3.11.10/3.12.4). A floating minor
+# tag could silently regress below that fix.
+FROM public.ecr.aws/docker/library/python:3.11.13-slim AS builder
 
 COPY --from=ghcr.io/astral-sh/uv:0.10.9 /uv /uvx /bin/
 
@@ -17,7 +21,7 @@ COPY apps/agent/agent /app/agent
 
 RUN uv sync --no-dev
 
-FROM public.ecr.aws/docker/library/python:3.11-slim
+FROM public.ecr.aws/docker/library/python:3.11.13-slim
 
 ENV PYTHONDONTWRITEBYTECODE=1 \
     PYTHONUNBUFFERED=1 \
--- a/apps/agent/agent/infrastructure/egress_errors.py
+++ b/apps/agent/agent/infrastructure/egress_errors.py
@@ -1,0 +1,67 @@
+"""Typed error taxonomy for the SSRF egress guard (#284 Task 1).
+
+Split out from `egress_guard.py` to keep that module under the 300-line file
+cap; imported by both `egress_guard.py` (raises) and `egress_transport.py`
+(raises on redirect refusal).
+"""
+
+from __future__ import annotations
+
+import enum
+from typing import Final
+
+
+class EgressBlockReason(enum.Enum):
+    """Machine-readable rejection category.
+
+    Task 3 maps these to a public error taxonomy — match on `.reason`, never
+    on `str(exc)`, which carries only a fixed, non-identifying message.
+    """
+
+    EMPTY_URL = "empty_url"
+    INVALID_URL = "invalid_url"
+    INVALID_SCHEME = "invalid_scheme"
+    INVALID_USERINFO = "invalid_userinfo"
+    INVALID_HOST = "invalid_host"
+    INVALID_HOSTNAME_ENCODING = "invalid_hostname_encoding"
+    INVALID_PORT = "invalid_port"
+    OWN_INFRASTRUCTURE = "own_infrastructure"
+    RESOLUTION_TIMEOUT = "resolution_timeout"
+    RESOLUTION_FAILED = "resolution_failed"
+    NO_ADDRESSES = "no_addresses"
+    ADDRESS_NOT_ROUTABLE = "address_not_routable"
+    REDIRECT_REFUSED = "redirect_refused"
+
+
+_REASON_MESSAGES: Final[dict[EgressBlockReason, str]] = {
+    EgressBlockReason.EMPTY_URL: "egress URL is empty",
+    EgressBlockReason.INVALID_URL: "egress URL could not be parsed",
+    EgressBlockReason.INVALID_SCHEME: "egress URL scheme is not allowed",
+    EgressBlockReason.INVALID_USERINFO: "userinfo is not allowed in the egress URL",
+    EgressBlockReason.INVALID_HOST: "egress URL is missing a host",
+    EgressBlockReason.INVALID_HOSTNAME_ENCODING: "egress URL host is not a valid hostname",
+    EgressBlockReason.INVALID_PORT: "egress URL port is not on the allowlist",
+    EgressBlockReason.OWN_INFRASTRUCTURE: "egress host is own infrastructure",
+    EgressBlockReason.RESOLUTION_TIMEOUT: "DNS resolution timed out",
+    EgressBlockReason.RESOLUTION_FAILED: "DNS resolution failed",
+    EgressBlockReason.NO_ADDRESSES: "no addresses resolved for host",
+    EgressBlockReason.ADDRESS_NOT_ROUTABLE: "resolved address is not publicly routable",
+    EgressBlockReason.REDIRECT_REFUSED: "redirect response refused",
+}
+
+
+class EgressBlocked(Exception):
+    """Raised whenever a candidate egress destination fails SSRF validation.
+
+    `str(exc)` is always one of the fixed `_REASON_MESSAGES` — it never
+    embeds the submitted URL, hostname, resolved IP, or the underlying
+    `OSError` text, none of which should reach a caller or an unguarded log
+    line (an internal-network DNS oracle is still an oracle). Anything
+    diagnostic is on `.detail`, for structured logging only, subject to the
+    same redaction path as everything else (Task 2).
+    """
+
+    def __init__(self, reason: EgressBlockReason, *, detail: str | None = None) -> None:
+        self.reason = reason
+        self.detail = detail
+        super().__init__(_REASON_MESSAGES[reason])
--- a/apps/agent/agent/infrastructure/egress_guard.py
+++ b/apps/agent/agent/infrastructure/egress_guard.py
@@ -1,31 +1,30 @@
-"""SSRF egress guard for user-influenceable outbound HTTP calls (#284 Task 1).
+"""SSRF pre-flight validation for user-influenceable outbound HTTP calls (#284 T1).
 
-Two layers, both required (see `docs/superpowers/specs/2026-07-28-284-byok-design.md`):
+`validate_base_url` is the pre-flight, request-boundary check: resolves the
+host exactly once for that call and accepts an address only under a **dual
+condition**: `ip.is_global is True` AND it trips none of the private/loopback/
+link-local/reserved/multicast/unspecified flags. Neither half is redundant —
+see the docstring on `_is_address_accepted` for the two address families each
+half alone would miss. No domain allowlist.
 
-1. `validate_base_url` — a pre-flight, resolve-once check at the request boundary.
-2. `GuardedAsyncTransport` — an `httpx` transport that re-validates at connect time,
-   pins the socket to the already-validated IP literal, preserves TLS hostname
-   verification via `sni_hostname`, and refuses every 3xx response.
-
-No domain allowlist. Acceptance is a **dual condition** on the resolved IP:
-`ip.is_global is True` AND it trips none of the private/loopback/link-local/
-reserved/multicast/unspecified flags. Neither half is redundant — see the
-docstring on `_is_address_accepted` for the two address families each half alone
-would miss.
+`GuardedAsyncTransport` (in `egress_transport.py`) is the second, independent
+layer: it re-runs this same check at connect time and pins the socket to the
+address *that* call validated — see its docstring for why that closes the
+TOCTOU/DNS-rebinding window.
 """
 
 from __future__ import annotations
 
 import ipaddress
 import socket
-import ssl
 from collections.abc import Awaitable, Callable
 from dataclasses import dataclass
 from typing import Final
-from urllib.parse import urlsplit
+from urllib.parse import SplitResult, urlsplit
 
 import anyio
-import httpx
+
+from agent.infrastructure.egress_errors import EgressBlocked, EgressBlockReason
 
 IpAddress = ipaddress.IPv4Address | ipaddress.IPv6Address
 
@@ -34,6 +33,12 @@ Resolver = Callable[[str, int], Awaitable[list[str]]]
 ALLOWED_SCHEME: Final[str] = "https"
 ALLOWED_PORTS: Final[frozenset[int]] = frozenset({443, 8000, 8443})
 RESOLUTION_TIMEOUT_SECONDS: Final[float] = 5.0
+
+# DNS resolution runs on its own capacity limiter, not anyio's shared default
+# thread pool (40 tokens, shared with every other `to_thread.run_sync` call in
+# the process). A `base_url` pointing at a black-hole nameserver must not be
+# able to exhaust threads other subsystems depend on.
+_DNS_THREAD_LIMITER: Final[anyio.CapacityLimiter] = anyio.CapacityLimiter(4)
 
 # Belt-and-suspenders explicit deny set for well-known cloud-metadata endpoints.
 # These IPs already fail the `is_global`/deny-flag classification below on their
@@ -50,12 +55,20 @@ _METADATA_DENY_IPS: Final[frozenset[str]] = frozenset(
 
 # Our own public-facing origins are legitimately globally routable, so every IP
 # check above passes them. Without an explicit deny, BYOK egress becomes a
-# confused-deputy path back into our own authenticated surfaces.
-OWN_INFRASTRUCTURE_HOSTNAMES: Final[frozenset[str]] = frozenset({"animichi.com"})
-
-
-class EgressBlocked(Exception):
-    """Raised whenever a candidate egress destination fails SSRF validation."""
+# confused-deputy path back into our own authenticated surfaces, where a
+# request may carry more implicit trust than it would from the open internet.
+#
+# This is a hostname-suffix match, not an IP-based one — it does not follow
+# CNAMEs and will not catch a hostname that merely *resolves to* our
+# infrastructure's IPs under a different name. Whenever a new public-facing
+# origin is added (a new custom domain, a new auth provider host, a new
+# internal service hostname), add it here in the same commit.
+OWN_INFRASTRUCTURE_HOSTNAMES: Final[frozenset[str]] = frozenset(
+    {
+        "animichi.com",  # apex + `*.animichi.com` (wrangler.toml route)
+        "stack-auth.com",  # Neon Auth (Better Auth) — per-project subdomain
+    }
+)
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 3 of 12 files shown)


lifeodyssey (commented):
## Fable seat — independent review of #458 (#284 Task 1, SSRF egress guard)

Method: detached review worktree at `e32e65ce`, attacker-input matrix run against the real `validate_base_url`, independent gates (`make lint` / `make typecheck` / targeted pytest), 3 mutation spot-checks, `uv.lock` diff audit.

**Verdict: request_changes** — the core (dual condition, resolve-once + pinned connect, SNI preservation, 3xx refusal, `trust_env=False`) is faithful and mutation-proven. The *own-infrastructure* control is bypassable with one character and has zero tests.

### P1

1. **Own-infrastructure deny is bypassable with a trailing dot / IDNA dot variants** — `egress_guard.py:66,70-74,153`.
   `_is_own_infrastructure` compares the IDNA A-label by string equality, and Python's `idna` codec preserves a trailing root dot and normalises U+3002/U+FF61 into one. Measured against the shipped code:
   ```
   'https://animichi.com'   -> BLOCKED  (own infrastructure)
   'https://animichi.com.'  -> ACCEPTED host='animichi.com.'
   'https://animichi.com。'  -> ACCEPTED host='animichi.com.'   (U+3002)
   'https://animichi.com｡'  -> ACCEPTED host='animichi.com.'   (U+FF61)
   'https://api.animichi.com.' -> ACCEPTED
   ```
   `getaddrinfo("example.com.")` resolves identically to `example.com` on both macOS and glibc (verified), so the confused-deputy path the spec calls out at line 263 is open. Fix: normalise `hostname.rstrip(".")` before the deny comparison (and before `_METADATA_DENY_IPS`-style string checks generally), with a test per variant.

2. **Deny set is incomplete versus spec, and completely untested.** Spec §(a)4(ii) requires denying "`animichi.com` and subdomains, **the Neon Auth origin, and the users/catalog service hostnames**". Only `animichi.com` is implemented. Coverage confirms line 154 (`is own infrastructure`) is **never executed by any test** — no test in the diff references `animichi.com`. Risk-table row "Own-infrastructure deny set … + AC" has no AC and no test; that is why (1) shipped undetected.

3. **Patch coverage 90% (< 95% bar).** `agent/infrastructure/egress_guard.py` 97 stmts / 9 missed. Uncovered lines are all security branches:
   - `116-119` DNS timeout / `OSError` → `EgressBlocked` (spec: "a resolution timeout is `EgressBlocked`, not a hang" — untested)
   - `131` missing host, `134-135` invalid-hostname/IDNA failure
   - `154` own-infrastructure deny (see above)
   - `158` empty resolution result

4. **CI never validated this PR.** `changes` (paths-filter) failed with `Resource not accessible by integration`, so **Agent CI / Edge worker tests / Codecov were all `skipping`**. There is no CI evidence of lint, typecheck, tests, or patch coverage on this branch. Must be green before merge. (`Security / Semgrep` and `osv-scanner` failures are pre-existing and unrelated: one `dangerouslySetInnerHTML` and npm `next`/`sharp`/`tar`/`valibot` advisories from `pnpm-lock.yaml`; the regenerated `uv.lock` introduced no Python advisory.)

### P2

5. **Uncaught `ValueError` breaks the typed-`EgressBlocked` contract** — `_parse_endpoint_shape:124`. `urlsplit` raises `ValueError` (not `UnicodeError`) on NFKC-invalid netlocs:
   `https://℀animichi.com` → `!!! UNCAUGHT ValueError: netloc '℀animichi.com' contains invalid characters under NFKC normalization`.
   Second site: `ipaddress.ip_address(candidate)` at line 162 raises `ValueError` for any non-IP resolver output. Fail-closed in effect, but every downstream task (3/5) will catch `EgressBlocked` and map to 4xx — these become 500s. Wrap both in `EgressBlocked`.

6. **Error strings disclose resolution results** — `"resolved address 10.1.2.3 is not publicly routable"`, `"DNS resolution failed for 'x': …"`. If Task 3/5 surfaces `EgressBlocked` text to the caller, this is an SSRF/internal-DNS oracle. Recommend a generic user-facing message with the detail logged only.

7. **Connection-pool coalescing keyed on the pinned IP.** After the rewrite, httpcore's pool origin becomes `(https, <IP>, port)`, so two different hostnames resolving to the same IP share one TLS connection established (and certificate-verified) for the *first* hostname. Safe only because the docstring says "per-request BYOK transport" — that lifecycle is currently a comment, not a constraint. Please pin it with an AC/assertion in the task that constructs the client (no caching/reuse of the guarded client across requests).

8. **Suppressions added without approval** — `_egress_tls_stub.py:44-45` `# type: ignore[operator]`, caused by typing the parameter `directory: object`. CLAUDE.md forbids new suppressions; `directory: Path` removes both.

9. **`test_egress_ssrf_guard.py` is 201 lines**, one over the project's 200-line test-file cap — the cap that `_egress_tls_stub.py` was extracted to satisfy. (The stub is a legitimate fixture helper, no `test_` functions; that split is fine.)

10. **1-10-50**: `validate_base_url` 22 lines, `_parse_endpoint_shape` 16, `handle_async_request` 15, `_is_address_accepted` 12, `_default_resolve` 11. Clean extract candidate: `_classify_addresses(addresses) -> list[IpAddress]` out of `validate_base_url`.

11. **Two ACs are weaker than their names claim.**
    - AC1 happy path monkeypatches `_is_address_accepted` to `True`, so the acceptance rule is not exercised on the only path that opens a real socket (documented in the test, acceptable, but the name over-claims).
    - AC3 `test_ip_literal_base_url_rejected_without_opening_a_socket` never asserts that no socket was opened (it does perform DNS). Add a `socket.socket` spy or assert the resolver/connect was not reached.
    - AC2 requires punycode `Host` **and** `sni_hostname` to carry the A-label; only `endpoint.hostname` is asserted — no transport-level assertion for the IDN case.

### P3

12. One observed, **non-reproducible** failure of `test_guarded_transport_ignores_proxy_env_and_pins_to_resolved_ip` (`assert 'host.example' == '8.8.8.8'`) on the first run in a pre-existing worktree; ~30 subsequent runs green (fixed order, randomised order, cold venv, with/without real proxy env). Symptom is byte-identical to my mutant-1 (pinning removed), so most likely stale worktree/pyc state rather than a product flake — flagging so CI reruns are watched.
13. The interpreter assertion `ipaddress.ip_address("100.64.0.1").is_global is False` is duplicated in both the unit and the integration file; one is enough.
14. Tests assert httpx privates (`client._mounts`, `_trust_env`, `_transport`) — fragile across httpx upgrades; `client.get`-level behaviour assertions would survive.
15. `metadata.google.internal` is only covered via its resolved `169.254.169.254`; acceptable, worth a comment.

### Verified good

- **Mutation gates (3/3 killed)**: (a) remove the transport's IP pin → the pinning ACs go red; (b) delete the `is_global` half → AC4 red; (c) disable the 3xx branch → AC6 red.
- Numeric-encoding attacks are handled correctly by classifying **post-resolution**: `2130706433`, `0x7f000001`, `017700000001`, `127.1`, `[::ffff:169.254.169.254]`, `[fe80::1%eth0]`, `localhost` all blocked with the real resolver.
- Userinfo (`safe.com@evil.com`, `host:443@evil.com`), non-https schemes (resolver proven not called), non-allowlisted ports, empty labels, over-long labels: all blocked.
- Dual condition, mixed A/AAAA "any-hit rejects", 3xx refusal + `follow_redirects=False`, `trust_env=False` with no mounts: faithful to spec.
- TLS: the real self-signed stub proves `sni_hostname` reaches the handshake as the original hostname while the socket goes to the pinned literal — the highest-value assertion in this PR.
- Interpreter triple insurance landed: `requires-python >=3.11.10`, both Dockerfile stages pinned to `3.11.13-slim`, direct `is_global` assertion. Note CI uses `PYTHON_VERSION: "3.11"` via `uv python install`, so the floor is enforced only by `requires-python` + the assertion — which is exactly the designed defence.
- `uv.lock` regen is clean: resolution-marker churn + `tomli` drop (a consequence of the floor). No unrelated version bumps.
- `make lint` and `make typecheck` pass locally on this branch.

Quality ratchet: 9 Task-1 ACs, ~8 with a real test (AC2 partial); two spec-body requirements — own-infrastructure deny and DNS-timeout→`EgressBlocked` — have no test at all.

lifeodyssey (commented):
## Fable seat — re-review of #458 at `bfa0cb45`

Re-ran my original probes against the new tree, plus a second-round normalization probe, all gates, and **five** mutants on the refactored modules.

**Verdict: approve.** Every P1 and every P2 from my first pass is closed, and three of the fixes are stronger than what I asked for (typed reason enum instead of just wrapping the `ValueError`; `abandon_on_cancel=True` + a dedicated 4-slot DNS limiter; `max_keepalive_connections=0` with a real two-handshake TLS regression test).

### ① Trailing-dot / full-width-dot bypass — CLOSED (re-attacked with the original probe)
```
'https://animichi.com'      BLOCKED[own_infrastructure]
'https://animichi.com.'     BLOCKED[own_infrastructure]
'https://animichi.com。'     BLOCKED[own_infrastructure]   (U+3002)
'https://animichi.com｡'     BLOCKED[own_infrastructure]   (U+FF61)
'https://animichi.com．'     BLOCKED[own_infrastructure]   (U+FF0E, not in the test set — also covered)
'https://ANIMICHI.com.'     BLOCKED[own_infrastructure]
'https://api.animichi.com.' BLOCKED[own_infrastructure]
'https://animichi.com.:8443' BLOCKED[own_infrastructure]
'https://a.b.animichi.com。'  BLOCKED[own_infrastructure]
'https://myproj.stack-auth.com.' BLOCKED[own_infrastructure]
```
Also blocked through the transport (`https://animichi.com./v1`, `https://animichi.com。/v1`), so the second layer inherits the normalization. Mutation M4 (drop `.rstrip(".")`) turns `test_own_infrastructure_hostname_is_rejected[animichi.com.]` red — the fix is pinned, not incidental.

### ② Deny-set completeness vs spec §(a)4(ii) — accepted, with one residual for a later PR
- `animichi.com` (+ subdomains) ✓ — matches `wrangler.toml` route `animichi.com/*`.
- Neon Auth ✓ — `stack-auth.com` covers the per-project subdomain (`project.stack-auth.com` is in the test matrix).
- **users worker**: no gap. `workers/users/wrangler.toml` sets `workers_dev = false` at top level **and** in staging **and** production, with the comment "No public workers.dev host — reached only via the root Worker USERS service binding." There is no public hostname to deny.
- **catalog worker (P3, not blocking)**: `workers/catalog/wrangler.toml` sets `workers_dev = false` only under `[env.preview]`; the default, `[env.staging]` and `[env.production]` blocks don't, so `catalog*.<account>.workers.dev` may exist as a public origin outside the deny set. I checked whether that matters: `workers/catalog/src/index.ts` and `router.ts` carry no auth/shared-secret check, so that surface is equally reachable from the open internet — reaching it from inside the container confers **no additional trust**, which is the entire premise of the own-infrastructure deny. Recommendation is therefore `workers_dev = false` on the catalog envs (defence in depth, catalog's own PR), not a change here.
- `*.neon.tech` / `console.neon.tech`: considered and argued not required — the guarded client never attaches our credentials, and the Neon API is bearer-authenticated, so there is no source-position trust to abuse. Noting it so the question isn't re-litigated later.
- The "add new origins in the same commit" comment on `OWN_INFRASTRUCTURE_HOSTNAMES:60-65` is the right durable answer to a list that will drift.

### ③ Bare `ValueError` → typed — CLOSED, and better than requested
`https://℀animichi.com` → `BLOCKED[invalid_url]`; `_parse_ip_address` wraps the `ipaddress` `ValueError` into `ADDRESS_NOT_ROUTABLE`. `EgressBlockReason` + fixed `_REASON_MESSAGES` also closes my P2-6 (SSRF oracle): `test_error_message_never_embeds_the_offending_address` proves `10.1.2.3` is on `.detail` only, never in `str(exc)`. Task 3 now has a machine-readable contract to map instead of string matching.

### ④ P2 sweep — all closed
| item | status |
|---|---|
| 201-line test file | split 4 ways: 189 / 198 / 107 / 133, stub 130 — all under the cap |
| 1-10-50 | only `_default_resolve` (15, try/except — the rule's own exception) and `_resolve_and_classify` (12, one loop) exceed by my AST count; classes 19/13/4 lines; modules 253/123/67 |
| `# type: ignore[operator]` in the stub | gone (zero suppressions across all seven files) |
| AC1 monkeypatch | still forces `_is_address_accepted`, now with the reason documented in-test — correct call, the classification is proven in `..._addresses.py` |
| A-label through the transport | `test_transport_rewrites_punycode_host_to_a_label` ✓ |
| "no socket opened" was never asserted | `test_ip_literal_url_never_opens_a_socket` monkeypatches `socket.socket` to raise, drives the **real** resolver through the full guarded client ✓ |
| connection-pool coalescing (P2-7) | fixed in code (`max_keepalive_connections=0`) **and** pinned by `test_distinct_hostnames_sharing_pinned_ip_get_isolated_connections`, which asserts `connection_count == 2` plus per-connection SNI and Host history against a real TLS stub |
| undeclared `cryptography` test import | now a declared dev dep (`>=46.0.0`), the only substantive `uv.lock` delta since `e32e65ce` |

### Mutation — 5/5 killed (re-verified on the refactored modules)
| mutant | killed by |
|---|---|
| M1 drop the IP pin in `_rewrite_for_pinned_endpoint` | `test_guarded_transport_ignores_proxy_env_and_pins_to_resolved_ip` |
| M2 delete the `is_global` half | `test_dual_condition_regression_addresses_are_denied[100.64.0.1]` |
| M3 disable the 3xx refusal | `test_redirect_response_is_refused_not_followed` |
| M4 delete `.rstrip(".")` | `test_own_infrastructure_hostname_is_rejected[animichi.com.]` |
| M5 `max_keepalive_connections` 0 → 10 | `test_distinct_hostnames_sharing_pinned_ip_get_isolated_connections` |

### Gates (independently run at `bfa0cb45`)
- `make lint` ✓ · `make typecheck` ✓ (122 files; the numpy-stub error I hit first is a `uv sync --all-groups` artifact — reproduces identically on the base branch, not PR-caused)
- 53 guard tests pass; **patch coverage 99%** on the three new modules (errors 100%, transport 100%, guard 98% — the only miss is the `OSError` branch at `egress_guard.py:149-150`, which is fine to leave uncovered)
- **CI is now green on the paths that matter**: `changes` pass, `Agent CI / Python CI` pass, `Python integration (Neon)` pass, SonarCloud pass — my P1-4 is resolved. The three still-red jobs are repo-wide and pre-existing, unchanged by this PR: Semgrep = one `dangerouslySetInnerHTML` in the frontend, osv-scanner = 17 npm advisories in `pnpm-lock.yaml`, gitleaks = the action's own `Resource not accessible by integration` permission error.

### Remaining nits (P3, non-blocking, fine to carry)
1. `test_default_resolve_dns_timeout_is_bounded` asserts on wall-clock elapsed (`< 1.0s` against a 2.0s blocking sleep and a 0.15s timeout). CLAUDE.md says mock the clock; the 6.7× margin makes this low-risk and the property is genuinely wall-clock-shaped, so I'd keep it — just don't tighten the bound.
2. `egress_transport.py` imports the private `_default_resolve` across module boundaries; consider exporting it under a public name.
3. `https://animichi.com%2e` is accepted at shape level (hostname stays literally `animichi.com%2e`) and only dies later at DNS — verified `getaddrinfo("example.com%2e")` fails, so it is not a live bypass. Rejecting `%` in a hostname would produce a clearer, earlier typed error.
4. `metadata.google.internal` is still covered only via its resolved `169.254.169.254` — correct, worth a one-line comment.

Quality ratchet: 9/9 Task-1 ACs have tests, plus tests for both spec-body requirements that had none (own-infrastructure deny, DNS-timeout → `EgressBlocked`). Ship it.

PR #461, opened 2026-07-28
feat(web): #273 T1 — E2 selection tray + selected_point_ids recompute bypass (living document)

lifeodyssey:
Implements **#273 S1.7 Task 1** — E2 checkbox selection tray and the `selected_point_ids` recompute bypass, per spec `docs/superpowers/specs/2026-07-28-s1.7-living-document-save-login-wall-design.md` (final, `fa8d4448` on `docs/spec-273-s1.7`).

## What changed

**New**
- `apps/web/src/features/chat/selection/useSpotSelection.tsx` — shared spot-selection Set (context; inert default outside a selection scope)
- `apps/web/src/features/chat/selection/useRecomputeTurn.ts` — recompute turn lifecycle (idle/busy/failed) + empty-fire guard
- `apps/web/src/features/chat/components/SelectionTray.tsx` — sticky E2 recompute bar (design `.rebar`)
- `apps/web/src/lib/chat/selectedPointsBypass.ts` — bypass body builder, `sameIds`, recompute-part detection

**Changed**
- `SearchResult.tsx` — the C3a checkbox becomes controlled
- `use-chat-session.ts` — `sendSelectedPoints`: `sendMessage(undefined, { body })` via ai@6.0.225's per-call body merge (verified against installed source: `{...resolvedBody, ...options.body}`)
- `MessageList.tsx` — recompute turns render the 「✓ 再計算 1.2s」 footprint instead of tool badges; part-less marker messages never render
- `ChatPage.tsx` — provider + tray mount; a failed recompute masks the full-page `TurnFailure`
- `search-i18n.ts` — tray/footprint copy ja/zh/en
- `styles/chat.css` — tray styles, Animal Island semantic tokens only

## The one non-obvious mechanism (please review)

`sendMessage(undefined, {body})` in ai@6.0.225 **continues the previous assistant message** (`createStreamingUIMessageState` reuses an assistant `lastMessage`), so the streamed same-ID `response` part would *overwrite* the prior route card — destroying the E1 append semantics. The fix appends a **part-less user marker message** before sending, forcing a fresh assistant message. Consequences:
- The marker is filtered from rendering (`MessageList.visibleMessages`).
- On the wire the marker joins `messages`; `routes/chat.py::_last_user_text` derives `text=""` (verified) — i.e. "no new utterance".

## AC coverage (9 ACs / 9 annotations)

| AC | Type | Test |
|---|---|---|
| Happy path: tray + skeleton→card, only skeleton/footprint states, no tool badges (structural) | browser | `e2e/web-chat-selection.spec.ts` (both specs **run green** against the local dev server) |
| Exactly one `POST /v1/chat` with `selected_point_ids`, no free-text turn | integration | `tests/unit/chat/chat-page-selection.test.tsx` (MSW spy on the real transport; body + user-turn-count asserted). Backend bypass dispatch is pre-existing behavior covered by the agent suite (`test_selected_route.py` etc.) |
| 「再計算 1.2s」 from an injected, mocked clock — exact string, no sleeps | unit | `tests/unit/chat/recompute-footprint.test.tsx` (fake timers advance 1200ms → `✓ 再計算 1.2s`) |
| Appended card + prior `.chat-card--superseded` + 「以前の版」 via the **existing** `supersededFlags`; static no-second-implementation check | unit | `recompute-footprint.test.tsx` + `selected-points-bypass.test.ts` (fs scan: exactly one definition; one class producer) |
| Null/empty: no tray at zero; untick-last removes; can never fire empty | unit | `selection-tray.test.tsx`, `use-recompute-turn.test.tsx` |
| Error path: inline tray retry, selection + prior card preserved, no `TurnFailure` | browser | `e2e/web-chat-selection.spec.ts` (+ jsdom reinforcement in `chat-page-selection.test.tsx`) |
| i18n ja/zh/en incl. count interpolation | unit | `selection-tray.test.tsx` |
| Multi-turn: three versions ascending, exactly newest un-superseded (render order) | unit | `recompute-footprint.test.tsx` |
| History integrity (P2-⑤): no duplicated user rows on recompute | integration | Client half here: the recompute adds **no genuine user turn** (asserted on the request body). The server half — `persistence.py` skipping the user-message insert when the turn carries no new utterance — is **Task 3 scope** and lands on `feat/273-t3-migration-retention` (parallel); this card's marker (`text=""`) is the interface signal it keys on. The end-to-end row-count integration test rides that branch. |

## Mutation verification (all killed)

1. drop empty-selection guard → `use-recompute-turn` fails
2. add a second `supersededFlags` implementation under `src/` → static check fails
3. drop the `TurnFailure` mask → "skips TurnFailure" fails
4. `sameIds` → always false → tray-visibility tests fail
5. drop the `selected_point_ids` body → body assertion fails
6. drop the marker append (same-ID overwrite returns) → card-append assertions fail
7. hardcode the footprint elapsed → exact-string clock test fails

## Design-sync alignment (`docs/design/2026-07-06-design-sync/Chat 完整状态.html`)

Followed: gold-bordered `.rebar` bar docked above the composer; sub line 「スポットの選択が変わりました」 + 「N件選択中」; gold pill button 「ルートを組み直す」 with 3D press (`:active` translate + shadow collapse); bar hidden while recomputing and after success until the selection changes again; E2 recompute renders only 「✓ 再計算 1.2s」 + card (no pipeline theater); superseded visual reuses the existing `.chat-card--superseded` (`opacity .55`).

Deviations (all token-driven):
1. Design specifies a dedicated gold shadow family (`--gold-deep` press shadow, `#5a3c20`-tinted soft card shadow) that has no token in `globals.css`. This card substitutes the neutral `var(--shadow-3d)` press family — a **different shadow hue than the design intends**, not an equivalent. Adding a `--color-gold-deep` token is a design-system decision left to the owner.
2. Design `.rebar` uses px paddings/radii → rem equivalents matching the file's scale (16px radius kept).
3. Design's tray also hosts chips + order segment (`.sptray`); those belong to later cards — this card ships the E2 recompute bar only. The syncRebar **minimum-selection rule is implemented** (below 2 picks: action disabled + 「2件以上選んでください」 in ja/zh/en).
4. Design shows the elapsed persisting on older recompute rows; ours shows elapsed on the latest settled turn only, matching the existing `SettledFootprint` precedent.

## Gates

- `pnpm run typecheck` ✅ · `pnpm run lint:oxlint` ✅ (10-line function cap respected)
- `pnpm test` ✅ 157 files / 969 tests; coverage **98.96 / 95.30 / 99.68 / 99.88** vs floors 95/94/95/95 (ratchet respected, not lowered)
- e2e: `web-chat-selection.spec.ts` 2/2 green against the local dev server
- Test quality: all new test files ≤200 lines, ≤5 mocks, no timing-dependent assertions (fake timers only)

## Review response (2026-07-29, both seats)

- **P1-1 (fixture false-green)** — accepted and fixed red-first. All fixtures (msw, e2e, unit) now carry the `tool-plan_selected` chunks the backend always streams for the bypass (`execute_selected_route` → `_ToolPartTranslator`); against them the old rail failed 3 tests (footprint never rendered, generic badge leaked). Fix: `isBypassTurn()` — a `plan_selected` card whose only tool part is the `plan_selected` step suppresses the pipeline and renders the footprint; agent-path multi-tool turns keep badges. e2e re-run green against the dev server with the realistic frames.
- **P1-2 (AC-9 orphan)** — pulled back into this card: `persist_messages` now skips the user-message insert when the turn carries no utterance (the bypass marker → `_last_user_text` = `""`), red-first pytest coverage (3 new tests, 1400 agent unit tests green, mypy/ruff clean). This also removes the empty-bubble symptom in reloaded conversation history (the empty row is never created). `upsert_conversation` remains untouched per the review's confirmation that `first_query` is not polluted.
- **P1-3 (mid-stream tray tap → concurrent POST)** — `trayStatus()` forces `busy` whenever the chat is `submitted`/`streaming`; pinned by a held-open-stream test.
- **P2 all applied** — marker wire-shape absolute assertion (`{role:"user",parts:[]}` — the shape the persistence skip keys on); `intentOf` exported from `supersession.ts` and reused (no near-duplicate); min-2 selection rule per design syncRebar; 44px action target; skeleton-appearance test on a controlled recompute stream; the settle watcher takes a props object.
- **P3/Fable follow-ups** — awaited body spy (no assertion race); `isLast` compared by id; dead `position: sticky` removed; `generateId` from the AI SDK; `role="status"` scoped to the count line; a persistent polite live region announces the recompute after the tray unmounts (a11y handoff). The e2e skeleton-order assertion stays jsdom-side (the controlled-stream test); Playwright `route.fulfill` cannot hold a stream open mid-body.
- **Mutation verification, re-run against the committed tree**: chat-busy gate dropped → P1-3 test red; min-2 disable dropped → tray test red; rail suppression dropped → 2 tests red; persistence guard dropped → pytest red. All killed.

## Dependency note

- ~~Task 3 dependency~~ — resolved: the `persistence.py` skip now ships in this PR (review P1-2). Task 3 no longer carries it.

### DIFF
--- a/apps/agent/agent/interfaces/persistence.py
+++ b/apps/agent/agent/interfaces/persistence.py
@@ -159,9 +159,13 @@ async def persist_messages(
     if insert_message is None:
         return
 
-    await _safe_insert_message(
-        insert_message, session_id, "user", user_text, label="insert_user_message"
-    )
+    # #273 T1: a selected_point_ids recompute carries no new utterance (the
+    # client sends a part-less marker; ``_last_user_text`` derives ""). Never
+    # persist an empty user row — history would render it as an empty bubble.
+    if user_text != "":
+        await _safe_insert_message(
+            insert_message, session_id, "user", user_text, label="insert_user_message"
+        )
 
     if persist_user_only:
         return
--- a/apps/agent/agent/tests/unit/test_persistence_helpers.py
+++ b/apps/agent/agent/tests/unit/test_persistence_helpers.py
@@ -11,7 +11,54 @@ from agent.interfaces.persistence import (
     _safe_insert_message,
     build_response_session,
     extract_plan_steps,
+    persist_messages,
 )
+from agent.interfaces.schemas import PublicAPIResponse
+
+
+class _Db:
+    def __init__(self) -> None:
+        self.insert_message = AsyncMock()
+
+
+def _response() -> PublicAPIResponse:
+    return PublicAPIResponse(
+        success=True, status="ok", intent="plan_selected", message="route"
+    )
+
+
+async def test_persist_messages_skips_empty_user_utterance() -> None:
+    """#273 T1: a bypass recompute carries no new utterance (marker ->
+    ``text == ""``) and must not persist an empty user row that the
+    conversation history would render as an empty bubble."""
+    db = _Db()
+    await persist_messages(
+        db=db, session_id="s1", user_text="", result=None, response=_response()
+    )
+    roles = [call.args[1] for call in db.insert_message.await_args_list]
+    assert roles == ["assistant"]
+
+
+async def test_persist_messages_inserts_genuine_user_turn() -> None:
+    db = _Db()
+    await persist_messages(
+        db=db, session_id="s1", user_text="ユーフォ", result=None, response=_response()
+    )
+    roles = [call.args[1] for call in db.insert_message.await_args_list]
+    assert roles == ["user", "assistant"]
+
+
+async def test_persist_messages_empty_utterance_failure_persists_nothing() -> None:
+    db = _Db()
+    await persist_messages(
+        db=db,
+        session_id="s1",
+        user_text="",
+        result=None,
+        response=_response(),
+        persist_user_only=True,
+    )
+    assert db.insert_message.await_args_list == []
 
 
 async def test_safe_insert_message_succeeds() -> None:
--- a/apps/web/src/features/chat/ChatPage.tsx
+++ b/apps/web/src/features/chat/ChatPage.tsx
@@ -103,10 +103,25 @@ function isInputLocked(props: ShellProps): boolean {
   return props.entry === "A5" || busy || historyBlocked;
 }
 
-/** The E2 sticky recompute bar docks above the composer (design `.dock`). */
-function RecomputeTray({ dict, recompute }: Readonly<{ dict: ChatDict; recompute: RecomputeTurn }>) {
+/** P1-3: the tray must not fire while ANY turn is in flight — the AI SDK's
+ * `makeRequest` has no concurrency guard, so a mid-stream tap would clobber
+ * the active response. Chat busy always reads as `busy` here. */
+function trayStatus(chat: ChatSession, recompute: RecomputeTurn): RecomputeTurn["status"] {
+  const active = chat.status === "submitted" || chat.status === "streaming";
+  return active ? "busy" : recompute.status;
+}
+
+type TrayHostProps = Readonly<{ dict: ChatDict; chat: ChatSession; recompute: RecomputeTurn }>;
+
+/** The E2 recompute bar docks above the composer (design `.dock`); the live
+ * region keeps announcing after the bar unmounts on fire (a11y handoff). */
+function RecomputeTray({ dict, chat, recompute }: TrayHostProps) {
+  const status = trayStatus(chat, recompute);
   return (
-    <SelectionTray dict={dict} status={recompute.status} lastSentIds={recompute.lastSentIds} onRecompute={recompute.fire} />
+    <>
+      <span className="chat-live-note" aria-live="polite">{status === "busy" ? dict.preparing : ""}</span>
+      <SelectionTray dict={dict} status={status} lastSentIds={recompute.lastSentIds} onRecompute={recompute.fire} />
+    </>
   );
 }
 
@@ -124,7 +139,7 @@ function ChatShell(props: ShellProps) {
     <main className="chat-page">
       <ShellNotices {...props} />
       <ChatBody {...props} />
-      <RecomputeTray dict={props.dict} recompute={props.recompute} />
+      <RecomputeTray dict={props.dict} chat={props.chat} recompute={props.recompute} />
       <ChatInput dict={props.dict} disabled={isInputLocked(props)} onSend={props.onSend} />
     </main>
   );
--- a/apps/web/src/features/chat/ChatPage.tsx
+++ b/apps/web/src/features/chat/ChatPage.tsx
@@ -7,6 +7,7 @@ import type { ChatActions } from "./chat-actions";
 import { ChatInput } from "./components/ChatInput";
 import { ColdStart } from "./components/ColdStart";
 import { ErrorBanner } from "./components/ErrorBanner";
+import { SelectionTray } from "./components/SelectionTray";
 import { TurnFailure } from "./components/ErrorStates/TurnFailure";
 import type { TurnFailureView } from "./components/ErrorStates/TurnFailure";
 import { HistoryList } from "./components/HistoryList";
@@ -18,6 +19,9 @@ import type { ChatEntryState } from "./entry-state";
 import { chatDictFor } from "./i18n";
 import type { ChatDict } from "./i18n";
 import type { ChatSearch } from "./search";
+import { SpotSelectionProvider, useSpotSelectionState } from "./selection/useSpotSelection";
+import { useRecomputeTurn } from "./selection/useRecomputeTurn";
+import type { RecomputeTurn } from "./selection/useRecomputeTurn";
 import { useAutoSend } from "./use-auto-send";
 import { useBackendHealth } from "./use-backend-health";
 import type { BackendHealth } from "./use-backend-health";
@@ -39,6 +43,7 @@ type ShellProps = Readonly<{
   chat: ChatSession;
   history: ConversationHistory;
   failure: TurnFailureView | undefined;
+  recompute: RecomputeTurn;
   onRetry: () => void;
   onSend: (text: string) => void;
 }>;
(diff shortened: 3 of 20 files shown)


lifeodyssey (commented):
## Fable independent review (seat 2) — #273 Task 1, spec fa8d4448

**Verdict: request_changes** — gates re-run green independently (typecheck/oxlint/969 tests, coverage 98.96/95.30/99.68/99.88 >= floors), both spot-check mutations killed (marker-append drop; empty-guard drop), but two P1s.

### P1-1 (correctness): the footprint/no-badges AC fails against the real backend; green only on fabricated fixtures
The real bypass path **does** emit tool frames: `apps/agent/agent/agents/selected_route.py:37-88` fires `StepEvent("plan_selected", ..., running/done)` through `on_step`, and `chat_stream.py::_ToolPartTranslator` turns every step into `ToolInputStart/ToolInputAvailable/ToolOutputAvailable` chunks. So the real recompute assistant message carries `tool-plan_selected` parts, and `MessageList.tsx::MessageRail`'s gate `tools.length === 0 && hasRecomputePart(...)` routes it to `Pipeline` — the user gets a step badge + a generic settled details row, **not** the recompute footprint, and the structural "tool-badge selector absent" AC fails. Every passing test depends on fixtures that strip tool frames by hand (`chat-handlers.ts::stripToolFrames`, and the same filter inline in `e2e/web-chat-selection.spec.ts`); the e2e "against the local dev server" intercepts `**/v1/chat` with `page.route`, so the real stream was never exercised. (My mutation-A run incidentally rendered exactly this failure mode: tool badges + plan_selected card, no recompute footprint.)
**Fix:** gate on `hasRecomputePart` first (footprint whenever the turn is a plan_selected bypass, tool parts or not) or filter `tool-plan_selected` parts — and re-record the MSW/e2e fixture from the true wire (with the plan_selected tool frames) so tests can no longer diverge from production.

### P1-2 (scope/quality ratchet): AC-9 server half exists nowhere; the dependency declaration is inaccurate
Spec Task 1's file list **includes** `apps/agent/agent/interfaces/persistence.py` ("required by the history-integrity AC"). The PR defers it to Task 3 — but `origin/feat/273-t3-migration-retention` tip (`a81ad555`) contains **no** skip-empty-utterance change: `persist_messages` (line ~163) still inserts the user row unconditionally. So AC-9's integration test is not in this diff and its implementation is on no branch; quality ratchet is effectively 8/9 in-diff. Interim user-visible defect confirmed: each recompute persists an empty-content user row, and `use-conversation-history.ts` + `HistoryList.tsx` render it as an empty `chat-bubble` on session reload (no empty-content filter).
**Fix:** land the persistence skip here per spec, or get Coordinator sign-off on the re-scope + correct the dependency note + add an interim empty-entry filter in `HistoryList`.

### P2
1. **Tray fires during a live agent turn** — `trayHidden` only checks the recompute status; while a normal turn streams, the tray is visible and `fire()` calls `sendMessage` mid-stream. Worse, `useRecomputeTurn`'s settle watcher sets `started.current` on the *previous* turn's activity and can settle the recompute verdict against the wrong turn. Disable/hide the tray while chat status is submitted/streaming (mirror `isInputLocked`).
2. **Three CI checks red** (SAST / credential scan / lockfile CVE) — all three died on the same octokit `GET /pulls/461/commits` API error, i.e. infra, not findings; still must be rerun green before merge.
3. **Latent flake** — `chat-page-selection.test.tsx` spy pushes bodies via un-awaited `request.clone().json().then(...)`; `expect(bodies).toHaveLength(2)` races the parse. Await the parse in the handler or `waitFor` on the length.

### P3
1. Static supersession check: solid for its purpose (both halves mutation-killed), but evadable via template-literal class construction and scans only `.ts/.tsx` — note for the record.
2. AC1's "skeleton **then** card" ordering is never asserted (only skeleton-absence after settle); acceptable under the spec's anti-time-sampling ruling, but the skeleton's appearance itself is unverified.
3. A11y nit: firing the recompute unmounts the tray while its button holds focus -> focus drops to body. Consider focus handoff or an aria-live note. (Otherwise a11y is fine: labeled checkboxes, real button, `role="status"`.)
4. `MessageList` `isLast={message === last}` reference-compares against the possibly invisible marker; works, but id-based comparison would be sturdier.

### Verified good
- Bypass request shape: exactly one extra POST, `selected_point_ids` present, no new user turn; `chat.py::_user_message_text` tolerates the part-less marker (`parts: []` -> empty text) and `_dispatch_request` takes the point-selection branch first — all verified in source on this base.
- `plan_selected` intent is real (`selected_route.py:117/127`, contract `chat-data-parts.ts:88`) — detection isn't invented.
- Design tokens all exist (`--color-gold`, `--color-gold-fg`, `--shadow-3d`-as-color per precedent, `--color-card`); BEM naming matches chat.css; no hardcoded palette.
- i18n ja/zh/en incl. count interpolation tested; test files <=200 lines, <=5 mocks, mocked-clock exact-string assertion (no sleeps).
- Mutations re-verified independently: (A) drop marker append -> footprint/card test fails (killed); (B) drop empty-selection guard -> 2 tests fail (killed).

```json
{
  "verdict": "request_changes",
  "eval_results": { "test_eval": "skipped" },
  "quality_ratchet": { "ac_total": 9, "ac_with_test": 8, "missing": ["AC-9 history integrity (server half + integration test on no branch)"] }
}
```

lifeodyssey (commented):
## Fable seat 2 — final verification round, HEAD `86ee6764`

**Verdict: APPROVE.**

All findings from my first pass re-verified independently in a fresh detached worktree at `86ee6764`:

### P1-1 (fixture diverged from production) — fixed and load-bearing
- `isBypassTurn` (`selectedPointsBypass.ts`) now keys on the plan_selected card + suppresses only the bypass's own `tool-plan_selected` step part; agent-path turns that ran other tools keep their badges (tested).
- MSW (`chat-handlers.ts::toRecomputeStream`) and e2e fixtures now carry the true wire shape — the `plan_selected` running/done tool chunks `chat_stream._ToolPartTranslator` actually emits.
- **Regression-mutation verified**: reverting `MessageRail` to the old `tools.length === 0 &&` gate against the new fixtures makes "shows the recompute footprint and zero step badges" fail (footprint null). 先红后绿 confirmed — the fixtures can no longer certify the wrong tree.

### P1-2 (AC-9 ownerless + empty-bubble symptom) — fixed at the root, in-card per spec
- `persistence.py::persist_messages` now skips the user insert on an empty utterance; 3 new backend unit tests (`test_persistence_helpers.py`) pass — ran `uv run pytest` independently, 10/10.
- Client half: the marker's wire shape (`role: "user"`, `parts: []`) is now asserted on the recompute body, which is exactly what the skip keys on. The empty-bubble history symptom disappears with the root cause.

### P2 (tray mid-stream race) — fixed
- `trayStatus` maps any active chat turn to `busy` → tray hidden, cannot fire concurrently; jsdom test with a held-open stream proves it. This also closes the settle-watcher wrong-turn race, since `fire()` is unreachable while another turn is active.

### Small items — all landed
- Spy race → `waitFor` on `bodies.length`; `isLast` by id; a11y handoff via persistent polite `.chat-live-note` (+ 44px touch target, `role="status"` on the count); skeleton-appearance now positively asserted via the controlled MSW stream (`chatRecomputeControlledHandler`), with the Playwright `route.fulfill` limitation documented — accepted.
- Bonus items are sound: `intentOf` unified into `supersession.ts` (single implementation), min-2-selection disable matches design syncRebar with i18n copy tested, `generateId()` over `crypto.randomUUID`.

### Independent gates at `86ee6764`
- `pnpm run typecheck` 0 · `lint:oxlint` 0 · full web suite green, coverage 98.97 / 95.41 / 99.68 / 99.88 (floors respected, up vs previous head).
- Agent: `test_persistence_helpers.py` 10 passed.

### For the merge queue (not this PR's fault)
- **Semgrep red is inherited from base**: the 1 blocking finding is `apps/web/src/features/seo/JsonLd.tsx` (react-dangerouslysetinnerhtml) — untouched by this diff.
- **gitleaks / osv-scanner red is infra**: octokit `GET /pulls/461/commits` → 403 "Resource not accessible by integration" (workflow permissions), on both runs. Needs a repo-level fix or an authorized rerun; nothing in this diff touches secrets or lockfiles.
- P3 (record only): AC-9's end-to-end row-count *integration* test (follow-up + two recomputes → persisted user rows == genuine turns) still exists nowhere; the mechanism is unit-proven at both halves, so non-blocking — worth a card-level note.

```json
{
  "verdict": "approve",
  "eval_results": { "test_eval": "skipped" },
  "quality_ratchet": { "ac_total": 9, "ac_with_test": 9, "missing": [] }
}
```

Issue #268, opened 2026-07-07
S2.3 — Gold CTA logic + sticky dock + Walk entry point


lifeodyssey:
## Releasable statement
The route detail page's single gold CTA changes its copy and action based on the route's data state (share on a weekday, go to Walk Mode today, or make a keepsake after completion), and the cream secondary CTA jumps to the reference-edit chat flow. This also stands up the "Depart in Walk Mode" entry point.

## AC summary
- AC count: 5
- Test type distribution: unit=2, integration=0, browser=3, eval=0, api=0, manual/ops=0

## Dependencies
S2.1.

## Spec source
`docs/superpowers/specs/2026-07-06-frontend-rebuild/iter-2.md` (commit 6fc13b8)

### S2.3 Gold CTA logic + sticky dock + Walk entry point #2

**User story**: As a user, I want the page's single gold CTA to always say the right thing for the moment (share on a weekday / go walk today / make a keepsake しおり after completing) — and I want this to be an actually-working 「歩くモードで出発」("Depart in Walk Mode") entry point (one of the three entry points Q4 requires).

**Design basis**: `spec-route-detail.md` §3 Gold CTA.

**Releasable statement**: the gold CTA's copy/action switches with the data state (weekday = share the しおり / today = go to Walk Mode / completed = make a commemorative しおり); the cream secondary CTA (編集する, "edit") jumps to the Chat A2b reference-edit flow; the sticky dock's gradient handoff is correct.

**AC**:
- Happy path: for a route dated today, the gold CTA reads "歩くモードへ" ("to Walk Mode"), and clicking it navigates to the Walk target route (this iteration only verifies navigation intent / that the route target exists — the real Walk screens ship in Iteration 3, so a placeholder here is acceptable) -> browser
- Happy path: for a completed route, the gold CTA reads "記念しおりを作る" ("make a commemorative しおり") (deep-linking to the しおり flow's entry point from Iteration 4 — a placeholder here is acceptable) -> browser
- Empty: the cream secondary CTA (編集する) gracefully disables when the route has no chat origin to reference back to -> unit
- Error: if the data state changes mid-press (e.g. a check-in happens to complete the route right then), the CTA never gets stuck showing stale/incorrect copy -> browser
- i18n: all three CTA copy variants render in ja/zh/en -> unit

**Changed files**: `apps/web/src/components/route-detail/StickyDock.tsx`, `apps/web/src/components/route-detail/GoldCta.tsx`.

[...]

PR #462, opened 2026-07-28
feat(#273 T2): save CTA, P5 login wall, deferred intent + create-on-login

lifeodyssey:
Implements **#273 S1.7 Task 2** — the 「保存する」 CTA, the P5 login-wall trigger, the deferred save intent, and create-on-login. Spec: `docs/superpowers/specs/2026-07-28-s1.7-living-document-save-login-wall-design.md` @ `fa8d4448` (branch `docs/spec-273-s1.7`).

## AC coverage — 12/12, every AC has a test

| # | AC | Type | Test |
|---|---|---|---|
| 1 | Anonymous tap opens the magic-link `LoginModal` | browser | `e2e/web-chat-save-login-wall.spec.ts` |
| 2 | Signed-in tap: `saveRoute` once, `status: "saved"`, non-empty `point_ids` == stop count, no dialog, saved confirmation | unit | `tests/unit/chat/save-gate.test.tsx` |
| 3 | Create-on-login: row content + `listRoutes` visibility, ids in order | integration | `tests/integration/create-on-login.test.ts` |
| 4 | Deferred intent lifecycle across a real navigation in a new page of the same profile; deleted after replay; a non-save login replays nothing | browser | `e2e/web-chat-save-login-wall.spec.ts` (+ unit twins in `deferred-save.test.ts`, `save-login-carveout.test.tsx`) |
| 5 | TTL: an abandoned intent is ignored **and cleared** | unit | `tests/unit/chat/deferred-save.test.ts` |
| 6 | No route yet ⇒ CTA disabled, not actionable, opens nothing, writes nothing | unit | `tests/unit/chat/save-gate.test.tsx` |
| 7 | Network/5xx failure ⇒ inline retryable error; card survives; intent not dropped; no full-page error, no forced logout | unit | `save-gate.test.tsx`, `deferred-save.test.ts` |
| 8 | P5 trigger timing: no `role="dialog"` login node before the tap across the anonymous flow | browser | `e2e/web-chat-save-login-wall.spec.ts` |
| 9 | D8/D11 carve-out: both still open on their own tap, neither fires on a happy-path step | unit | `tests/unit/chat/save-login-carveout.test.tsx` |
| 10 | Server-side flow returns no 401/403 and carries no `Authorization` header | integration | `apps/agent/.../test_anonymous_flow_uninterrupted.py` |
| 11 | Title derivation: non-empty, within 1–200, work title + stop count, localized, no untranslated fallback | unit | `tests/unit/chat/save-copy-i18n.test.ts` |
| 12 | i18n: CTA / confirmation / error copy in ja·zh·en | unit | `tests/unit/chat/save-copy-i18n.test.ts` |

`ac_total == ac_with_test` = 12 == 12 (unit 7, integration 2, browser 3) — matches the spec's ratchet.

## Design-sync alignment

Baseline: `docs/design/2026-07-06-design-sync/Chat 完整状态.html` and `DS 补全 - Chat 桌面.html`.

- **CTA placement** — the design's card-foot CTA row is `しおりにして共有 / Mapsで開く / 保存する / 歩くモード`; 保存する lands in `TimedItinerary`'s `chat-cta-row`, between Maps and the dormant Walk seam. `RouteCard.tsx` renders no button (the spec's corrected siting).
- **CTA colour — deliberately cream, not gold.** The design稿 assigns `金=しおり共有（唯一）· cream=Maps/保存/歩くモード` with the standing rule `永不同屏两金`, and the DS token board lists 保存する under `Primary cream`. So the CTA reuses `.chat-chip`'s cream press style (`--color-card` / `--color-fg`, 3D shadow), and a CSS test asserts it never borrows a gold token. The #268 gold-CTA semantics are respected by *reserving* gold, not by spending it here. **Flagging explicitly** since the card brief mentioned a gold CTA — if the owner wants gold on 保存する, the design system requires demoting しおり共有 on the same screen.
- **Feedback states** — inline saved confirmation (`--color-primary-strong`) and an inline retryable error (`--color-error-strong`), matching the design's `ルートを保存しました ✓` toast semantics as an in-card affordance. No hardcoded palette values anywhere.
- `LoginModal` is reused as-is; it mounts only while open, so the P5 invariant is structural rather than merely asserted.

## Implementation notes

- **Deferred intent = namespaced `localStorage` + 30-min TTL** (`animichi.chat.deferredSave.v1`). `sessionStorage` is tab-scoped and is provably empty in the magic-link's new browsing context; URL state is rejected for the point-id leak into email/referrer chains. Deleted on success **or** expiry; kept on failure so a failed replay is never silently dropped.
- **One login-wall predicate** — `saveAction(target, authStatus)` in `useSaveGate.ts`. An unresolved auth status is inert (no false interruption for a signed-in caller).
- **Create-on-login, no claim** — `replayDeferredSave` calls `users.saveRoute` with the client-held ids and never sends an `id`, so `ROUTE_NOT_OWNED` is structurally unreachable. `users.claimRoutes` stays unwired.
- `use-save-route.ts` is a plain state hook rather than a TanStack mutation because the auth-callback replay runs outside any `QueryClientProvider`; the UI→hook→client direction from `apps/web/AGENTS.md` is preserved.

## Mutation verification — 10 mutants, 10 killed

| Mutation | Result |
|---|---|
| Delete the signed-in branch (wall always opens) | 5 failed |
| Delete the empty/absent-target guard | 3 failed |
| `localStorage` → `sessionStorage` (the falsified design) | 5 failed |
| Delete the TTL check | 2 failed |
| Replay even without an intent | 2 failed |
| Clear the intent after a *failed* replay | 1 failed |
| Mount `LoginModal` unconditionally | 5 failed |
| Paint the save CTA gold | 2 failed |
| Skip the replay in `useAuthCallback` | 2 failed |
| Untranslated title template | 1 failed |

## Gates

- `apps/web` typecheck ✅ · `lint:oxlint` (type-aware, deny-warnings) ✅
- `apps/web` unit: **982 passed** · coverage **98.82 / 95.27 / 99.67 / 99.82** vs floors 95/94/95/95 ✅ (new `features/chat/save` module: 95.52/96.29/100/98.03)
- `apps/web` integration: 11 passed ✅ · `pnpm run test:worker`: 97 passed ✅
- `apps/agent`: `ruff check` + `ruff format` ✅; the new integration test passes against the pinned Atlas 0.30.0 + testcontainer ✅
- `e2e`: the new spec compiles and lists 4 tests; added to `test:web`. Browser ACs are for the Tester against staging/local (`ANON_ACCESS_ENABLED=false` in production), and the lifecycle test needs `VITE_NEON_AUTH_BASE_URL` set for the callback to redeem.
- Every new test file ≤200 lines; production functions ≤10 lines; **no suppressions added**.

### Notes for the reviewer

- **Coverage ratchet not raised.** Actual coverage already sat well above the floor before this card (98.73/95.07/…), and Tasks 1 and 3 are in flight on the same base — raising the config mid-wave would break their branches on rebase. Flagging for the Coordinator to ratchet once all three land.
- **Local env caveat:** DB-backed agent integration tests need Atlas pinned to 0.30.0; this machine ships 1.2.4, which fails *all* such tests (pre-existing, reproduced on `test_conversation_hydration_contract.py`). Verified with a checksum-matched 0.30.0 binary on a temporary PATH.
- **Concurrency:** `apps/web/src/styles/chat.css`, `tests/unit/chat/chat-design-css.test.ts` and `tests/msw/users.ts` are shared with Task 1's expected edits (tray copy/CSS). Changes here are additive blocks at distinct locations. `TimedItinerary.tsx` is owned by this task per the spec's file allocation; `RouteCard.tsx` gains two lines to hand the CTA its save payload.
- **Not a bug (spec §7):** a magic link opened in a *different* browser leaves the intent behind in the originating browser's `localStorage`, so the save does not auto-fire.

Refs #273 Task 2. **Do not merge** — hand off to Reviewer.

### DIFF
--- a/apps/agent/agent/tests/integration/test_anonymous_flow_uninterrupted.py
+++ b/apps/agent/agent/tests/integration/test_anonymous_flow_uninterrupted.py
@@ -3,9 +3,13 @@ interrupted by an auth challenge.
 
 The browser ACs pin that no login dialog opens before the 「保存する」 tap. This is
 the server-side half of the same invariant: the identical flow driven through
-the runtime endpoint returns no 401/403 and carries no ``Authorization`` header
-on any request, so the login wall is a product decision made in the client
-rather than something the backend forces.
+the runtime endpoint returns no 401/403, so the login wall is a product decision
+made in the client rather than something the backend forces.
+
+Every assertion here is on **app** behaviour — the route's declared security, the
+`_reject_credentialed_anonymous` guard, and the missing-identity status — rather
+than on headers the test itself constructed, which no production change could
+falsify.
 
 Uses ``httpx.AsyncClient`` + ``ASGITransport`` (the pattern in
 ``test_runtime_journey_contract.py``) because ``TestClient``'s portal thread
@@ -117,8 +121,52 @@ async def test_anonymous_flow_returns_no_auth_challenge(anon_client) -> None:
 
 
 @pytest.mark.integration
-async def test_anonymous_flow_sends_no_authorization_header(anon_client) -> None:
-    for text in _FLOW:
-        resp = await _turn(anon_client, text)
-        sent = {name.lower() for name in resp.request.headers}
-        assert "authorization" not in sent
+async def test_runtime_declares_no_bearer_requirement(anon_client) -> None:
+    """The route must not *demand* a credential the anonymous client cannot mint.
+
+    Asserted against the app's own OpenAPI schema rather than against headers the
+    test constructed: adding an `HTTPBearer` dependency to `/v1/runtime` — the
+    production change that would break the anonymous flow — turns this red.
+    """
+    schema = (await anon_client.get("/openapi.json")).json()
+    operation = schema["paths"]["/v1/runtime"]["post"]
+    assert not operation.get("security"), operation.get("security")
+    assert not schema.get("components", {}).get("securitySchemes")
+
+
+@pytest.mark.integration
+async def test_anonymous_stamp_with_a_credential_is_rejected(anon_client) -> None:
+    """The app enforces the flow's no-`Authorization` property; the test does not
+    merely restate its own request headers.
+
+    `_reject_credentialed_anonymous` (issue #441) refuses an anonymous stamp that
+    arrives with a credential, so a client that leaked a stale bearer into the
+    anonymous flow would be 401'd — which is exactly the interruption the P5
+    invariant forbids. Deleting that guard, or having the web client attach a
+    header while signed out, turns this red.
+    """
+    resp = await anon_client.post(
+        "/v1/runtime",
+        json={"text": _FLOW[0], "locale": "zh"},
+        headers={**_ANON_HEADERS, "Authorization": "Bearer stale-token"},
+    )
+    assert resp.status_code == 401
+
+    clean = await _turn(anon_client, _FLOW[0])
+    assert clean.status_code == 200
+
+
+@pytest.mark.integration
+async def test_identityless_request_is_never_auth_challenged(anon_client) -> None:
+    """Even with no edge-stamped identity at all, the route answers rather than
+    challenging: no 401/403 and no ``WWW-Authenticate`` for the client to react
+    to. Adding an authentication dependency here — the change that would put a
+    login in front of the anonymous flow — turns this red. The exact success
+    status is deliberately not pinned, so a future stricter *request* validation
+    stays free to return 4xx without faking an auth challenge.
+    """
+    resp = await anon_client.post(
+        "/v1/runtime", json={"text": _FLOW[0], "locale": "zh"}
+    )
+    assert resp.status_code not in _AUTH_CHALLENGES
+    assert "www-authenticate" not in {name.lower() for name in resp.headers}
--- a/apps/agent/agent/tests/integration/test_anonymous_flow_uninterrupted.py
+++ b/apps/agent/agent/tests/integration/test_anonymous_flow_uninterrupted.py
@@ -1,0 +1,122 @@
+"""P5 login wall (issue #273 S1.7 Task 2): the anonymous happy path is never
+interrupted by an auth challenge.
+
+The browser ACs pin that no login dialog opens before the 「保存する」 tap. This is
+the server-side half of the same invariant: the identical flow driven through
+the runtime endpoint returns no 401/403 and carries no ``Authorization`` header
+on any request, so the login wall is a product decision made in the client
+rather than something the backend forces.
+
+Uses ``httpx.AsyncClient`` + ``ASGITransport`` (the pattern in
+``test_runtime_journey_contract.py``) because ``TestClient``'s portal thread
+conflicts with asyncpg's event loop.
+"""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock, patch
+
+import httpx
+import pytest
+
+from agent.agents.agent_result import AgentResult, StepRecord
+from agent.agents.runtime_models import ClarifyResponseModel, QAResponseModel
+from agent.agents.session_state import (
+    OrderedCandidate,
+    PendingClarification,
+    SessionState,
+)
+from agent.infrastructure.session.memory import InMemorySessionStore
+from agent.interfaces.fastapi_service import create_fastapi_app
+from agent.interfaces.public_api import RuntimeAPI
+
+# The edge stamps anonymous callers with this exact wire format
+# (``worker/anonymous.test.ts`` pins ``^anon_[0-9a-f]{32}$``).
+_ANON_HEADERS = {
+    "X-User-Id": "anon_0123456789abcdef0123456789abcdef",
+    "X-User-Type": "anonymous",
+}
+
+# search -> clarify -> route refinement -> conversational follow-up.
+_FLOW = ("宇治站附近的圣地", "凉宫", "帮我规划路线", "你好")
+
+_AUTH_CHALLENGES = (401, 403)
+
+
+def _clarify_result() -> AgentResult:
+    candidates = [
+        OrderedCandidate(id="11291", title="凉宫春日的忧郁", points_count=2, city="西宫")
+    ]
+    state = SessionState(
+        pending_clarification=PendingClarification(
+            reason="anime_ambiguity",
+            candidate_ids=["11291"],
+            ordered_candidates=candidates,
+            revision=1,
+        ),
+        clarification_revision=1,
+    )
+    output = ClarifyResponseModel(
+        reason="anime_ambiguity", message="你是指哪部凉宫？", candidate_ids=["11291"]
+    )
+    return AgentResult(
+        output=output,
+        intent="clarify",
+        session_state=state,
+        steps=[StepRecord(tool="clarify", success=True, model_initiated=False)],
+    )
+
+
+def _qa_result() -> AgentResult:
+    return AgentResult(
+        output=QAResponseModel(message="我可以帮你找动漫圣地。"),
+        intent="general_qa",
+        session_state=SessionState(),
+    )
+
+
+async def _fake_agent(**kwargs: object) -> AgentResult:
+    text = str(kwargs.get("text", ""))
+    return _clarify_result() if "凉宫" in text else _qa_result()
+
+
+@pytest.fixture
+async def anon_client(tc_db):
+    runtime_api = RuntimeAPI(
+        tc_db, session_store=InMemorySessionStore(), model_http_client=MagicMock()
+    )
+    app = create_fastapi_app(runtime_api=runtime_api, db=tc_db)
+    app.state.runtime_api = runtime_api
+    app.state.db_client = tc_db
+    transport = httpx.ASGITransport(app=app, raise_app_exceptions=False)
+    with patch(
+        "agent.interfaces.public_api.run_animichi_agent", side_effect=_fake_agent
+    ):
+        async with httpx.AsyncClient(
+            transport=transport, base_url="https://test"
+        ) as client:
+            yield client
+
+
+async def _turn(client: httpx.AsyncClient, text: str) -> httpx.Response:
+    return await client.post(
+        "/v1/runtime", json={"text": text, "locale": "zh"}, headers=_ANON_HEADERS
+    )
+
+
+@pytest.mark.integration
+async def test_anonymous_flow_returns_no_auth_challenge(anon_client) -> None:
+    for text in _FLOW:
+        resp = await _turn(anon_client, text)
+        assert resp.status_code not in _AUTH_CHALLENGES, (
+            f"anonymous turn {text!r} was auth-challenged with {resp.status_code}"
+        )
+        assert resp.status_code == 200
+
+
+@pytest.mark.integration
+async def test_anonymous_flow_sends_no_authorization_header(anon_client) -> None:
+    for text in _FLOW:
+        resp = await _turn(anon_client, text)
+        sent = {name.lower() for name in resp.request.headers}
+        assert "authorization" not in sent
--- a/apps/agent/agent/tests/integration/test_anonymous_flow_uninterrupted.py
+++ b/apps/agent/agent/tests/integration/test_anonymous_flow_uninterrupted.py
@@ -45,7 +45,9 @@ _AUTH_CHALLENGES = (401, 403)
 
 def _clarify_result() -> AgentResult:
     candidates = [
-        OrderedCandidate(id="11291", title="凉宫春日的忧郁", points_count=2, city="西宫")
+        OrderedCandidate(
+            id="11291", title="凉宫春日的忧郁", points_count=2, city="西宫"
+        )
     ]
     state = SessionState(
         pending_clarification=PendingClarification(
--- a/apps/web/AGENTS.md
+++ b/apps/web/AGENTS.md
@@ -55,8 +55,10 @@ landing page and branded 404 through a Cloudflare SSR bundle. Root guide: `../..
 - `src/routeTree.gen.ts` is generated. Do not hand-edit it; oxlint and coverage ignore it. The unit
   pool runs without the TanStack Start vite plugin, so `tests/setup/generate-route-tree.ts`
   (vitest `globalSetup`) emits it before the suite; a normal build regenerates it otherwise.
-- Coverage sweeps `src/**` (routes + `router.tsx` included, per campaign plan §0.6); the floor is
-  90/90/90/90 — ratchet UP only. `routeTree.gen.ts` is the only exclusion.
+- Coverage sweeps `src/**` (routes + `router.tsx` included, per campaign plan §0.6); the enforced
+  floor is `statements 95 / branches 94 / functions 95 / lines 95`, read from `vitest.config.ts`,
+  which is the authority — ratchet UP only. `routeTree.gen.ts` is the only exclusion (plus the
+  WebGL map glue listed in the config's exclude ledger).
 - `OpenAPILink` captures `globalThis.fetch` at construction: build clients at call time (the lazy
   `catalog()` / `users()` singletons do), not at module top, or MSW patching is missed in tests.
 - A normal build must run before preview or build-output integration checks inspect `.output/`.
--- a/apps/web/src/api/hooks/use-save-route.ts
+++ b/apps/web/src/api/hooks/use-save-route.ts
@@ -1,0 +1,37 @@
+import { useCallback, useState } from "react";
+import type { SaveRouteInput, UserRoute } from "@seichijunrei/contract";
+import { users } from "../orpc";
+
+/** The one transport call behind both the in-chat save and create-on-login. */
+export type SaveRouteRequest = (input: SaveRouteInput) => Promise<UserRoute>;
+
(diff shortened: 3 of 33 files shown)


lifeodyssey (commented):
## Reviewer (Fable seat, independent) — PR #462 / #273 Task 2

**Verdict: approve** (P2s are non-blocking; two are cheap in-PR fixes if the executor is still around).

Baseline: spec `origin/docs/spec-273-s1.7` @ fa8d4448, Task 2 (12 ACs). Independent detached worktree at PR head 2be9436c.

### Gates (run independently)
- `apps/web` typecheck: green. oxlint: green. Unit: 157 files / 982 passed. Integration (`vitest.integration.config.ts`): green incl. `create-on-login.test.ts`. Coverage 98.82/95.27/99.67/99.82 vs floors 95/94/95/95 — no ratchet drift; AGENTS.md 90→95/94/95/95 correction matches `vitest.config.ts` (P1-3 applied).
- Mutation re-verification (3/3 killed): (1) remove clear-on-expiry in `readDeferredSave` → 1 test fails; (2) clear intent unconditionally in `replayDeferredSave` → "keeps the intent after a failed replay" fails; (3) `saveAction` anonymous→"save" → 4 save-gate tests fail. Consistent with the executor's 10/10 claim.
- `make test-eval`: **skipped** — diff adds only an agent integration test; no agent behavior/prompt/tool change.
- Browser ACs (e2e specs) not executed here (no running app) — Tester must run `e2e/web-chat-save-login-wall.spec.ts` (wired into `test:web`).

### Deep checks
1. **Deferred-intent lifecycle** — namespaced key `animichi.chat.deferredSave.v1`, 30min TTL, cleared on success and on expiry, retained on failure. The failure-retention choice is spec-compliant ("not silently dropped") and the negative AC ("a login not initiated by a save tap finds no intent") stays sound *because* an intent can only exist after a genuine 保存する tap — an unrelated login can only ever complete a save the user actually initiated within the last 30min. Clearing on modal close would be WRONG (the user closes the modal to go read the magic-link email), so TTL-as-abandonment is the correct operationalization of the spec's "deleted on success or abandonment". Residual semantics gap noted as P2-1/P2-2 below.
2. **create-on-login** — `toSaveInput` sends `{title, point_ids, status:"saved"}`, never `id`; verified against `packages/contract/src/users-contract.ts:76` (title 1–200, point_ids ≤500×≤128) and `workers/users/src/api/routes.ts:117` (`input.id ? update : create`) — with no `id`, `ROUTE_NOT_OWNED` is genuinely unreachable; the integration test additionally pins `sentId === undefined` and asserts row content + listRoutes echo, not mere call-made. Auth header: the `users()` client attaches `authHeaders` per call (`apps/web/src/api/orpc.ts`), and the replay runs only after `establish()` resolves in `useAuthCallback` — ordering correct.
3. **P5 invariant** — `LoginModal` is conditionally *mounted* (`gate.loginOpen ? <LoginModal open/> : null`), a structural guarantee, not just `open=false`. Single predicate `saveAction` with pending→none (no false interruption). D8/D11 carve-out has its own unit file.
4. **Design ruling (cream vs gold)** — the implementer's override of the dispatch brief is **correct**. DS ledger `docs/design/2026-07-06-design-sync/docs/user-journey.md:153`: 「しおりにして共有」(金色,主) / Maps (cream) / **「保存する」(cream)**; plus 「一屏唯一金色主行动 / 永不同屏两金」 (DS补全 Chat 桌面). CSS test reliability: all gold tokens are literally named `--color-gold*`, so `not.toContain("gold")` is a real guard, and the positive assertion pins `var(--color-card)` exactly — a hardcoded gold hex would slip the negative but not the positive.
5. **Browser lifecycle test** — `context.newPage()` + real `goto("/auth/callback")` is a genuine cross-navigation in the same profile (shared origin storage), and the spec explicitly rules that a fresh `BrowserContext` would be the wrong test. It also pins sessionStorage-empty in the new tab. P5 timing e2e has no timing assertions.
6. **Conflict surface with #461** — #461 does **not** touch `TimedItinerary.tsx`; the only overlap is `apps/web/src/styles/chat.css` (both append near the `.chat-cta-row` block; trivial textual conflict at worst). Merge order free.

### Findings

**P2-1** (`apps/web/src/components/auth/useAuthCallback.ts`, `redeem`) — the replay's boolean is discarded: a failed create-on-login is invisible to the user (callback still shows "done", no retry surface). The intent is retained per spec, but nothing will replay it except another login within 30min — which then fires a save the user long stopped waiting for. Suggest: surface the replay failure on the callback page. test: unit.

**P2-2** (semantics, same cluster) — after a failed replay or an abandoned wall, a D8/D11-initiated login inside the TTL silently completes the earlier save. Defensible (the user did tap save), but undocumented; the AC4-negative unit tests only cover the no-intent population. Suggest one unit test pinning the *intended* behavior of "unrelated login while a live intent exists", so the semantics are chosen rather than accidental. test: unit.

**P2-3** (`apps/agent/agent/tests/integration/test_anonymous_flow_uninterrupted.py::test_anonymous_flow_sends_no_authorization_header`) — vacuous: it asserts on the headers of the request the test itself constructed; no production mutation can fail it. The 401/403 twin is the meaningful half of AC10. Drop or reframe (e.g., assert the app-side auth dependency was never invoked). test: none.

**P2-4** (CI) — 3 Security jobs red on this PR, **none attributable to this diff**: Semgrep blocking finding is in `apps/web/src/features/seo/JsonLd.tsx` (untouched); osv-scanner is lockfile CVEs (lockfile untouched); gitleaks died on an octokit API error (infra), not a leak. Coordinator: confirm base-red and track separately.

**P3-1** (`apps/web/src/features/chat/route-copy.ts::saveRouteTitle`) — `String.replace("{title}", work)` interprets dollar-patterns in the work title (injection via `$&` etc.); use a replacer function `() => work`. Also `.slice(0, 200)` can split a surrogate pair (astral chars), leaving a lone surrogate; zod counts UTF-16 units so it validates but the stored title may end corrupt.

**P3-2** (`e2e/web-chat-save-login-wall.spec.ts:149`) — `waitForTimeout(500)` before the negative assertion is the timing-dependent pattern the repo bans; wait for a deterministic signal (callback done state) instead.

**P3-3** (AC8 partial) — the P5-timing e2e flow omits the spec's "checkbox recompute" step because Task 1's tray lives in #461. Extend the flow after #461 merges.

**P3-4** — PR base is 2 commits behind `feat/frontend-rebuild` (#448, #449); no file overlap with this diff, but rebase before merge per harness rule.

### Machine summary
```json
{
  "verdict": "approve",
  "findings": [
    {"priority":"P2","category":"clean-code","file":"apps/web/src/components/auth/useAuthCallback.ts","line":11,"issue":"replay result discarded; failed create-on-login has no user-visible surface","smell":"swallowed-result","fix":"surface replay failure state on the callback page","test_needed":"unit"},
    {"priority":"P2","category":"tdd","file":"apps/web/tests/unit/chat/deferred-save.test.ts","line":1,"issue":"unrelated-login-with-live-intent semantics untested (D8 login inside TTL after save tap completes the save)","smell":"missing-test","fix":"pin intended behavior with an explicit unit test","test_needed":"unit"},
    {"priority":"P2","category":"tdd","file":"apps/agent/agent/tests/integration/test_anonymous_flow_uninterrupted.py","line":126,"issue":"test asserts headers of its own request; vacuous","smell":"weak-assertion","fix":"drop or assert on app-side auth dependency invocation","test_needed":"none"},
    {"priority":"P2","category":"security","file":".github/workflows (Security jobs)","line":0,"issue":"Semgrep/osv/gitleaks red, none caused by this diff (JsonLd.tsx pre-existing, lockfile CVEs, gitleaks infra error)","smell":"pre-existing-red","fix":"coordinator: confirm base-red and track separately","test_needed":"none"},
    {"priority":"P3","category":"clean-code","file":"apps/web/src/features/chat/route-copy.ts","line":16,"issue":"replace() dollar-pattern injection; slice(0,200) can split surrogate pair","smell":"string-handling","fix":"replacer function + grapheme-safe truncation","test_needed":"unit"},
    {"priority":"P3","category":"framework","file":"e2e/web-chat-save-login-wall.spec.ts","line":149,"issue":"waitForTimeout(500) timing-dependent negative","smell":"flaky-pattern","fix":"await deterministic callback-done signal","test_needed":"browser"},
    {"priority":"P3","category":"tdd","file":"e2e/web-chat-save-login-wall.spec.ts","line":58,"issue":"AC8 flow lacks checkbox-recompute step (blocked on #461)","smell":"missing-test","fix":"extend e2e after #461 merges","test_needed":"browser"},
    {"priority":"P3","category":"framework","file":"(branch)","line":0,"issue":"base 2 commits stale vs feat/frontend-rebuild","smell":"stale-base","fix":"rebase before merge","test_needed":"none"}
  ],
  "quality_ratchet": {"ac_total": 12, "ac_with_test": 12, "missing": []}
}
```

lifeodyssey (commented):
## Reviewer (Fable seat) — PR #462 final-verification round, HEAD a2a5426f

**Verdict: request_changes** — one new P1 introduced by the round-2 fix itself. Everything else in the delta is verified fixed and well above bar; the P1 is a small, surgical change.

### Delta verification (independent, fresh detached worktree)
- Gates at a2a5426f: typecheck green, oxlint green, **1013 unit + 13 integration passed**, coverage 98.69/95.18/99.48/99.72 vs floors 95/94/95/95.
- Branch properly rebased onto current `feat/frontend-rebuild` (round-1 P3-4 closed).
- Delta mutations (2/2 killed): (A) restore-on-failure with a *fresh* timestamp instead of the original (TTL extension) → `restores a failed claim with its original timestamp` fails; (B) flatten `save-failed` into `done` in `stateFor` → 4 carve-out/callback tests fail.

**P2-1 — fixed.** `replayDeferredSave` returns typed `none|saved|failed`; `useAuthCallback` returns `{state, retrySave, dismissSave}` with a distinct `save-failed` state; `AuthCallback` renders the failure with retry + explicit continue and does NOT call `onDone` silently; rejection maps to `error`; 8s `REPLAY_TIMEOUT_MS` guarded with a fake-timer test (no timing assertions). API change reviewed — clean.

**P2-2 — pinned, both directions**: `an unrelated login inside the TTL still completes the user's own save` + `consumes the intent, so a second unrelated login replays nothing`, plus the claim-before-send two-tab dedupe. Good.

**P2-3 — fixed properly.** The vacuous self-headers test is gone, replaced by three genuine app-side assertions: OpenAPI declares no security requirement on `/v1/runtime`; the #441 guard rejects a credentialed anonymous stamp with 401; identityless request gets no 401/403 and no `WWW-Authenticate`. Each is falsifiable by a production change.

**P3s — all landed**: replacer functions (no dollar-pattern injection), `Intl.Segmenter` grapheme truncation that trims the *work title* and preserves the stop count, e2e negative now waits on `waitForURL` (deterministic) instead of `waitForTimeout(500)`.

**Bonus hardening verified**: double-tap dedupe + `aria-busy`, failure classification (401 → re-wall with fresh intent; permanent 4xx → no retry; 5xx → retry), `pruneDeferredSave` mount sweep, malformed-entry erasure.

### P1 (new, blocking) — dismissing the wall after the magic link was SENT destroys the pending save

`useSaveGate.ts::useCloseLogin` now runs `clearDeferredSave()` unconditionally, and `LoginModal`'s `onClose` fires on **the close button, Esc, and scrim click**. `useMagicLinkForm` has a `sent` state ("check your email") — the natural next action after reaching it is to close the dialog and go to the email client. On that path:

tap 保存する → enter email → send link → **close dialog** → click magic link → login succeeds → `replayDeferredSave` returns `none` → callback shows done → redirect home → **no save, no error**.

That silently breaks AC3's happy-path promise ("after a magic-link login initiated by the save tap, `users.saveRoute` fires once with the deferred `point_ids`") on a common, even accidental (scrim/Esc) interaction. The unit pin `drops the intent on dismissal, so a later unrelated login saves nothing` and the e2e lifecycle test both avoid this case — the e2e keeps the modal open while the callback tab logs in.

The round-1 concern this clear-on-dismiss was answering (D8/D11 login surprise-completing an old save) is *already* neutralized by the newly pinned P2-2 consume-once semantics plus TTL; unconditional clearing overshoots.

**Fix (surgical):** make dismissal-clear state-aware — clear the intent only when the form has **not** reached `sent` (lift `sent` from `useMagicLinkForm` via `LoginModal` to the gate — e.g. an `onSent` callback that marks the intent committed, or pass form status into `onClose`). Pin with two unit tests: close-before-send clears; close-after-send keeps. Optionally extend the e2e lifecycle test to close the dialog after send before opening the callback tab — that would have caught this.

(If the owner explicitly rules "dismissal after send = cancellation", I will withdraw the P1 — but then the AC3 e2e must cover the close-after-send path showing *intentional* no-save, and the LoginModal sent-state copy must say the save was cancelled, because today it says "check your email" while the save is already dead.)

### Machine summary
```json
{
  "verdict": "request_changes",
  "findings": [
    {"priority":"P1","category":"framework","file":"apps/web/src/features/chat/save/useSaveGate.ts","line":63,"issue":"useCloseLogin clears the deferred intent unconditionally; LoginModal onClose fires on close/Esc/scrim, so closing after the magic link was sent silently kills the pending save (AC3 happy path)","smell":"conditional-side-effect","fix":"state-aware dismissal: clear only when useMagicLinkForm never reached sent; pin close-before-send clears / close-after-send keeps; extend e2e lifecycle to close-after-send","test_needed":"unit"}
  ],
  "quality_ratchet": {"ac_total": 12, "ac_with_test": 12, "missing": []},
  "resolved_from_round_1": ["P2-1","P2-2","P2-3","P3-1","P3-2","P3-4"],
  "still_open_nonblocking": ["P3-3 (AC8 recompute step, blocked on #461)","P2-4 (Security CI red, pre-existing/base)"]
}
```

lifeodyssey (commented):
## Reviewer (Fable seat) — PR #462 round 3, HEAD 0f62c247

**Verdict: APPROVE.** The round-2 P1 is fixed exactly along the adjudicated line, correctly implemented, and pinned from every angle I asked for.

### Delta verification (fresh detached worktree at 0f62c247)
- typecheck green, oxlint green, **1020 unit passed**, coverage 98.70/95.11/99.48/99.72 vs floors 95/94/95/95.
- Integration: first run had 5 failures in `build-output.test.ts` — **environmental flake, not the PR**: root cause was a cold `pnpm run build` exceeding the 180s spawn limit, with the 4 wrangler dry-runs cascading off the missing `.output/server/index.mjs`. The same suite passed in this identical environment at round 2, the PR delta cannot influence the vite/wrangler build, and a rerun passed **13/13**. No action needed.
- Mutations on the new guard (2/2 killed): (X) make dismissal-clear unconditional again (the exact round-2 P1) → 2 close-after-send tests fail; (Y) remove the re-arm in `openLogin` → the re-arm test fails.

### P1 fix review
- `LoginForm.onSent` fires only on a genuine `sent` status (the failed-send negative is pinned), passes through `LoginModal`, and `useSaveGate` records it in a **ref** — the right call: the flag never drives rendering, and the source comment documents the batching/stale-closure hazard the `useState` first version was caught by. `useWall` keeps the whole invariant in one place.
- Dismissal semantics now: clear only before a link goes out; close-button/Esc (and scrim, via the same `onClose`) all agree; each fresh trip through the wall re-arms. The surprise-save risk that motivated round-2's unconditional clear remains covered by consume-once + TTL.
- Tests: the four-quadrant unit matrix (close-button/Esc x before/after send), re-arm, onSent-once, onSent-not-on-failure; e2e extended to 6 scenarios including the mainline **send → close → new-tab callback → replay fires** and the cancellation twin. The mainline e2e is precisely the test whose absence let round 2's regression through.
- Product semantics implemented per my adjudication (close-after-send is not a cancellation), with the reverse left as a one-line owner switch — recorded, fine.
- One residual noted, not blocking (P3, no action): after send+dismiss, re-tapping 保存する and then cancelling clears the fresh intent, so the *earlier* emailed link, if clicked later, replays nothing. Direct consequence of the chosen re-arm + consume-once semantics; acceptable.

Non-blocking carryovers unchanged: AC8 recompute e2e step waits on #461; Security CI red is pre-existing/base. Tester still owns executing the e2e pack against the running app.

With the Opus seat's approve, #462 has its dual approve — clear to rebase and enter the merge queue.

### Machine summary
```json
{
  "verdict": "approve",
  "findings": [
    {"priority":"P3","category":"framework","file":"apps/web/src/features/chat/save/useSaveGate.ts","line":54,"issue":"send+dismiss, then re-open and cancel clears the fresh intent, so the earlier emailed link replays nothing — consequence of chosen re-arm + consume-once semantics","smell":"semantics-edge","fix":"none required; note for docs","test_needed":"none"},
    {"priority":"P3","category":"tdd","file":"apps/web/tests/integration/build-output.test.ts","line":53,"issue":"cold-build 180s spawn limit flaked one local run (passed on rerun and in prior round)","smell":"env-flake","fix":"none for this PR; consider a larger spawn allowance if it recurs in CI","test_needed":"none"}
  ],
  "quality_ratchet": {"ac_total": 12, "ac_with_test": 12, "missing": []},
  "mutations_this_round": "2/2 killed (unconditional-clear, no-re-arm)",
  "resolved": ["round-2 P1 (dismissal-after-send intent loss)"]
}
```

Issue #447, opened 2026-07-28
arm Turnstile for anonymous access, then enable prod ANON_ACCESS_ENABLED


lifeodyssey:
Owner decision 2026-07-28. Prod `ANON_ACCESS_ENABLED` stays `false` until Turnstile is armed.

Context: #436 shipped the Turnstile edge gate **dormant** (built + tested, never called). Until it's live, anonymous identity is cookie-based — dropping the `aid` cookie mints a fresh identity and resets the per-identity rate limiter, so the only real guard is the daily budget breaker: an attacker can burn the daily LLM budget and the breaker then walls out legitimate visitors (paid-for daily DoS).

Scope:
1. Wire the dormant Turnstile verification into the anonymous path (edge), per the #436 implementation.
2. Frontend: render the Turnstile widget on the anonymous chat entry (site key already in `apps/web/.env` `VITE_TURNSTILE_SITE_KEY`).
3. E2E/API test: anonymous request without a valid Turnstile token is challenged; with token passes.
4. After green + deploy: flip prod `ANON_ACCESS_ENABLED` to `true` (owner-approved as part of this card).

Ordering: after #441 (JWT fallthrough fix) merges. Related: #438 (anonymous access), #436 (dormant gate), #441.

lifeodyssey:
Delivered by PR #463 (merged): guardTurnstile armed inside handleAnonymousV1 (after resolveAnonymous, before rate limiter/container), pass window keyed (identity, token) closing the cookie-drop replay, interaction-only widget with data-error/expired callbacks, siteverify 5s timeout fail-open (uncached, structured log), missing-secret structured diagnostics, retry driven by token callback, readiness wait unified in session-headers.ts covering chat/photo/confirm. Dual-approved with probe re-verification (25-identity replay, 8 wait boundaries). Follow-up: #471 (photo challenge recovery). NOTE: staging smoke with real keys is a hard prerequisite before flipping ANON_ACCESS_ENABLED anywhere; TURNSTILE_SECRET must be set wherever the flag is true.

PR #445, opened 2026-07-26
feat: S1.3 clarification flows, location prompts, photo search phase 1 (#260)

lifeodyssey:
Closes #260

S1.3 — Clarification flows, location prompts, and photo search (phase 1). All 12 ACs implemented with covering tests in this diff.

## AC → covering test

| AC | Test |
|---|---|
| 1. C2 clarify: 2–4 candidate buttons + escape hatch; selection becomes user bubble, rest fade | `apps/web/tests/unit/chat/clarify-card.test.tsx` (caps at four + escape hatch / sends chosen + fades rest / escape hatch fades without sending) |
| 2. C2t: chips when departure point AND time both missing; skipped when stated | `apps/web/tests/unit/chat/departure-prompt.test.tsx` (trigger/skip detector, hold-then-chip flows, 4-chip render, 現在地→C4 swap) |
| 3. C4 denial → manual text entry, not a dead end | `apps/web/tests/unit/chat/location-prompt.test.tsx` (denial → manual entry → onManual) |
| 4. Photo happy path: `fixtures/vision/yourname_screenshot.jpg` → 君の名は。/160209 → pilgrimage map | `apps/agent/agent/tests/integration/test_photo_search_contract.py::test_yourname_fixture_resolves_to_its_pilgrimage_map` (+ client render path: `photo-search-upload.test.tsx` "renders the recognized title's search card through the shared path") |
| 5. `unknown_landscape.jpg` → C2 clarify + manual-entry chip (existing clarify branch) | `test_photo_search_contract.py::test_landscape_fixture_degrades_to_the_clarify_branch` + `photo-search-upload.test.tsx` degradation tests |
| 6. Layer 2: vision miss + granted location → one `search_nearby` coarse call, merged with vision candidates | `test_photo_search_contract.py::test_layer_two_merges_nearby_source_with_vision_candidates` + `test_photo_search_pipeline.py::test_layer_two_merges_nearby_works_with_vision_candidates` (asserts merged order + radius 2000) |
| 7. Unsupported format / failed upload → on-brand error, no stuck spinner | `photo-search-upload.test.tsx` errors block (no request on bad type; failure copy + retry; spinner gone; invalid envelope → failure copy) + backend 415/422 in `test_photo_search_route.py` |
[...]

PR #463, opened 2026-07-28
feat: arm the Turnstile gate for anonymous chat (#447)

lifeodyssey:
Arms the dormant S1.9 Turnstile gate (#436) on the anonymous `/v1/chat` branch and gives the web app the widget that feeds it. Closes the "paid-for daily DoS" hole in #447: until now, dropping the `aid` cookie minted a fresh anonymous identity and reset the per-identity limiter, leaving the daily dollar breaker as the only guard.

`ANON_ACCESS_ENABLED` is untouched everywhere — flipping prod is the owner's separate call after deploy validation.

> **Review rounds 2 and 3 are folded in.** Round 2 (Fable + Opus) found five holes — two of them (the token-keyed pass window, the same-tick retry) meant the first cut did not actually close the attack it was written for. Round 3 covered the merged state with #445 (photo search) plus two polish items. The mutation table marks them all.

## 1. Edge wiring (`worker/app.ts`, `worker/turnstile.ts`)

`guardTurnstile` runs inside `handleAnonymousV1`, positioned deliberately:

- **after `resolveAnonymous`** — a challenge is only raised for a caller we would otherwise serve anonymously, so anonymous-access-off still answers 401, never 403;
- **before the limiter and the container** — an unsolved turn costs neither a bucket slot nor an LLM call.

**Pass window keyed by (identity, token)** *(review P1-1, Fable)*. Keyed by token alone, one solved challenge covered **any** identity for its whole window: drop the cookie → fresh identity, fresh rate-limit bucket, replay the same token. The attack merely got a price tag of one challenge per five minutes. Keyed by identity, a cookie-dropper misses the cache, siteverify sees the replay and answers `timeout-or-duplicate` → 403.

**siteverify outages no longer 500 the chat** *(review P1-3, Opus)*. `verifySiteverify` had no try/catch and no timeout: a rejected fetch or a 502 HTML body escaped as an unhandled rejection and every anonymous turn became a bare 500. It now uses `AbortSignal.timeout(5s)` and, on any infrastructure failure, **fails open** with `console.error({event:"edge_turnstile_siteverify_unavailable"})` on every occurrence, and that verdict is **never cached** so verification resumes the instant siteverify does.
  > **Tradeoff, flagged for the owner:** fail-open on a security gate means a siteverify outage reopens the anonymous path unverified. It follows the precedent already set by the edge rate limiter (#438/#451) — infrastructure failure must not take chat down — and the daily-budget breaker still caps what an unverified wave can cost. If you'd rather fail closed, it is a one-constant change plus its two tests.

**A missing `TURNSTILE_SECRET` is now visible** *(review P2-1 / P1-4)*. It previously reached Cloudflare as the literal string `"undefined"` and 403'd everyone with zero logging — indistinguishable from a bot wave. Now an explicit edge-internal record; callers still get a plain 403 disclosing nothing.

## 2. Widget (`apps/web`)

The token store and transport header shipped in #436; nothing rendered a widget. `TurnstileGate` is now mounted from `ChatPage` via `useTurnstileChallenge`:

- **Placement** — the dock's hint slot, directly after the composer, matching `docs/design/2026-07-06-design-sync/Chat 初始状态.html` (`.dock > .composer + .hint`). Never in the thread.
- **Mode: `interaction-only`** (the least intrusive of the three). `always` parks a permanent challenge box under the composer; `execute` needs a `turnstile.execute()` handshake before a token exists. `interaction-only` solves silently and renders **nothing** unless Cloudflare decides a human interaction is genuinely required. `data-size="flexible"`; `data-theme` follows the app's own day/night choice, not the OS.
- **Anonymous-only** — a signed-in visitor never loads api.js.
- **Retry is driven by the widget, not by the click** *(review P1-1, Opus)*. The old handler cleared the token, called the async `turnstile.reset()` and resent in the same tick — the resend was reliably tokenless and ate the identical 403, and a token that landed first was thrown away. Now: clear the spent token → subscribe → reset → resend **only** when the replacement callback fires. No token, no resend; the challenge just stays on screen.
- **`?q=` auto-send waits for the first token** *(same review item)*. A visitor arriving from the hero would otherwise race the widget's first solve and land on an unrecoverable 403 on the very first turn.
- **403 `turnstile_required` is not D8** — an anonymous visitor never had a session to expire — **but only when a widget is on the page** *(review P1-3, Fable)*. A build with no site key, or an edge with no secret, 403s every turn; suppressing the strip there left the chat silently dead. Now the generic D4 retry renders instead.
- **Styling** — semantic tokens only, pinned by CSS tests; no hardcoded palette.
- **Dev/E2E** — an unconfigured **production** build renders no widget (no crash); an unconfigured **dev** build falls back to Cloudflare's always-passing test key `1x00000000000000000000AA`. A wrong-length value still throws — pasting the 35-char SECRET must never reach a render.

## 2b. Photo search — the merged-state blocker (round 3)

#445 put `/v1/photo-search` and `/v1/photo-search/confirm` on `ANON_V1`, so after the merge the gate challenges photo uploads too — and photo has no challenge UI, so an unsolved upload would have thrown `photo_search_failed`: a generic failure with no challenge and no retry.

- **The wait moved into the shared header path.** `session-headers.ts` (which #445 extracted for exactly this reason — chat and photo hit the edge with identical identity semantics) now waits for the widget's token on anonymous requests before resolving. One mechanism covers chat, photo upload and the confirm ping. It is bounded by `TURNSTILE_WAIT_MS`, it only engages when this build actually renders a widget, and it never drops a held token when it does not.
- **A challenged upload reads as a challenge.** `photo-search.ts` inspects the rejection code and `PhotoSearchUpload` renders `dict.turnstile.failed` (localized, already shipped) instead of the photo-failure copy; a genuine backend failure still reads as one.
- **Worker side**: both photo routes are covered in `turnstileArm.test.ts` (challenged unsolved, forwarded solved, confirm ping challenged).
- **Still open, tracked in #471**: photo has no challenge *recovery* — its retry resets the uploader, so the visitor re-picks the file. Chat's re-arm/resend handshake should be reused there. Low urgency: with the readiness wait in place, a challenged upload now only happens when a held token is spent or expired.

## 2c. Two polish items (round 3)

- **`data-error-callback` + `data-expired-callback`** are now wired: a dead loader or an aged-out token clears the store, so no stale token is sent to be rejected. Previously a failed api.js load left the page silently stuck.
- **`remoteip` was already sent** — `guardTurnstile` reads `CF-Connecting-IP` and `verifySiteverify` puts it in the siteverify body; `turnstile.test.ts` pins both the populated and the header-absent case. No change needed.
- **A leak found while fixing the above:** each parked waiter held a pending 15s timer, and abandoned ones starved the whole vitest pool (unrelated suites timing out at 5s+). `clearTurnstileToken()` now abandons waiters, and the hermetic setup clears between tests. Two consecutive clean full runs since.

## 3. Tests

| Layer | File | Covers |
|---|---|---|
| worker unit | `worker/turnstile.test.ts` (+9) | identity-scoped window · `timeout-or-duplicate` replay · missing secret logged + not sent · outage fails open, logs, is not cached, and carries an abort signal |
| worker integration | `worker/turnstileReplay.test.ts` (4, new) | the **real gate** against Cloudflare's single-use semantics: solved turn forwards · same visitor rides the window without re-verifying · **cookie-drop replay rejected** · stays rejected on repeat |
| worker integration | `worker/turnstileArm.test.ts` (10, new) | composition: gate receives token/IP/secret · no token → 403, container untouched · retryable envelope · no cookie minted on a challenge · challenge does not spend the burst budget · authed never challenged · anon-off stays 401 · non-allowlisted and public `/v1` never challenged |
| web unit | `turnstile-armed-flow.test.tsx` (5, new) | against a handler that **rejects tokenless turns**: retry resends only with the replacement token · no tokenless request while solving · a token solved before the click is not discarded · `?q=` waits for the first token · ungated when no challenge is in play |
| web unit | `turnstile-mount.test.tsx` (10, new) | mounts for anonymous · sits after `.chat-input` · never for signed-in · nothing for an unconfigured prod build · **misconfigured build still shows the generic failure** · 403 shows the check retry not the login banner · retry resends · retry re-arms the widget · challenge clears on success |
| web unit | `turnstile-gate.test.tsx` (+7), `chat-design-css.test.ts` (+3) | appearance/size/theme attributes · optional + dev-fallback key resolution · `resetTurnstileWidget` · dock tokens, zero vertical cost until rendered, 3D retry affordance |
| web unit | `turnstile-photo-search.test.tsx` (8, new) | the header path does not resolve until a token exists (deterministic, no clock) and resolves at once when no widget is configured · upload holds, then carries the token · challenged upload shows the redo-the-check copy, a real failure still shows the photo copy · retry affordance survives |
| web unit | `turnstile-token-store.test.ts` (+5), `turnstile-gate.test.tsx` (+4) | parked waiter resolves on solve / on clear / on timeout · error + expired callbacks wired and dropping the token |
| worker integration | `turnstileArm.test.ts` (+3) | photo upload challenged · solved upload forwarded · confirm ping challenged |
| E2E | `e2e/web-chat-anonymous.spec.ts` (+2) | widget present in the dock with a 24-char key and `interaction-only` · a challenged turn shows the retry, never the login prompt |

`tests/setup/turnstile-hermetic.ts` neutralizes Turnstile for every suite that is not about it (vitest runs with `DEV === true`, which would otherwise mount a widget under every chat test and stall `?q=`).

## 4. Mutation evidence

| Mutation | Result |
|---|---|
| remove the `guardTurnstile` call | 4 worker tests fail |
| guard **before** `resolveAnonymous` | anon-off-stays-401 + entry 401 fail |
| guard **after** the rate limiter | burst-budget test fails |
| **window keyed by token alone** (the shipped P1-1 bug) | 3 replay tests fail |
| **no try/catch around siteverify** (the shipped P1-3 bug) | 4 outage tests fail |
| **missing secret not detected** | 2 tests fail |
| **retry resends in the same tick** (the shipped P1-1 bug) | tokenless-resend test fails |
| **auto-send not gated on readiness** | `?q=` test fails |
| **unconditional D8 suppression** (the shipped P1-3 bug) | 2 misconfiguration tests fail |
| **photo path does not wait for a token** (the merged-state gap) | header-layer test fails |
| **challenged upload reported as a photo failure** | 2 photo tests fail |
| error/expired callbacks unwired · cleared token does not abandon waiters | respective tests fail |
| unmount `<ChallengeGate>` / signed-in widget / `appearance: always` / retry without reset | respective tests fail |

**Three** tests were rewritten after **surviving** a mutation: the burst-budget one (vacuous — unsolved turns mint no cookie, so strangers never share a bucket), the widget re-arm, and the photo readiness one. The last is worth calling out: an upload spends several async turns on EXIF-stripping and base64 before its headers are read, so a token solved during that window is picked up even by a transport that never waits — the component-level test could not tell the two apart. It is now pinned deterministically at the header layer (microtask drain, no clock).

## 5. Gates

*(re-run after rebasing onto #445/#461 — the `ComposerDock` refactor)*

- `pnpm run test:worker` — **176 pass / 0 fail**
- `apps/web`: `typecheck` clean · `lint:oxlint --deny-warnings` clean (no suppressions) · `pnpm test` **166 files / 1072 tests**, coverage **98.68 stmts / 95.12 branches / 99.08 funcs / 99.56 lines** (floors 95/94/95/95). Run twice back-to-back clean after the timer-leak fix.
- E2E `web-chat-anonymous.spec.ts` — 5/5 against a live `apps/web` dev server

## 6. Honest scope of the E2E, and the staging prerequisite

Both new browser specs **stub `/v1/chat` and block `challenges.cloudflare.com`** — they pin the page's contract (widget present in the dock, rejection handled), *not* a live edge. The armed path end-to-end is covered hermetically at the worker level (`turnstileReplay.test.ts` drives the real gate against real single-use semantics with only the network stubbed), but **no test in this PR exercises a real Cloudflare challenge against a real Worker.**

**Prerequisite before `ANON_ACCESS_ENABLED` is flipped anywhere:** a manual smoke on staging with a real site key + secret — one anonymous turn succeeds, and a turn with a stripped `cf-turnstile-response` header is 403'd.

## 7. Operator + merge notes

1. **`TURNSTILE_SECRET` is now required wherever `ANON_ACCESS_ENABLED="true"`** (documented in `wrangler.toml` and `apps/web/.env.example`). **Staging runs with anon on** — its secret must be in place before this lands there, or every anonymous turn 403s. `VITE_TURNSTILE_SITE_KEY` is likewise required for production web builds.
2. **#445 interaction — handled**, see §2b. The remaining recovery gap is tracked in **#471**.
3. **Rebases absorbed:** #451, #458, #273 T1, and **#445/#461** (the challenge slot assembles into the new `ComposerDock`/`Composer` split). Base tests that send an anonymous `/v1` request and expect a 200 (`byok.test.ts`, `authFallthrough.test.ts`, `photoSearch.test.ts`) were given the passing-gate injection the other non-Turnstile suites use — their subject is limiter scoping, credential verdicts and header hygiene, not the challenge.
4. Manual sends **are** now gated, via the shared header wait added in round 3 — the earlier caveat no longer applies.

Refs #447

### DIFF
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -11,4 +11,8 @@ VITE_NEON_AUTH_BASE_URL=
 # Cloudflare Turnstile SITE key — public by design (24 characters; it ships in
 # the client bundle). The 35-character SECRET is a Worker secret binding
 # (TURNSTILE_SECRET) and must never appear here or anywhere client-side.
+# Required for PRODUCTION builds since issue #447 armed the edge gate: without
+# it the anonymous chat entry renders no widget and every anonymous turn is
+# answered 403 turnstile_required. A dev build falls back to Cloudflare's
+# always-passing test key (1x00000000000000000000AA) on its own.
 VITE_TURNSTILE_SITE_KEY=
--- a/apps/web/src/components/TurnstileGate.tsx
+++ b/apps/web/src/components/TurnstileGate.tsx
@@ -1,6 +1,6 @@
 import { useEffect } from "react";
 import type { ChatDict } from "../features/chat/i18n";
-import { rememberTurnstileToken } from "../lib/turnstile/tokenStore";
+import { clearTurnstileToken, rememberTurnstileToken } from "../lib/turnstile/tokenStore";
 
 /** Cloudflare's widget loader. `async defer` per the official embed. */
 export const TURNSTILE_SCRIPT_SRC = "https://challenges.cloudflare.com/turnstile/v0/api.js";
@@ -11,13 +11,51 @@ export const TURNSTILE_ACTION = "turnstile-spin-v2";
 /** Global the widget invokes with the solved token (`data-callback`). */
 export const TURNSTILE_CALLBACK = "onAnimichiTurnstile";
 
+/**
+ * Globals for the two ways a token stops being usable: the challenge failed or
+ * the loader never came up (`data-error-callback`), and the solved token aged
+ * out (`data-expired-callback`). Both clear the store, so the next turn waits
+ * for a fresh solve instead of sending a token the edge will reject.
+ */
+export const TURNSTILE_ERROR_CALLBACK = "onAnimichiTurnstileError";
+export const TURNSTILE_EXPIRED_CALLBACK = "onAnimichiTurnstileExpired";
+
+/**
+ * The least intrusive of Turnstile's three appearances.
+ *
+ * `always` parks a permanent challenge box under the composer; `execute` needs
+ * an explicit `turnstile.execute()` handshake before a token exists, which
+ * would put a round trip in front of the very first message. `interaction-only`
+ * solves silently and renders NOTHING unless Cloudflare actually decides a
+ * human interaction is required — so the dock keeps the design's rhythm
+ * (composer + one quiet hint line) on the overwhelming majority of turns, and
+ * the challenge appears only in the rare case it is genuinely needed.
+ */
+export const TURNSTILE_APPEARANCE = "interaction-only";
+
+/** Fills the composer's width instead of a fixed 300px box when it does show. */
+export const TURNSTILE_SIZE = "flexible";
+
 /** A Turnstile site key is 24 characters; a secret is 35. */
 const SITE_KEY_LENGTH = 24;
 
 declare global {
   interface Window {
     onAnimichiTurnstile?: (token: string) => void;
+    onAnimichiTurnstileError?: () => void;
+    onAnimichiTurnstileExpired?: () => void;
+    /** Injected by api.js; only `reset` is used (re-arm after a rejection). */
+    turnstile?: { reset: (widget?: string) => void };
   }
+}
+
+/**
+ * Re-arm the widget after the edge rejected a token (403 `turnstile_required`).
+ * A Turnstile token is single-use at siteverify, so a retry that reuses the
+ * held token would be rejected again as `timeout-or-duplicate`.
+ */
+export function resetTurnstileWidget(): void {
+  window.turnstile?.reset();
 }
 
 type EnvRecord = Readonly<Record<string, string | undefined>>;
@@ -46,6 +84,35 @@ export function currentTurnstileSiteKey(): string {
   return resolveTurnstileSiteKey(import.meta.env);
 }
 
+/**
+ * Cloudflare's published always-passing TEST site key. Used only as the dev
+ * fallback below, so `pnpm dev` and the E2E run exercise the real armed path
+ * (issue #447) without anyone provisioning a widget.
+ */
+export const TURNSTILE_TEST_SITE_KEY = "1x00000000000000000000AA";
+
+/**
+ * The site key to render with, or `undefined` when this build has none.
+ *
+ * An unconfigured PRODUCTION build renders no widget rather than crashing the
+ * chat page; an unconfigured DEV build falls back to the test key, because the
+ * edge now challenges every anonymous turn and a local page with no widget
+ * could never answer. A present-but-wrong-length value still throws in both:
+ * pasting the 35-char SECRET must never reach a render.
+ */
+export function configuredTurnstileSiteKey(
+  env: EnvRecord = import.meta.env,
+  dev: boolean = import.meta.env.DEV,
+): string | undefined {
+  if ((env.VITE_TURNSTILE_SITE_KEY ?? "") !== "") return resolveTurnstileSiteKey(env);
+  return dev ? TURNSTILE_TEST_SITE_KEY : undefined;
+}
+
+/** Match the widget to the app's own day/night choice rather than the OS's. */
+function widgetTheme(): "light" | "dark" {
+  return document.documentElement.dataset.theme === "night" ? "dark" : "light";
+}
+
 /** Inject the loader once per document. */
 function useTurnstileScript(): void {
   useEffect(() => {
@@ -66,6 +133,21 @@ function useTurnstileCallback(onToken: (token: string) => void): void {
       window.onAnimichiTurnstile = undefined;
     };
   }, [onToken]);
+}
+
+function bindInvalidation(): () => void {
+  const invalidate = () => { clearTurnstileToken(); };
+  window.onAnimichiTurnstileError = invalidate;
+  window.onAnimichiTurnstileExpired = invalidate;
+  return () => {
+    window.onAnimichiTurnstileError = undefined;
+    window.onAnimichiTurnstileExpired = undefined;
+  };
+}
+
+/** Drop the held token whenever the widget says it is no longer good. */
+function useTurnstileInvalidation(): void {
+  useEffect(bindInvalidation, []);
 }
 
 type RetryProps = Readonly<{ dict: ChatDict; onRetry: () => void }>;
@@ -89,14 +171,19 @@ type Props = Readonly<{
   onToken?: (token: string) => void;
 }>;
 
+/** The attributes api.js reads that never vary between renders. */
+const WIDGET_ATTRIBUTES = {
+  "data-action": TURNSTILE_ACTION,
+  "data-callback": TURNSTILE_CALLBACK,
+  "data-error-callback": TURNSTILE_ERROR_CALLBACK,
+  "data-expired-callback": TURNSTILE_EXPIRED_CALLBACK,
+  "data-appearance": TURNSTILE_APPEARANCE,
+  "data-size": TURNSTILE_SIZE,
+} as const;
+
 function TurnstileWidget({ siteKey }: Readonly<{ siteKey: string }>) {
   return (
-    <div
-      className="cf-turnstile"
-      data-sitekey={siteKey}
-      data-action={TURNSTILE_ACTION}
-      data-callback={TURNSTILE_CALLBACK}
-    />
+    <div className="cf-turnstile" data-sitekey={siteKey} data-theme={widgetTheme()} {...WIDGET_ATTRIBUTES} />
   );
 }
 
@@ -105,9 +192,14 @@ function TurnstileWidget({ siteKey }: Readonly<{ siteKey: string }>) {
  * (`worker/turnstile.ts`) — this only collects the token and hands it to the
  * store the chat transport reads.
  */
-export function TurnstileGate({ dict, siteKey, failed = false, onRetry, onToken }: Props) {
+function useTurnstileWidget(onToken: ((token: string) => void) | undefined): void {
   useTurnstileScript();
   useTurnstileCallback(onToken ?? rememberTurnstileToken);
+  useTurnstileInvalidation();
+}
+
+export function TurnstileGate({ dict, siteKey, failed = false, onRetry, onToken }: Props) {
+  useTurnstileWidget(onToken);
   return (
     <section className="turnstile-gate" aria-label={dict.turnstile.label}>
       <TurnstileWidget siteKey={siteKey} />
--- a/apps/web/src/features/chat/ChatPage.tsx
+++ b/apps/web/src/features/chat/ChatPage.tsx
@@ -1,6 +1,7 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { TurnstileGate } from "../../components/TurnstileGate";
 import { useLocale } from "../../i18n/context";
-import { classifyFailure } from "../../lib/chat/errorClassifier";
+import { TURNSTILE_REQUIRED_CODE, classifyFailure } from "../../lib/chat/errorClassifier";
 import type { ChatErrorState, FailureSignal } from "../../lib/chat/errorClassifier";
 import { ChatActionsProvider } from "./chat-actions";
 import type { ChatActions } from "./chat-actions";
@@ -36,6 +37,8 @@ import { useConversationHistory } from "./use-conversation-history";
 import { useStreamRecovery } from "./use-stream-recovery";
 import { useTurnTimeout } from "./use-turn-timeout";
 import { useTurnTiming } from "./use-turn-timing";
+import { useTurnstileChallenge, useTurnstileReady } from "./use-turnstile-challenge";
+import type { TurnstileChallenge } from "./use-turnstile-challenge";
 import type { ConversationHistory } from "./use-conversation-history";
 
 export interface ChatPageProps {
@@ -49,6 +52,7 @@ type ShellProps = Readonly<{
   history: ConversationHistory;
   failure: TurnFailureView | undefined;
   recompute: RecomputeTurn;
+  challenge: TurnstileChallenge | undefined;
   onRetry: () => void;
   onSend: (text: string) => void;
   departure: DeparturePromptState;
@@ -160,6 +164,12 @@ function ComposerExtras(props: ShellProps) {
   );
 }
 
+/** The dock's hint slot: silent until Turnstile decides a human check is due. */
+function ChallengeGate({ dict, challenge }: Readonly<{ dict: ChatDict; challenge: TurnstileChallenge | undefined }>) {
+  if (challenge === undefined) return null;
+  return <TurnstileGate dict={dict} {...challenge} />;
+}
+
 /** Chips, photo upload, and the E2 tray dock between the stream and composer. */
 function ComposerDock(props: ShellProps) {
   return (
@@ -170,13 +180,23 @@ function ComposerDock(props: ShellProps) {
   );
 }
 
+/** Composer plus the quiet challenge slot beneath it (design sync `.hint`). */
+function Composer(props: ShellProps) {
+  return (
+    <>
+      <ChatInput dict={props.dict} disabled={isInputLocked(props)} onSend={props.onSend} />
+      <ChallengeGate dict={props.dict} challenge={props.challenge} />
+    </>
+  );
+}
+
 function ChatShell(props: ShellProps) {
   return (
     <main className="chat-page">
       <ShellNotices {...props} />
       <ChatBody {...props} />
       <ComposerDock {...props} />
-      <ChatInput dict={props.dict} disabled={isInputLocked(props)} onSend={props.onSend} />
+      <Composer {...props} />
     </main>
   );
 }
@@ -218,18 +238,28 @@ function isActiveTurn(status: ChatSession["status"]): boolean {
   return status === "submitted" || status === "streaming";
 }
 
-function turnFailureState(chat: ChatSession, timedOut: boolean): ChatErrorState | undefined {
+/**
+ * A challenged turn is NOT D8 — an anonymous visitor never had a session to
+ * expire — but the strip is only suppressed when a widget is actually on the
+ * page to offer the recovery. A misconfigured build (no site key, or an edge
+ * with no secret) rejects every turn with nothing to click, so there the
+ * generic failure must still render rather than the chat dying silently
+ * (issue #447 review, P1-3).
+ */
+function turnFailureState(chat: ChatSession, timedOut: boolean, challenged: boolean): ChatErrorState | undefined {
   if (isActiveTurn(chat.status)) return undefined;
   if (timedOut) return "D5";
   if (chat.error === undefined) return undefined;
-  return classifyFailure(turnFailureSignal(chat.lastHttpStatus(), chat.lastErrorCode())) ?? "D4";
+  const code = chat.lastErrorCode();
+  if (challenged && code === TURNSTILE_REQUIRED_CODE) return undefined;
+  return classifyFailure(turnFailureSignal(chat.lastHttpStatus(), code)) ?? "D4";
 }
 
 /** Compose the D4/D5/D8/D11 view: watchdog + classification + P6 recovery. */
-function useTurnFailure(chat: ChatSession, baseUrl: string): TurnFailureView | undefined {
+function useTurnFailure(chat: ChatSession, baseUrl: string, challenged: boolean): TurnFailureView | undefined {
   const timeout = useTurnTimeout(chat.status, () => void chat.stop());
   const recovery = useStreamRecovery(baseUrl, chat, chat.sessionIdOf);
-  const state = turnFailureState(chat, timeout.timedOut);
+  const state = turnFailureState(chat, timeout.timedOut, challenged);
   const onRetry = useCallback(() => { timeout.reset(); recovery.recover(); }, [timeout, recovery]);
   const onExpiredResume = useCallback(() => { timeout.reset(); recovery.recoverExpired(); }, [timeout, recovery]);
   if (state === undefined) return undefined;
@@ -246,10 +276,12 @@ function entryStateOf(search: ChatSearch, health: BackendHealth): ChatEntryState
   });
 }
 
-function useAutoSendFromQuery(search: ChatSearch, health: BackendHealth, send: (text: string) => void) {
+function useAutoSendFromQuery(
+  search: ChatSearch, health: BackendHealth, send: (text: string) => void, ready: boolean,
+) {
   useAutoSend({
     query: search.q,
-    enabled: health.healthy && !search.session,
+    enabled: health.healthy && ready && !search.session,
     send,
     sessionId: search.session,
   });
@@ -277,9 +309,9 @@ function usePhotoContext(locale: ReturnType<typeof useLocale>, chat: ChatSession
 }
 
 /** Tray state: the recompute turn, its masked failure, and the spot store. */
-function useTrayState(chat: ChatSession, baseUrl: string) {
+function useTrayState(chat: ChatSession, baseUrl: string, challenged: boolean) {
   const recompute = useRecomputeTurn(chat);
-  const failure = maskRecomputeFailure(recompute, useTurnFailure(chat, baseUrl));
+  const failure = maskRecomputeFailure(recompute, useTurnFailure(chat, baseUrl, challenged));
   const selection = useSpotSelectionState();
   return { recompute, failure, selection };
 }
@@ -297,16 +329,18 @@ function useChatPage(search: ChatSearch) {
   const { config, health, chat, history } = useChatState(search);
   const { actions, gps } = useOriginTracking(useTurnActions(chat));
   const surfaces = usePageSurfaces(chat, actions, gps);
-  const tray = useTrayState(chat, config.baseUrl);
-  useAutoSendFromQuery(search, health, actions.send);
-  return { config, health, chat, history, actions, ...surfaces, ...tray };
+  // `?q=` must not fire before the widget has a token to send (#447 review).
+  const challenge = useTurnstileChallenge(chat);
+  const tray = useTrayState(chat, config.baseUrl, challenge !== undefined);
+  useAutoSendFromQuery(search, health, actions.send, useTurnstileReady(challenge !== undefined));
+  return { config, health, chat, history, actions, challenge, ...surfaces, ...tray };
 }
 
 type PageState = ReturnType<typeof useChatPage>;
 
 function ChatPageView({ search, page }: Readonly<{ search: ChatSearch; page: PageState }>) {
   return (
-    <ChatShell entry={entryStateOf(search, page.health)} dict={page.dict} chat={page.chat} history={page.history} failure={page.failure} recompute={page.recompute} onRetry={page.health.retry} onSend={page.departure.onSend} departure={page.departure} baseUrl={page.config.baseUrl} photo={page.photo} />
+    <ChatShell entry={entryStateOf(search, page.health)} dict={page.dict} chat={page.chat} history={page.history} failure={page.failure} recompute={page.recompute} challenge={page.challenge} onRetry={page.health.retry} onSend={page.departure.onSend} departure={page.departure} baseUrl={page.config.baseUrl} photo={page.photo} />
   );
 }
 
--- a/apps/web/src/features/chat/components/PhotoSearchUpload.tsx
+++ b/apps/web/src/features/chat/components/PhotoSearchUpload.tsx
@@ -5,6 +5,7 @@ import { ChatActionsProvider, useChatActions } from "../chat-actions";
 import type { ChatActions } from "../chat-actions";
 import type { ChatDict } from "../i18n";
 import {
+  PHOTO_CHALLENGED,
   confirmPhotoSearch,
   isOversizedPhoto,
   isSupportedPhoto,
@@ -23,7 +24,7 @@ import { DataPartCard } from "./DataPartCard";
  * through DataPartCard, sharing the text-search render path; failures show
  * on-brand copy with a retry — never a stuck spinner. */
 
-type UploadError = "unsupported" | "tooLarge" | "failed";
+type UploadError = "unsupported" | "tooLarge" | "failed" | "challenge";
 
 type UploadState =
   | { readonly kind: "idle" }
@@ -42,6 +43,7 @@ function quotaCopy(dict: ChatDict, guidance: PhotoGuidance): string {
 function errorCopy(dict: ChatDict, error: UploadError): string {
   if (error === "unsupported") return dict.photo.unsupported;
   if (error === "tooLarge") return dict.photo.tooLarge;
+  if (error === "challenge") return dict.turnstile.failed;
   return dict.photo.failed;
 }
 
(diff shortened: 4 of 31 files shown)


lifeodyssey (commented):
## Fable independent review (second seat) — PR #463

Independently verified: `pnpm run test:worker` 125/125 · apps/web `tsc` clean · oxlint clean · targeted vitest suites 47/47 · mutation re-checks (3 sampled, all killed: guard-before-resolveAnonymous → 2 fail; guard-after-limiter → burst test fails; D8-fallthrough → login-banner test fails). Placement matrix (anon-off→401, authed never challenged, public/non-allowlisted untouched, no cookie mint / no bucket spend / no container on a challenge) is fully pinned by `worker/turnstileArm.test.ts` and holds under mutation. gitleaks/osv/Semgrep failures are fleet-wide (identical on #462), not this PR.

**Verdict: request_changes** — two P1s on the exact properties this PR exists to deliver.

### P1

**P1-1 (security, `worker/turnstile.ts:101-127` + `worker/app.ts`)** — The pass window is keyed by **token only** (`passed: Map<token, expiry>`), not by identity. Sequence: solve once with cookie A → token T cached for 5 min → drop the `aid` cookie → fresh identity minted, fresh rate bucket → replay T → `isWithinWindow` passes without touching siteverify → container reached. Within the window, the cookie-drop limiter-reset attack this PR closes (#447) still works; the challenge cost amortizes to one solve per 5 min per isolate. Fix is small and UX-preserving: `guardTurnstile` runs after `resolveAnonymous`, so key the window on `token + identity.userId`. A returning visitor keeps their cookie → same key → still never re-challenged per message; a cookie-dropper gets a cache miss → siteverify replies `timeout-or-duplicate` → 403. Add a worker test: same token, second request without the cookie → 403.

**P1-2 (correctness, `apps/web/src/features/chat/use-turnstile-challenge.ts:247-255`)** — Retry is a guaranteed first-click failure against the real edge. `onRetry` runs `clearTurnstileToken(); resetTurnstileWidget(); clearError(); void regenerate()` in one tick; `reset()` re-solves **asynchronously**, and the transport's `headers` (`use-chat-session.ts:121`) evaluates at request time — so the resend carries **no token** → 403 again. Verified empirically with a scratch probe: initial request carried the held token, the retry request's `cf-turnstile-response` header was `null`. Your own tests can't see this because the MSW success handler (`chatStreamHandler`) accepts tokenless requests. Fix: defer the resend until a fresh token lands (trigger `regenerate` from the token callback / await the store with a timeout), and make the resend-test assert the retry request actually carries a token.

**P1-3 (diagnosability, `apps/web/src/features/chat/ChatPage.tssx` `turnFailureState`)** — D8 suppression is unconditional on `code === TURNSTILE_REQUIRED_CODE`, but the challenge UI only exists when a site key is configured. In the documented misconfiguration (prod/staging build without `VITE_TURNSTILE_SITE_KEY`, or edge missing `TURNSTILE_SECRET` → all-403), the client shows **nothing at all**: no widget, no retry, no D-strip — chat silently dead. Suppress D8 only when `challenge !== undefined`; otherwise fall through to the generic failure so the breakage is at least visible.

### P2

- **Missing-secret diagnosability (`worker/turnstile.ts`)** — `Env`/`TurnstileEnv` type `TURNSTILE_SECRET` as required `string`, but an unbound secret is `undefined` at runtime; it flows into `URLSearchParams` as the literal `"undefined"` and every solved token 403s with no log and (deliberately) no echoed error codes. Staging (anon on) hits exactly this until the secret is set. A one-line guard (`if (!env.TURNSTILE_SECRET) console.error(...)` or a distinct internal log) would make the wrangler.toml doc enforceable in practice.
- **E2E claim overstated** — both new specs in `e2e/web-chat-anonymous.spec.ts` abort `challenges.cloudflare.com/**` and stub `/v1/chat`; no test anywhere exercises widget → header → worker gate → siteverify as one path. The layering (worker composition in `turnstileArm.test.ts`, widget contract in Playwright) is legitimate, but the PR body's "E2E exercise the real armed path" only holds for the widget render, not the edge. Worth one dev-stack smoke (test key + test secret) before staging flip.

### P3

- **#445 conflict surface** — #445 in flight rewrites `ChatPage.tsx` (+88) and `use-chat-session.ts`; this PR restructures `ChatShell`/dock in the same region. Whoever lands second rebases carefully around `ShellProps`/`ChatDock`.
- `TURNSTILE_REQUIRED_CODE` as a wire-contract literal: justified and precedented (`ANON_BUDGET_EXHAUSTED_CODE`) — fine.
- Design tokens, 1-10-50, ≤200-line test files, no suppressions: all clean. `widgetTheme()`'s `document` access is render-time but unreachable during SSR (widget mounts only after the client-side auth effect) — OK, worth a comment.

Positives: the placement rationale in `handleAnonymousV1` is exactly right and mutation-hardened; the rewritten burst-budget test (returning visitor's own cookie) is the correct observable; `interaction-only` justification is sound; typed 403 envelope with no error-code echo is the right disclosure posture.

— Fable (independent seat; did not read the other reviewer's findings)

lifeodyssey (commented):
## Independent review (second seat) — PR #463 / issue #447

Reviewed in a detached worktree at `0bb69304`, without reference to any other review. Gates reproduced locally: `test:worker` 125/125, `apps/web` typecheck clean, `lint:oxlint --deny-warnings` clean (no suppressions anywhere in the diff), `pnpm test` coverage 98.93 / 95.14 / 99.66 / 99.88. The three red CI security jobs are pre-existing and unrelated (semgrep = `JsonLd.tsx` dangerouslySetInnerHTML; osv = `docs/superpowers/spikes/map-stack` lockfile; gitleaks is not reporting a leak at all — it dies on an octokit `GET /pulls/463/commits` error, i.e. the secret gate is silently broken repo-wide, worth its own issue).

The placement argument in the PR body is right and I have no quarrel with it: after `resolveAnonymous`, before limiter and container, is the correct position and the tests pin it. My findings are all about what happens on the *edges* of that decision, and I verified every one of them by running code rather than reading it.

**Verdict: request changes.** Two of these (P1-1, P1-2) mean the feature does not do what the PR says it does.

---

### P1-1 — the retry can never succeed, and the landing funnel's first turn is always challenged

`useChallengeRetry` clears the token, calls `turnstile.reset()`, and resends **synchronously**. `reset()` is asynchronous — the fresh token arrives via `data-callback` hundreds of ms later — so the resend always goes out with no token. Worse, because the retry *clears* the store first, a token that has already arrived is thrown away, so the second and every subsequent click is equally tokenless. Against a real armed edge this is a deterministic loop: 403 → retry → 403 → forever.

Probe output (my own test, run against this branch):

```
first send token=first-token; resets=1; resent tokens=[null]
SECOND retry sent tokens=[null]     <- a fresh token was already in the store
```

Same probe on the landing funnel (`/chat?q=...`, the primary entry from the hero):

```
auto-send token=[null] widgetMountedAtSendTime=true
```

The widget is mounted but unsolved when `useAutoSendFromQuery` fires. So once `ANON_ACCESS_ENABLED` flips, **every anonymous visitor arriving from the landing page gets 403 on their very first turn, with a retry button that cannot recover.**

This is an architectural contradiction, not a typo: `interaction-only` was chosen (correctly, and well argued) to avoid putting a round trip in front of the first message — but the edge now hard-requires a token *on* the first message. One of the two has to give.

- Fix: gate anonymous sends on `currentTurnstileToken() !== undefined` (hold the turn / keep the composer in its busy state until the callback lands, rather than firing tokenless), and drive the retry's resend *from* the token callback instead of immediately after `reset()`.
- Also wire `data-error-callback` / `data-timeout-callback`. Today, if `api.js` is blocked (uBlock lists do block `challenges.cloudflare.com`, and your own E2E blocks it), `window.turnstile` is undefined, `resetTurnstileWidget()` is a silent no-op, and the visitor is locked out with a message that implies retrying will help.
- Both new tests here are green for the wrong reason: `chatTurnstileRequiredHandler` is swapped for an unconditional `chatStreamHandler("search")` that never inspects the header, and the re-arm test asserts `reset` was called but not that the resend happened *after* a new token. The test that would have caught this: an MSW handler that 403s unless a **fresh, different** `cf-turnstile-response` is present.

### P1-2 — one solved token still buys unlimited fresh anonymous identities (the hole #447 is supposed to close)

`isWithinWindow` keys the 5-minute pass window on the **raw token string**, not on identity or IP, and a cached hit skips `verifySiteverify` entirely — so `remoteip` is never re-checked either. `guardTurnstile` is called with `identity` already in hand and ignores it.

Probe, 25 requests, one token, 25 different `CF-Connecting-IP`s, no cookie:

```
siteverify calls: 1; container hits: 25; distinct identities: 25
```

That is the attack described in the issue, executed *through* the new gate: drop the `aid` cookie, replay one solved token, get a fresh identity and a fresh limiter bucket every time, for 5 minutes per isolate, with only the daily dollar breaker behind it. Cross-isolate replay does fail (siteverify returns `timeout-or-duplicate`), which limits blast radius, but a client can pin an isolate cheaply.

- Fix: key the pass window on `identity.userId` rather than the token. Then dropping the cookie forces a fresh solve, which is exactly the invariant the issue asks for. Optionally bind the entry to the client IP as well.
- Test: N requests with the same token and no cookie must forward exactly once.

### P1-3 — a siteverify blip 500s every anonymous turn

There is no `try`/`catch` and no timeout around the siteverify call, and `response.json()` is unguarded. Probes:

```
outage (fetch rejects)      status=500 body=Internal Server Error
siteverify returns 502 HTML status=500 body=Internal Server Error   (SyntaxError from .json())
```

The exception escapes `guardTurnstile` → `handleAnonymousV1` → Hono. The `BAD_RESPONSE` constant exists for shape-narrowing but never fires, because the parse throws first. This directly contradicts the sentence two lines above the new code — "the limiter fails open — a guard outage must not take chat down" — and a hanging siteverify will hang the turn to the Worker's wall clock.

- Fix: `AbortSignal.timeout(...)` on the fetch, `try`/`catch` around fetch+parse, and a deliberate, documented fail direction. My recommendation, consistent with #438 and the limiter: **fail open on transport/parse errors** (Cloudflare being unreachable must not take chat down) and keep fail-closed only for an explicit `success: false`. If you'd rather fail closed, then at minimum return the retryable 403 envelope, never a bare 500.

### P1-4 — a missing `TURNSTILE_SECRET` 403s everyone, silently

```
misconfig status=403 turnstile_required, secretSentAs=undefined
```

The literal string `undefined` is posted to Cloudflare as the secret, every request, and the visitor gets the ordinary retryable envelope. The wrangler.toml comment is documentation, not a guard, and staging already runs anon-on — this is the landing risk. Wanted: a loud, distinguishable signal when `ANON_ACCESS_ENABLED === "true"` and the binding is absent — a once-per-isolate `console.error` plus a distinct code (e.g. 503 `turnstile_misconfigured`), so a broken environment is diagnosable from the response instead of looking like a world full of bots.

### P2-1 — an unconfigured production build fails completely silently

`turnFailureState` suppresses the D-strip for `turnstile_required` unconditionally, but the challenge UI only renders when `siteKey !== undefined`. A prod build with no `VITE_TURNSTILE_SITE_KEY` against an armed edge therefore shows **nothing at all**: no widget, no error, no retry — the turn just dies. Suppress D8 only when a challenge is actually being rendered.

### P2-2 — the E2E coverage claim doesn't hold

`blockTurnstileLoader` aborts `api.js`, and the second test fulfils `/v1/chat` in Playwright, so the worker is not in the loop either. Neither E2E nor any unit test exercises a token → siteverify → forward path. The PR body's "local dev and E2E exercise the real armed path" is accurate for `pnpm dev` only. Either soften the claim or add a worker-level test against Cloudflare's published always-passing / always-failing secret pair.

### P2-3 — quality ratchet: the three highest-value edge behaviours have no tests

Ten new worker tests, none covering replay (P1-2), outage (P1-3), or misconfig (P1-4). Each of my probes above is a ready-made test.

### P3

- `remoteip` is posted as `""` when `CF-Connecting-IP` is absent (probe-confirmed). Omit the field rather than sending empty.
- `handleAnonymousV1` now takes four positional params (`env, request, nowMs, gate`) — a context object would read better and stop the next arm-a-guard PR from adding a fifth.
- `widgetTheme()` is read at render only, with no dependency on the theme; toggling day/night leaves the widget on the old theme until remount.
- Minor timing oracle: a window-cached pass returns immediately, an uncached one pays the siteverify RTT, which lets a caller probe which isolate holds their token. Low value on its own; disappears if P1-2 is fixed by keying on identity.
- The 401-vs-403 ordering is correct and I could not find an information leak there. Anon-off → 401 with the gate never consulted, verified.

---

## Merge order (three-way with #445 and #451)

`git merge-tree` against both siblings:

- **#463 × #451** — conflicts in `worker/app.ts` (#451 rewrites the `if (auth.ok)` return line immediately above the dormant-Turnstile comment block that #463 replaces), `package.json` (`test:worker` list), `wrangler.toml`. All mechanical; the two are semantically independent (authed branch vs anonymous branch).
- **#463 × #445** — conflicts in `ChatPage.tsx` (both extend the same single-line `<ChatShell … />` prop list), `chat.css` (both append at EOF), `package.json`.

**Recommended order: #451 → #463 → #445.** #451 is the smallest and worker-only; #463 next while its ChatPage prop is still the newest thing on that line; #445 last, since its conflict is easiest to resolve once the `challenge` prop already exists.

Two combined-state items for whoever rebases #445, neither of which either PR can see alone:

1. #445 extends `ANON_V1` to `/v1/photo-search` and `/v1/photo-search/confirm`. After both land, **photo search is challenged too** — that is defensible, but it is a behaviour change nobody signed off on in either PR.
2. #445's `session-headers.ts` does attach `turnstileHeaders()` for photo search, so the token rides along correctly. But `useTurnstileChallenge` reads only `chat.lastErrorCode()`, so a `turnstile_required` from a **photo-search** call surfaces no retry UI at all — it falls into photo search's generic failure path. Needs a follow-up once both are in.

lifeodyssey (commented):
## Fable re-review (second seat) — PR #463, post-fix round

Independently re-verified at head `5064ef39` (rebased on #451/#458/#461):

- `pnpm run test:worker` **167/167** · apps/web `tsc` clean · `oxlint --deny-warnings` clean · full `pnpm test` **159 files / 1002 tests pass**, coverage 98.91 / 95.38 / 99.58 / 99.88 (floors 95/94/95/95)
- **P1-1 (window replay)** — fixed for real. `windowKey = identity + "\n" + token`, `guardTurnstile` takes `identity.userId` from `resolveAnonymous`. `turnstileReplay.test.ts` drives the **real gate** (only siteverify fetch stubbed, with genuine single-use semantics) and pins the exact attack: cookie-drop replay reaches siteverify, gets `timeout-or-duplicate`, 403, container untouched. Mutation check: reverting the key to token-only → **3 tests fail**. Also confirmed the fail-open verdict is excluded from the cache (`isVerifiedPass`).
- **P1-2 (retry race)** — fixed. `resendWhenSolved` subscribes **before** `reset()` and resends only when the callback delivers a token; `?q=` auto-send gated on `useTurnstileReady`. Confirmed the coordinator's specific ask: `armedChatHandler` now 403s tokenless turns, and the retry tests assert the header sequence directly (`seen === [null, "fresh-token"]`; the still-solving test asserts `seen === [null]` with no tokenless resend on the wire). Mutation check: restoring the same-tick resend → armed-flow suite fails.
- **P1-3 (silent misconfiguration)** — fixed. Suppression is now `challenged && code === TURNSTILE_REQUIRED_CODE`; the classifier maps `turnstile_required` to **D4** (generic retry, never the login banner) when no widget is on the page, and `turnstile-mount.test.tsx` pins the misconfigured-build D4 path.
- **P2-1 (missing secret)** — fixed: `usableSecret` guard + structured `edge_turnstile_secret_missing` log, literal `"undefined"` can no longer reach siteverify; disclosure posture preserved (plain 403 externally). Tested.
- **P2-2** — PR body claim corrected; staging real-key smoke stated as a hard precondition of the `ANON_ACCESS_ENABLED` flip. Accepted.
- The added siteverify **fail-open** on outage (other seat's item): coherent — precedented by the limiter's fail-open, loud structured log, `AbortSignal.timeout(5s)`, never cached, budget breaker still caps the blast radius. Well tested (5 tests incl. the never-cached property).

### Residual P3 (none blocking)

1. `turnstile-armed-flow.test.tsx` test 3 ("does not throw away a token the widget solved **before** the click") doesn't exercise its named scenario — it solves after the click, near-duplicating test 1. Behaviorally, a token solved between the 403 and the click *is* discarded by `clearTurnstileToken()` and costs one extra silent solve — fine, but rename or reshape the test.
2. `useTurnstileReady`: a token remembered between the `useState` initializer and the effect's subscription is missed (`held` stays false until the next solve). Millisecond window; re-checking `currentTurnstileToken()` inside the effect would close it.
3. Security CI (gitleaks/osv/Semgrep) still failing **fleet-wide** (identical on sibling PRs) — infra, not this PR, but flag before merge if branch protection counts them.

**Verdict: approve.** Both P1 mechanisms are now driven by tests that would have caught the original bugs, and both survived my independent mutation probes. Merge-order note acknowledged: #445 lands first (apps/web freeze); this PR's `ChatDock`/`ShellProps` restructuring sits exactly on #445's `ChatPage.tsx` surface, so expect one more rebase + a green re-run of `turnstile-armed-flow` and `turnstile-mount` after it — the approval carries over only if that rebase is conflict-free in `apps/web/src/features/chat/`.

— Fable (independent seat)

lifeodyssey (commented):
## Re-review (second seat) — #463 at `5064ef39`

I re-ran my **original probes verbatim** against the fixed tree, then mutation-tested the fixes rather than trusting the suite. All five items are genuinely fixed. One new blocking item is not about the code.

### 1. Replay — fixed

25 requests / 1 token / 25 IPs / no cookie, with a faithful siteverify stub (a token verifies once, replays answer `timeout-or-duplicate`):

```
before:  siteverify=1   container=25  200s=25
after:   siteverify=25  container=1   200s=1  403s=24
```

Cookie-dropping now misses the `(identity, token)` cache, siteverify sees the replay, the turn dies. And the window still does its real job — a returning visitor keeping their cookie: `siteverify=1 over 5 turns, container=5`. No per-message re-challenge regression.

### 2. Retry — fixed

```
before:  tokens on the wire = ["spent-token", null]   <- and null on every later click
after:   tokens on the wire = ["spent-token", "fresh-1"]
```

Subscribe-then-reset ordering is correct in `resendWhenSolved` (the subscription is taken before `reset()`, so a fast solve cannot land in the gap). With `api.js` blocked, no doomed tokenless resend fires at all and the challenge stays on screen — the infinite 403 loop is gone.

### 3. `?q=` gating — fixed

```
requests before the solve = 0; after the solve = 1, carrying "solved-late"
signed-in / unconfigured build: fires immediately, exactly 1 request (not gated)
```

I chased the one anomaly I saw (a duplicate request in my first pass) to ground: an isolated probe shows exactly **1** request in all four scenarios — held-before-mount, solved-late, signed-in, unconfigured-prod. It was bleed from my own probe file's leftover timers, not a double-send. Cleared.

### 4. Outage — fixed, and the fail-open is genuinely uncached

All three failure shapes, each run twice:

```
fetch rejects:  200 container | 2nd 200 | logs = 2 x edge_turnstile_siteverify_unavailable
502 HTML body:  200 container | 2nd 200 | logs = 2 x ...
hang > timeout: 200 container | 2nd 200 | logs = 2 x ...
```

Two requests produce two log lines, so nothing is cached — `isVerifiedPass` correctly refuses the window slot to a fail-open verdict, and `AbortSignal.timeout` returns promptly instead of hanging the turn. No more bare 500.

### 5. Missing binding — fixed

```
before:  403, and the literal string "undefined" was posted to siteverify   (silent)
after:   403, siteverify NEVER CALLED, logs = edge_turnstile_secret_missing
```

Short-circuits before siteverify, so the placeholder string no longer reaches Cloudflare, and a misconfigured environment is now diagnosable from the logs.

---

## My ruling on fail-open (requested)

**I concur with fail-open — land it as written.** Reasoning: the failure mode is not attacker-reachable. Nothing a caller controls influences whether the Worker can reach siteverify, so an attacker cannot *induce* the open state; they can only wait for one and hope to be online for it. Against that, fail-closed means a Cloudflare-side blip converts into a total anonymous-chat outage, which is a far more likely and more damaging event. The precedent (#438/#451) and the daily-budget cost ceiling both point the same way.

Two conditions I would attach, neither blocking:

1. `edge_turnstile_siteverify_unavailable` is only worth having if something watches it. It needs an alert, otherwise a silent multi-hour open window is indistinguishable from a healthy day.
2. The compounding case deserves a note somewhere: the rate limiter **also** fails open. A guard outage that takes out both leaves the daily budget breaker as the only remaining control — the exact single-control posture #447 set out to fix, just reached by a different road. Worth a follow-up to make sure those two cannot fail open simultaneously and silently.

## Mutation verification — 8/8 killed

I did not trust the suite; I broke the code and watched.

| # | Mutation | Result |
|---|---|---|
| A | window keyed by token alone (revert fix 1) | 3 worker tests fail |
| B | cache the fail-open verdict (`return result.ok`) | 1 worker test fails |
| C | fail **closed** on outage | 3 worker tests fail |
| D | drop the missing-binding guard | 2 worker tests fail |
| E | retry resends immediately (revert fix 2) | `turnstile-armed-flow` fails: `expected [null, null] to deeply equal [null]` |
| F | `?q=` readiness gate removed | 1 test fails |
| G | `useTurnstileReady` always true | 1 test fails |
| H | D8 suppression unconditional again | 2 tests fail |

`armedChatHandler` is the right fix for the false-green root cause — it now 403s whenever the header is absent, so the retry tests can no longer pass by talking to a handler that ignores the token. E is the proof: the old test set was blind to that regression, the new one catches it on the wire.

Gates on the branch tree: worker 167/167 (was 125), web 159 files / 1002 tests, coverage 98.91 / 95.38 / 99.58 / 99.88, typecheck and oxlint clean, zero suppressions in the diff.

---

## Blocking merge precondition (new, not a code defect)

**This branch is behind its base, and #445 has already landed there.**

```
base HEAD:  2726b677  (#445 photo search)
463 parent: dddd5db4  (#461)
merge-base --is-ancestor -> *** BEHIND ***
merge preview vs base -> CONFLICT in ChatPage.tsx, chat.css, package.json
```

So every green number above — mine and the PR's — was measured on a tree that does not contain #445. Two consequences:

1. `ANON_V1` on base is already `["/v1/chat", "/v1/photo-search", "/v1/photo-search/confirm"]`. The moment this merges, **photo search is Turnstile-guarded**, on a surface no test in this PR covers. That is not a follow-up any more; it is live on merge.
2. `useTurnstileReady` gates only the `?q=` auto-send. A photo upload is not gated, so an anonymous upload before the widget solves goes out tokenless, gets the 403, and base's `settleResponse` turns any non-429 failure into `throw new Error("photo_search_failed")`. The visitor gets a generic photo-search failure, no challenge, no retry, no explanation. A second manual attempt would succeed once the widget has solved, so it is degraded rather than broken — but it needs the same readiness gate and a `turnstile_required` branch.

**Rebase onto `2726b677`, re-run both suites, and confirm the ChatPage.tsx conflict resolves without disturbing the hook plumbing this PR restructured.** #451 is now the only sibling left; it will need its own rebase (its `worker/byok.test.ts` add/add against the new base is its problem, not yours).

## Residual, non-blocking

- **P2 — a blocked `api.js` is now a silent stall.** Probe with no widget available and `?q=` set: `requests=0, failureStripShown=false, alerts=0, composer present`. The page looks fine and simply never sends. This is a fair trade against the old unrecoverable 403 loop, but `data-error-callback` / `data-timeout-callback` wired to the existing `turnstile.failed` strip would close it, plus a bounded fallback on the auto-send path so the visitor is told rather than left waiting.
- **P3 — `remoteip` is still sent as an empty value** when `CF-Connecting-IP` is absent. Omit the field rather than sending it empty. Cosmetic in production, where the header is always present.
- The photo-search follow-up is correctly noted in the PR body; per the above it should be an issue filed now, not after.

---

## Verdict

**Approve the fixes — all five verified by re-run probes and 8/8 mutations killed.** Do not merge until the branch is rebased onto `2726b677` and both suites are re-run green there, since the current validation was measured against a tree without #445, and the merge is not clean.

Merge order, updated now that #445 is in: **#463 (after rebase) → #451**.

lifeodyssey (commented):
## Delta acceptance (second seat) — `2ca5eaea`

Three things checked, plus my original replay probe re-run on the new tree.

### Wait mechanism — all eight boundaries hold

Probed `sessionHeaders()` directly, racing it against a short timer to tell "settled" from "parked":

```
A no-widget build:      settled immediately, headers {}
B token already held:   settled immediately, headers {x-session-id, cf-turnstile-response: held}
C no token yet:         PARKED  (does not fire tokenless)
D solve while parked:   released WITH the token  -> {cf-turnstile-response: late-solve}
E clear while parked:   abandoned after ~45ms, not 15000ms -> headers {}
F bounded:              awaitTurnstileToken(80) -> undefined after ~83ms
G signed-in:            settled immediately, no challenge header
H three parked waiters: all three released by one clear, ~41ms
```

That is exactly the contract you described, and E/H are the ones that matter for the flake you fixed — `clearTurnstileToken` really does drop the pending timers rather than leaving them to starve the pool. The placement is right too: putting the wait in `session-headers.ts` means chat, `/v1/photo-search` and `/v1/photo-search/confirm` inherit it from the one path they already share, rather than three copies drifting apart. `configuredTurnstileSiteKey() !== undefined` is the correct guard — it keys the wait to "does this build render a widget", not to a runtime guess.

### `data-expired-callback` — wired

Both globals are declared and both attributes ride on the widget alongside `data-callback`, and both clear the store, so a dead or aged-out token can no longer be attached to the next turn. Worth knowing about one interaction you may not have traced: because clearing now also abandons waiters, a turn that races the ~5-minute expiry is released tokenless instead of waiting the second or so for the widget's own refresh. Bounded and self-correcting (chat shows the retryable challenge, photo shows the new classified error), so I would leave it — just noting it is a deliberate consequence, not an oversight.

### Replay probe — still holds

Re-ran the original 25 requests / 1 token / 25 IPs / no cookie on this tree:

```
siteverify=25  container=1  200s=1  403s=24
```

And I checked the merged allowlist end to end, since photo search is now anonymous:

```
/v1/chat                   noToken=403  withToken=200
/v1/photo-search           noToken=403  withToken=200
/v1/photo-search/confirm   noToken=403  withToken=200
```

### Mutation-2 rewrite — the right shape

`armedChatHandler(name, seen, [SPENT])` rejecting a specific token, driven through the widget's real `data-callback` inside `act`, asserting `seen).toEqual(["spent-token", "fresh-token"])` on the header layer. No sleeps, no thresholds, and the sibling test pins the negative (`["spent-token"]` only, retry still on screen). That is a deterministic kill, not a timing race. The E2E delta is additive — seeding a token so the challenged-turn assertion does not sit in the new wait — which is necessary and does not weaken anything.

Gates reproduced: worker **176/176**, web **1072/1072**.

---

## Two things to know before you merge

1. **I saw one test failure, once.** My first full web run on this tree came back `1 failed | 1071 passed`. Nine consecutive runs afterwards were clean and I could not capture the name — the failing run was the first after a cold `pnpm install`, so it looks like cold-cache timing sensitivity rather than live logic. Not blocking and not a reason to hold the merge, but since you have just finished chasing a pool-starvation flake, it is worth watching the next few CI runs; if it resurfaces, the wait-mechanism tests are the first place I would look.

2. **The branch is behind base again** — `2083b1c5` (#273 T2) landed after this was rebased. The merge is now textually **clean**, so this is low-risk, unlike last round. Just take the post-merge CI green as the real confirmation, since these numbers were again measured on a tree that is one commit short.

Carried-over P3, unchanged and still fine to ignore: `remoteip` is sent as an empty value when `CF-Connecting-IP` is absent.

Deferring full challenge recovery (photo re-send without re-picking the file) to #471 is the right call — it is a UX nicety, and the security and availability properties are all closed here.

## Verdict

**Approve — merge it.** Every blocking item I raised across both rounds is closed and independently verified.

Issue #452, opened 2026-07-28
Distinguish 'could not verify' from 'credential rejected' so a JWKS outage does not 401-loop clients


lifeodyssey:
## Problem

#448 (fixing #441) makes a presented-but-unverifiable bearer credential return 401 instead of
silently becoming an anonymous identity. That is correct for a *rejected* credential — the client's
D8 path clears the token and re-authenticates.

But `verifySupabase` / `verifyNeon` / `verifyJwt` collapse two very different failures into the same
`INVALID` verdict:

- **credential-rejected** — expired, bad signature, wrong issuer/audience/algorithm. The token really
  is no good; 401 is right and refreshing fixes it.
- **could-not-verify** — the JWKS fetch failed, timed out, returned a 5xx, or the cached remote key
  set could not be refreshed. The token may be perfectly valid; we simply could not check it.

Both paths run through the same `catch { return INVALID; }`, so a JWKS outage becomes a hard 401 for
every logged-in caller. A client built to treat 401 as "token is stale" will clear its token, refresh,
get a fresh and equally unverifiable token, and 401 again — a refresh loop for the duration of the
outage, on top of the outage itself.

## Why it was accepted in #448

Failing closed is the right default for an auth decision, and #441's whole point is that failing
*open* into the anonymous pool is worse. Distinguishing the two requires reworking the jose error
handling (`JWKSNoMatchingKey` / `JOSEError` subclasses vs. fetch/transport failures) and picking a
response for the unverifiable case — 503 with `Retry-After`, or a short-lived degraded mode — which
is a design decision, not a bug fix. Documented as a known trade-off in #448 rather than silently
shipped.

## Suggested direction

1. Split the reason further: `"credential_rejected"` vs `"verification_unavailable"`.
2. Map `verification_unavailable` to 503 + `Retry-After` so the client backs off instead of
   entering the refresh path.
3. Make sure the web client's D8/D11 classification does not treat 503 as a token problem.
4. Add edge tests: JWKS returning 500, JWKS timing out, and JWKS serving a key set with no matching
   `kid` (that last one IS a rejection, not an outage — worth pinning explicitly).

## Origin

Raised as P2-3 in the second-seat review of #448.

PR #467, opened 2026-07-28
feat(web): BYOK storage wrapper + header wiring (#284 Task 6, storage half)

lifeodyssey:
## Summary

Implements the **storage/non-render half** of Task 6 from the BYOK spec (#284, `docs/superpowers/specs/2026-07-28-284-byok-design.md`, revision efc0abaa). Per the split with the UI wave:

**In this PR (non-UI):**
- `apps/web/src/lib/byok/byokStorage.ts` — session-scoped BYOK credential storage. SSR-safe and Safari-partitioned-storage-safe (every `sessionStorage` access, **including the property read itself**, is wrapped in `try`/`catch`; see "Review round 2" below). Exposes `saveByokConfig`/`getByokConfig`/`clearByokConfig`, `validateByokConfig` (model required for openai-compatible only, per OQ-1; key required + character-safety validated for every family), `byokHeaders()` (`X-BYOK-Provider/Key/Model/Base-Url`), vision-flag accessors cleared in lockstep with the credential.
- `apps/web/src/features/chat/session-headers.ts` — the shared `sessionHeaders()` (extracted by #445, used by both chat and photo search) now spreads `byokHeaders()` on top of auth/turnstile headers.
- `apps/web/src/features/chat/byok-i18n.ts` (new) + `i18n.ts` wiring — ja/zh/en BYOK panel copy, including key-required/key-invalid/model-required validation messages.
- Tests: see "Test plan" below.

**Left for the UI wave** (explicitly noted here): `ByokSettings.tsx`, `ChatInput.tsx` entry point, `chat.css` design tokens, and the browser-level ACs that need those components to exist (inline error render, "key not accepted" state, vision badge render, design-token grep).

## Review round 2 — rebase + Fable/Opus follow-up

Both review seats (Fable, Opus) returned `request_changes` against the first revision (base `fa3f2385`, before #445 landed). This revision addresses both in one pass:

**Rebase (Fable P1-1).** #445 merged and extracted the shared `sessionHeaders()` out of `use-chat-session.ts` into `apps/web/src/features/chat/session-headers.ts` (used by both chat and the new photo-search feature). Rebased onto current `feat/frontend-rebuild`; the BYOK header injection point moved to `session-headers.ts` instead of re-adding it to `use-chat-session.ts`, with a new dedicated `session-headers.test.ts`.

**Photo-search BYOK semantics — explicit ruling (Fable P1-2, Fable architecture seat).** Because photo search now shares `sessionHeaders()` with chat, a saved BYOK credential's headers reach photo search's upload request too. **Ruling: this is the correct semantics, not an accident of sharing a module.** The vision-capability probe/badge (Task 5, D5) exists precisely so a BYOK user's image turns are answered by their own key, and Task 9 (per-identity rate limiting) requires a BYOK turn to never silently fall back to the platform key — photo search is a BYOK turn like any other. Documented in `session-headers.ts`'s doc comment and covered by a dedicated "BYOK headers on photo search" describe block in `photo-search-client.test.ts` (asserts the full header set present when a credential is saved, and absent when none is).

**Opus P1-2 — SecurityError from the `sessionStorage` accessor itself (reproduced).** `typeof window.sessionStorage` **evaluates the accessor**, not just a type check — Safari's partitioned/blocked-storage mode (and similar restrictions) implements `sessionStorage` as a getter that throws `SecurityError` on read, not merely on `.getItem()`/`.setItem()`. Before this fix, that would have propagated out of `byokHeaders()` and failed **every chat turn for a user who has never touched BYOK**. Fixed: `trySessionStorage()` wraps the property read itself in `try`/`catch`, and every storage operation (`safeGet`/`safeSet`/`safeRemove`) independently wraps its call too (a `.getItem()`-level throw is a distinct failure mode from an accessor-level throw). Reproduced in jsdom via `Object.defineProperty(window, "sessionStorage", { get() { throw ... } })` (accessor-level) and `vi.spyOn(Storage.prototype, "getItem")` (method-level — jsdom's `Storage` silently ignores a direct instance-property override, so the prototype spy is what actually takes effect) in `byokStorage-security-error.test.ts`.

**P2 follow-ups (Fable + Opus, merged into one pass):**
1. **Key validation extended.** `validateByokConfig` now rejects an empty/whitespace key (`key_required` — previously this would have shipped `X-BYOK-Key: ""` and relied on a guaranteed 400 downstream) and a key containing `\r`/`\n` or any non-Latin-1/control character (`key_invalid` — previously this would have thrown an opaque, per-turn-**sticky** `TypeError` from `Headers()`/`fetch()` deep in the transport instead of a validation rejection surfaced once at save time). New `apiKeyRequired`/`apiKeyInvalid` i18n copy added.
2. **Orphaned vision flag fixed.** `setByokVisionSupported()` is now a no-op when no config is saved, closing a probe-after-clear race: a vision probe that resolves after `clearByokConfig()` (or before any config was ever saved) could previously leave a vision flag with nothing to attach to, which a later, unrelated `saveByokConfig()` call would then silently inherit.
3. **AC1 grep upgraded to full-tree.** The "no component touches `sessionStorage` directly" guard now recursively walks all of `apps/web/src/` (excluding `byokStorage.ts`) instead of two flat directories — the AC says "no component", not "no component in these two folders". Verified this costs single-digit milliseconds locally; no reason to keep the narrower scan.
4. **Test-file budget.** `byokStorage.test.ts` (was 333 lines) split into `byokStorage.test.ts` / `byokStorage-validation.test.ts` / `byokStorage-security-error.test.ts`; oversized `describe` blocks in `byokStorage.test.ts`, `byokStorage-validation.test.ts`, and `byok-i18n.test.ts` split further to satisfy the repo's 50-line-per-function lint rule on test bodies. The originally-flagged 238-line `use-chat-session.test.tsx` addition was removed entirely once the injection point moved to `session-headers.ts` (no BYOK-specific changes remain in that file — it's back to its pre-this-PR state).
5. **P3 (baseUrl / naming):** `baseUrl` of `""` or whitespace-only is now treated as absent for header purposes, same as `undefined`. Added a one-line jsdoc note reconciling the spec prose's "readByokConfig" with this module's `getByokConfig` export name.

**Threat-model wording corrected.** `byokStorage.ts`'s header comment previously implied `sessionStorage` was a hard per-tab boundary. It isn't (per threat T11): a duplicated tab or a browser-restored session can carry `sessionStorage` along. Wording now says the credential is bounded to the browser session, not to a single tab — no design change, just an accuracy fix.

**Residual risk tracked separately.** OQ-6 (a CSP for `apps/web` as BYOK residual-XSS hardening) and OQ-7 (retrofitting the SSRF egress guard onto pre-existing egress paths like `web_tools`/image proxy) are out of Task 6's scope and tracked as #469 and #470 respectively.

## Task 6 AC coverage (9 total)

| # | AC | Status |
|---|---|---|
| 1 | No component calls `sessionStorage` directly (spy + grep) | ✅ storage half done; grep test now walks the full `src/` tree |
| 2 | `sessionHeaders()` emits the right X-BYOK-* set / `{}` set when unsaved | ✅ done + tested (now at the `session-headers.ts` injection point) |
| 3 | Vision badge write/clear lockstep with `clearByokConfig()` | ✅ storage lockstep done (incl. the orphan-flag race fix); the visible badge render is UI-wave |
| 4 | Model required for openai-compatible; pre-filled default for anthropic/gemini | ✅ `validateByokConfig` + `BYOK_DEFAULT_MODEL` done; the inline validation message *render* is UI-wave |
| 5 | SSR-safe, corrupt/absent → null, `byokHeaders()` → `{}` | ✅ done + tested, now also covering Safari-partitioned-storage `SecurityError` |
| 6 | Inline error inside settings panel, no raw key in DOM | ⏳ UI-wave (browser AC, needs the component) |
| 7 | Rejected credential doesn't silently continue on platform key | ⏳ UI-wave (browser AC) |
| 8 | i18n: every BYOK string resolves ja/zh/en, no leaked keys | ✅ done + tested |
| 9 | Design tokens grep (no `bg-white`/`bg-gray-`/hardcoded palette) | ⏳ UI-wave (no CSS touched yet) |

## Mutation testing evidence

No mutation-testing tool is configured in the repo. Verified by hand across both review rounds — introduced and reverted each mutation, re-running the relevant unit suite:

Round 1:
1. Inverted the openai-compatible model-required condition → **caught**
2. Stopped clearing the vision flag on save/clear → **caught**
3. Dropped the provider gate on the `X-BYOK-Base-Url` header → **initially survived** — added a test asserting the header stays omitted for anthropic even with a stray `baseUrl` present, re-ran: **caught**
4. Forced `hasSessionStorage()` to always return `true` (breaks SSR safety) → **caught**
5. Made `isByokProvider` accept any string → **caught**

Round 2 (this revision):
6. Dropped the `try`/`catch` around the `sessionStorage` property read (reverting to a bare `typeof` guard) → **caught** by `byokStorage-security-error.test.ts`
7. Weakened `HEADER_SAFE` to only reject `\r`/`\n`, allowing non-Latin-1 through → **caught** by `byokStorage-validation.test.ts`
8. Removed the orphaned-vision-flag guard (`setByokVisionSupported` writing with no config saved) → **caught**
9. Removed the `try`/`catch` inside `safeGet` around `.getItem()` (keeping only the property-level guard) → **caught** by the `Storage.prototype.getItem` spy test — confirming the two guard layers are independently load-bearing

All mutations reverted; working tree confirmed clean after each sweep.

## Gates

- `pnpm test` (apps/web, full suite): **1090 tests, all passing** (`Test Files 170 passed`, `Tests 1090 passed`)
- Coverage: `src/lib/byok/byokStorage.ts` at 100% stmts/branches/funcs/lines (fully covered — no longer listed in the istanbul non-100% report); overall apps/web coverage 98.7% stmts / 95.13% branches / 99.16% funcs / 99.56% lines, well above the 90/90/90/90 floor in `apps/web/AGENTS.md`
- `pnpm run typecheck`: clean
- `pnpm run lint:oxlint` (type-aware, `--deny-warnings`): clean
- No `eslint-disable`/`@ts-ignore`/suppression added

## Test plan

- [x] `apps/web/tests/unit/byokStorage.test.ts` — persistence, header emission (incl. Base-Url edge cases), vision lockstep (incl. orphan-flag fix), null/corrupt storage
- [x] `apps/web/tests/unit/byokStorage-validation.test.ts` — model-required (OQ-1) and key-required/key-invalid (character-safety) validation rules
- [x] `apps/web/tests/unit/byokStorage-security-error.test.ts` — SecurityError from both the `sessionStorage` accessor and a method call, reproduced in jsdom
- [x] `apps/web/tests/unit/byokStorage-ssr.test.ts` — node-environment SSR safety (no `window` global)
- [x] `apps/web/tests/unit/byokStorage-source-guard.test.ts` — AC1 full-tree grep guard + no top-level `window` access
- [x] `apps/web/tests/unit/chat/byok-i18n.test.ts` — ja/zh/en copy resolves, no leaked raw keys/header names
- [x] `apps/web/tests/unit/chat/session-headers.test.ts` — the shared injection point directly: identity headers (regression) + BYOK header combinations
- [x] `apps/web/tests/unit/chat/photo-search-client.test.ts` — new "BYOK headers on photo search" describe block asserting the deliberate semantics

References issue #284, Task 6 of `docs/superpowers/specs/2026-07-28-284-byok-design.md` (rev4, efc0abaa). Residual risk tracked as #469 (CSP/OQ-6) and #470 (egress-guard retrofit/OQ-7). Not merging per instructions — for review only.

---

## Review round 3 pushed — SHA `fb975550061a527c1d7b07d2f3c45705809475fb`

Addressed both approval-gated follow-ups from round 2's sign-off:

**P2 — HEADER_SAFE extended to model (all families) and base_url.** Previously `model`'s character-safety check only ran for the openai-compatible family, and `base_url` had no character check at all — so a Japanese model name on anthropic/gemini, a CRLF-injected model on any family, or an internationalized-domain (non-ASCII) `base_url` would all have saved successfully and only crashed `Headers()`/`fetch()` later, at turn time (a sticky per-turn `TypeError`). Both now go through `HEADER_SAFE`, with a new `base_url_invalid` error code (IDN hosts are rejected outright rather than punycode-decoded here — the settings panel is expected to prompt for an ASCII/A-label host, matching what the Task 1 egress guard already expects on the wire). Five probes added in `byokStorage-field-safety.test.ts`: Japanese model on anthropic, CRLF-injected model on gemini, IDN base_url, CRLF-injected base_url, and an ASCII base_url regression. New `baseUrlInvalid` i18n copy (ja/zh/en).

**P3 — genuine safeSet catch coverage.** The `safeSet` catch block previously showed 100% coverage only via an accessor-level `SecurityError` block, which short-circuits before ever reaching `.setItem()` — a false-positive signal, exactly as flagged. Added a `Storage.prototype.setItem`-throwing (`QuotaExceededError`, reproducing Safari private-mode) test exercising `saveByokConfig()` and `setByokVisionSupported()` through the actual catch path in `byokStorage-security-error.test.ts`. Verified by mutation: removing the `try`/`catch` around `.setItem()` makes both new tests fail.

Also trimmed `byokStorage.test.ts`'s comments (202 → 199 lines) to clear the ~200-line test-file budget.

Gates: `pnpm test` 1102 tests passing, `byokStorage.ts` at 100% coverage, typecheck clean, `lint:oxlint --deny-warnings` clean. 2 new mutations (model-family restriction, dropped base_url check) + the safeSet-catch mutation all caught; working tree confirmed clean after each sweep.

Aware I'm queued behind #463 in the merge order and may need a rebase once that lands (Turnstile wait-logic changes to `session-headers.ts`) — will handle when it's my turn.

### DIFF
--- a/apps/web/src/features/chat/byok-i18n.ts
+++ b/apps/web/src/features/chat/byok-i18n.ts
@@ -1,0 +1,104 @@
+/**
+ * BYOK settings panel copy (issue #284 Task 6).
+ *
+ * Covers the three-family selector, the model field (required for
+ * openai-compatible, pre-filled elsewhere), the base_url helper, the vision
+ * badge, and every error code the container/edge can surface for a BYOK
+ * turn. Kept feature-local and split out of `i18n.ts`, same as
+ * `route-i18n.ts` / `search-i18n.ts`.
+ */
+export interface ChatByokDict {
+  readonly title: string;
+  readonly familyLabel: string;
+  readonly familyOpenaiCompatible: string;
+  readonly familyAnthropic: string;
+  readonly familyGemini: string;
+  readonly apiKeyLabel: string;
+  readonly modelLabel: string;
+  readonly modelRequired: string;
+  readonly baseUrlLabel: string;
+  readonly baseUrlHelp: string;
+  readonly save: string;
+  readonly clear: string;
+  readonly maskedSummary: string;
+  readonly visionBadge: string;
+  readonly errorCredentialRejected: string;
+  readonly errorRequiresLogin: string;
+  readonly errorInvalidRequest: string;
+  readonly errorEgressBlocked: string;
+  readonly notAccepted: string;
+  readonly anonymousTeaser: string;
+  readonly signInToSetUp: string;
+}
+
+export const jaByok: ChatByokDict = {
+  title: "自分のAPIキーを使う",
+  familyLabel: "サービスをえらんでね",
+  familyOpenaiCompatible: "OpenAI互換",
+  familyAnthropic: "Anthropic",
+  familyGemini: "Gemini",
+  apiKeyLabel: "APIキー",
+  modelLabel: "モデル名",
+  modelRequired: "モデル名を入力してね",
+  baseUrlLabel: "接続先URL",
+  baseUrlHelp: "https:// から始まるURLを入れてね",
+  save: "保存する",
+  clear: "解除する",
+  maskedSummary: "登録ずみ(キーはひみつ)",
+  visionBadge: "✓ 画像対応",
+  errorCredentialRejected: "このキーは使えなかったみたい。もう一度確かめてね",
+  errorRequiresLogin: "自分のキーを使うにはログインしてね",
+  errorInvalidRequest: "入力に不備があるみたい。確認してね",
+  errorEgressBlocked: "その接続先には安全のためつなげられないよ",
+  notAccepted: "キーが受け付けられなかったよ。標準のモードでは続けないから、直してね",
+  anonymousTeaser: "自分のAPIキーを使うと、待たずにたくさん話せるよ",
+  signInToSetUp: "ログインして設定する",
+};
+
+export const zhByok: ChatByokDict = {
+  title: "使用你自己的 API 密钥",
+  familyLabel: "选择服务商",
+  familyOpenaiCompatible: "OpenAI 兼容",
+  familyAnthropic: "Anthropic",
+  familyGemini: "Gemini",
+  apiKeyLabel: "API 密钥",
+  modelLabel: "模型名称",
+  modelRequired: "请填写模型名称",
+  baseUrlLabel: "接口地址",
+  baseUrlHelp: "请填写以 https:// 开头的地址",
+  save: "保存",
+  clear: "清除",
+  maskedSummary: "已保存(密钥已隐藏)",
+  visionBadge: "✓ 支持图片",
+  errorCredentialRejected: "这个密钥没能用,请再检查一下",
+  errorRequiresLogin: "使用自己的密钥需要先登录",
+  errorInvalidRequest: "填写的内容好像不对,请检查一下",
+  errorEgressBlocked: "出于安全考虑,无法连接这个地址",
+  notAccepted: "密钥没有被接受。不会用默认模式继续,请先修正",
+  anonymousTeaser: "用你自己的密钥,就不用受限额限制啦",
+  signInToSetUp: "登录后设置",
+};
+
+export const enByok: ChatByokDict = {
+  title: "Use your own API key",
+  familyLabel: "Choose a provider",
+  familyOpenaiCompatible: "OpenAI-compatible",
+  familyAnthropic: "Anthropic",
+  familyGemini: "Gemini",
+  apiKeyLabel: "API key",
+  modelLabel: "Model name",
+  modelRequired: "Please enter a model name",
+  baseUrlLabel: "Base URL",
+  baseUrlHelp: "Enter a URL starting with https://",
+  save: "Save",
+  clear: "Clear",
+  maskedSummary: "Saved (key hidden)",
+  visionBadge: "✓ Image support",
+  errorCredentialRejected: "That key wasn't accepted. Please double-check it",
+  errorRequiresLogin: "Sign in to use your own key",
+  errorInvalidRequest: "Something in that entry looks off — please check it",
+  errorEgressBlocked: "That address can't be reached for safety reasons",
+  notAccepted: "Your key was not accepted. We won't continue on the standard mode — please fix it",
+  anonymousTeaser: "Bring your own key to chat without the daily limit",
+  signInToSetUp: "Sign in to set up",
+};
--- a/apps/web/src/features/chat/byok-i18n.ts
+++ b/apps/web/src/features/chat/byok-i18n.ts
@@ -20,6 +20,7 @@ export interface ChatByokDict {
   readonly modelRequired: string;
   readonly baseUrlLabel: string;
   readonly baseUrlHelp: string;
+  readonly baseUrlInvalid: string;
   readonly save: string;
   readonly clear: string;
   readonly maskedSummary: string;
@@ -46,6 +47,7 @@ export const jaByok: ChatByokDict = {
   modelRequired: "モデル名を入力してね",
   baseUrlLabel: "接続先URL",
   baseUrlHelp: "https:// から始まるURLを入れてね",
+  baseUrlInvalid: "そのURLは使えない文字が入ってるみたい。ローマ字のアドレスを試してね",
   save: "保存する",
   clear: "解除する",
   maskedSummary: "登録ずみ(キーはひみつ)",
@@ -72,6 +74,7 @@ export const zhByok: ChatByokDict = {
   modelRequired: "请填写模型名称",
   baseUrlLabel: "接口地址",
   baseUrlHelp: "请填写以 https:// 开头的地址",
+  baseUrlInvalid: "这个地址里有不能使用的字符,请换成英文/数字格式的地址试试",
   save: "保存",
   clear: "清除",
   maskedSummary: "已保存(密钥已隐藏)",
@@ -98,6 +101,7 @@ export const enByok: ChatByokDict = {
   modelRequired: "Please enter a model name",
   baseUrlLabel: "Base URL",
   baseUrlHelp: "Enter a URL starting with https://",
+  baseUrlInvalid: "That URL contains characters that can't be used — try an ASCII/punycode address",
   save: "Save",
   clear: "Clear",
   maskedSummary: "Saved (key hidden)",
--- a/apps/web/src/features/chat/byok-i18n.ts
+++ b/apps/web/src/features/chat/byok-i18n.ts
@@ -14,6 +14,8 @@ export interface ChatByokDict {
   readonly familyAnthropic: string;
   readonly familyGemini: string;
   readonly apiKeyLabel: string;
+  readonly apiKeyRequired: string;
+  readonly apiKeyInvalid: string;
   readonly modelLabel: string;
   readonly modelRequired: string;
   readonly baseUrlLabel: string;
@@ -38,6 +40,8 @@ export const jaByok: ChatByokDict = {
   familyAnthropic: "Anthropic",
   familyGemini: "Gemini",
   apiKeyLabel: "APIキー",
+  apiKeyRequired: "APIキーを入力してね",
+  apiKeyInvalid: "そのAPIキーは使えない文字が入ってるみたい。確認してね",
   modelLabel: "モデル名",
   modelRequired: "モデル名を入力してね",
   baseUrlLabel: "接続先URL",
@@ -62,6 +66,8 @@ export const zhByok: ChatByokDict = {
   familyAnthropic: "Anthropic",
   familyGemini: "Gemini",
   apiKeyLabel: "API 密钥",
+  apiKeyRequired: "请填写 API 密钥",
+  apiKeyInvalid: "这个 API 密钥里有不能使用的字符,请检查一下",
   modelLabel: "模型名称",
   modelRequired: "请填写模型名称",
   baseUrlLabel: "接口地址",
@@ -86,6 +92,8 @@ export const enByok: ChatByokDict = {
   familyAnthropic: "Anthropic",
   familyGemini: "Gemini",
   apiKeyLabel: "API key",
+  apiKeyRequired: "Please enter an API key",
+  apiKeyInvalid: "That API key contains characters that can't be used — please check it",
   modelLabel: "Model name",
   modelRequired: "Please enter a model name",
   baseUrlLabel: "Base URL",
--- a/apps/web/src/features/chat/i18n.ts
+++ b/apps/web/src/features/chat/i18n.ts
@@ -1,4 +1,6 @@
 import type { Locale } from "../../i18n/locales";
+import { enByok, jaByok, zhByok } from "./byok-i18n";
+import type { ChatByokDict } from "./byok-i18n";
 import {
   enClarify,
   enDeparture,
@@ -28,6 +30,7 @@ import type { ChatSearchDict } from "./search-i18n";
 
 export type { ChatRouteDict } from "./route-i18n";
 export type { ChatSearchDict } from "./search-i18n";
+export type { ChatByokDict } from "./byok-i18n";
 export type {
   ChatClarifyDict,
   ChatDepartureDict,
@@ -120,6 +123,7 @@ export interface ChatDict {
   readonly departure: ChatDepartureDict;
   readonly location: ChatLocationDict;
   readonly photo: ChatPhotoDict;
+  readonly byok: ChatByokDict;
 }
 
 const jaToolSteps: ChatToolStepsDict = {
@@ -193,6 +197,7 @@ const ja: ChatDict = {
   departure: jaDeparture,
   location: jaLocation,
   photo: jaPhoto,
+  byok: jaByok,
 };
 
 const zh: ChatDict = {
@@ -220,6 +225,7 @@ const zh: ChatDict = {
   departure: zhDeparture,
   location: zhLocation,
   photo: zhPhoto,
+  byok: zhByok,
 };
 
 const en: ChatDict = {
@@ -251,6 +257,7 @@ const en: ChatDict = {
   departure: enDeparture,
   location: enLocation,
   photo: enPhoto,
+  byok: enByok,
 };
 
 const CHAT_DICTIONARIES: Record<Locale, ChatDict> = { ja, zh, en };
--- a/apps/web/src/features/chat/session-headers.ts
+++ b/apps/web/src/features/chat/session-headers.ts
@@ -1,5 +1,6 @@
-import { authHeaders } from "../../lib/auth/authSession";
 import { configuredTurnstileSiteKey } from "../../components/TurnstileGate";
+import { authHeaders } from "../../lib/auth/authSession";
+import { byokHeaders } from "../../lib/byok/byokStorage";
 import { awaitTurnstileToken, turnstileHeaders } from "../../lib/turnstile/tokenStore";
 
 /**
@@ -8,12 +9,23 @@ import { awaitTurnstileToken, turnstileHeaders } from "../../lib/turnstile/token
  * carry the held Turnstile token (S1.9 #281) — one solved challenge covers
  * every turn in its window. Used by the chat transport and photo search so
  * both surfaces hit the edge with identical identity semantics.
+ *
+ * A saved BYOK credential (#284 Task 6) adds its `X-BYOK-*` headers on top of
+ * every caller of this function — deliberately, not by omission. Photo
+ * search's vision probe/badge (Task 5, D5) exists precisely so a BYOK user's
+ * image turns are answered by their own key; T9 requires a BYOK turn to
+ * never silently fall back to the platform key, so photo search inheriting
+ * the same headers as chat is the correct semantics, not an accident of
+ * sharing this module. `byokHeaders()` returns `{}` with nothing saved, so
+ * this is a no-op for every caller until a credential exists. It is read
+ * synchronously and does not participate in the Turnstile wait below in any
+ * way — its ordering relative to that wait has no observable effect.
  */
 export async function sessionHeaders(sessionId?: string): Promise<Record<string, string>> {
   const base: Record<string, string> = sessionId ? { "x-session-id": sessionId } : {};
   const auth = await authHeaders();
-  if (auth.Authorization !== undefined) return { ...base, ...auth };
-  return { ...base, ...(await challengeHeaders()) };
+  const challenge = await challengeHeaders(auth);
+  return { ...base, ...challenge, ...auth, ...byokHeaders() };
 }
 
 /**
@@ -23,14 +35,17 @@ export async function sessionHeaders(sessionId?: string): Promise<Record<string,
  * photo search (#445 added `/v1/photo-search` to it) — so a request fired
  * before the widget has solved is rejected. Chat surfaces that as a retryable
  * challenge; photo upload has no challenge UI of its own, which is exactly why
- * the wait lives here, in the header path both surfaces share.
+ * the wait lives here, in the header path both surfaces share. Skipped
+ * entirely once authenticated — the same short-circuit the pre-wait version
+ * of this function had.
  *
  * When this build renders no widget there is nothing to wait for, so the
  * request goes out immediately and any rejection surfaces normally. The wait
  * itself is bounded (`TURNSTILE_WAIT_MS`) — it delays a turn, it never hangs
  * one.
  */
-async function challengeHeaders(): Promise<Record<string, string>> {
+async function challengeHeaders(auth: Record<string, string>): Promise<Record<string, string>> {
+  if (auth.Authorization !== undefined) return {};
   if (configuredTurnstileSiteKey() !== undefined) await awaitTurnstileToken();
   return turnstileHeaders();
 }
--- a/apps/web/src/lib/byok/byokStorage.ts
+++ b/apps/web/src/lib/byok/byokStorage.ts
@@ -35,7 +35,7 @@ export interface ByokConfig {
   readonly baseUrl?: string;
 }
 
-export type ByokSaveError = "model_required" | "key_required" | "key_invalid";
+export type ByokSaveError = "model_required" | "key_required" | "key_invalid" | "base_url_invalid";
 export type ByokSaveResult = { readonly ok: true } | { readonly ok: false; readonly error: ByokSaveError };
 
 const CONFIG_KEY = "animichi.byok.config";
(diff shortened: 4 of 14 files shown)


lifeodyssey (commented):
## Second-seat review (Opus) — independent, spec rev5 `efc0abaa`

Verdict: **request_changes**. The storage module itself is well-shaped and its unit tests are real (I reproduced two mutations and both die). What blocks it is (a) the branch is cut from a **stale base** and its one integration point has since been refactored into a helper shared with photo-search, and (b) `byokHeaders()` now sits in the hot path of *every* chat request and can throw.

Gates I ran myself in a detached worktree at `c1c9c641`: `pnpm run typecheck` exit 0, `pnpm run lint:oxlint --deny-warnings` exit 0, `pnpm vitest run --coverage` **156 files / 974 tests passed**, thresholds (95/94/95/95) met, All files 98.94% stmts / 95.34% br. No suppression added. (PR body says 993 tests; I count 974 — cosmetic.) No Codecov status is attached to this PR — only 2 checks ran.

---

### P1-1 — Stale base: the function this PR patches no longer exists, and the rebase silently widens the credential's blast radius

`origin/feat/frontend-rebuild` is at `2726b677` (#445) — this branch's parent is `fa3f2385` (#458). Verified: `git merge-base --is-ancestor 2726b677 HEAD` → 1. Neither #445 nor #461 is in the base.

On the real base, `sessionHeaders()` has been extracted to `apps/web/src/features/chat/session-headers.ts` and is imported by **two** callers: `use-chat-session.ts` *and* `features/chat/photo-search.ts`. So the natural rebase resolution puts `...byokHeaders()` into the shared helper, and the user's raw provider key starts riding on every `/v1/photo-search` upload and confirm ping — an endpoint with no BYOK consumption, no Task 2 inbound stripping, and a multipart body. The spec's non-goals confine BYOK to chat + `/v1/byok/probe`.

This is not hypothetical ambiguity: the real base's `PhotoSearchUpload.tsx` already ships `quotaNoByok` / `quotaByokNoVision` copy and a `switch_vision_endpoint` guidance code, i.e. photo search *wants* a BYOK vision key eventually — so a reviewer resolving the conflict will plausibly think spreading it there is correct. Decide it deliberately, in code, now.

Fix: rebase onto `2726b677`; attach `byokHeaders()` at the **chat transport call site** (or a `chatSessionHeaders()` wrapper), never inside the shared `sessionHeaders()`; add a test pinning that a photo-search request carries **no** `X-BYOK-*`. Note `worker/app.ts` on the real base already does `headers.delete("x-byok-endpoint")` with the comment "closed until BYOK launches" — the new `X-BYOK-*` set has no equivalent gate, which is Task 4's job but worth stating in the card. `i18n.ts` will also conflict (#445 split `error-states-i18n.ts` out).

### P1-2 — `byokHeaders()` throws in blocked-storage browsers, breaking chat for users who never touched BYOK

`hasSessionStorage()` guards only `typeof window === "undefined"` and `typeof window.sessionStorage !== "undefined"`. `typeof` on a member expression evaluates the getter — and in Safari with cookies/site-data blocked, in a partitioned third-party frame, or under some privacy extensions, that getter **throws `SecurityError`**. Reproduced in jsdom by redefining the property with a throwing getter:

```
BLOCKED-STORAGE byokHeaders() threw: SecurityError: The operation is insecure.
```

Because the PR spreads `byokHeaders()` into `sessionHeaders()`, that exception rejects the headers promise for *every* chat turn, for every affected user, BYOK configured or not. The module docstring's claim — "every accessor checks availability first and degrades to 'no credential' rather than throwing" — is only true for the `undefined` case. `setItem` can also throw `QuotaExceededError` (Safari private mode).

Fix: wrap the probe and each `getItem`/`setItem`/`removeItem` in `try/catch`, returning `null` / no-op. Add a unit test with a throwing `sessionStorage` getter (the exact shape above) and one with a throwing `setItem`.

### P1-3 — AC1's grep guard is directory-scoped and non-recursive; the AC is claimed ✅ but escapes trivially

Spec AC1 asks for "a lint-level grep assertion that `sessionStorage` appears only in `byokStorage.ts`" — repo-wide. The test scans exactly two directories with a non-recursive `readdirSync`. I planted a `window.sessionStorage.getItem(...)` line in four places and ran the guard:

| planted in | guard |
|---|---|
| `features/chat/components/ChatInput.tsx` | **caught** (exit 1) |
| `features/chat/components/ErrorStates/BudgetExhausted.tsx` | **survives** |
| `src/lib/byok/nested/leak.ts` | **survives** |
| `src/lib/chat/byokLeak.ts` (any other dir) | **survives** |

A repo-wide recursive assertion is achievable with **zero carve-outs today** — I grepped `apps/web/src`: `sessionStorage` currently appears only in `byokStorage.ts` (the turnstile `tokenStore.ts` precedent does not use it). So there is no reason to scope it. Also, the PR body describes the chat components dir as "(currently empty)"; it holds 19 entries plus `ErrorStates/`.

Fix: walk `apps/web/src` recursively, assert the set of matching files equals `["src/lib/byok/byokStorage.ts"]`, and until then mark AC1 partial rather than ✅.

### P1-4 — The spec's own merge precondition is unmet: the CSP follow-up issue does not exist

Spec L136-139 accepts the XSS-reads-`sessionStorage` residual risk **on the condition** that "the real mitigation is XSS prevention + CSP, deferred by the OQ-6 ruling to a follow-up issue that **must exist and be linked here before this card merges**" (repeated in Verification Plan §8 and the OQ-6/OQ-7 rulings). `gh issue list --state all --search CSP` returns nothing; no open issue covers CSP hardening or the OQ-7 SSRF retrofit. This PR is the one that actually puts a user credential into `sessionStorage`, so the condition binds here.

Fix: file + link the CSP hardening issue and the OQ-7 retrofit issue.

---

### On the XSS surface itself (asked explicitly)

I agree with the spec's ruling and found nothing cheap left on the table at this layer. Client-side encryption is theater (the key sits next to the ciphertext); an in-memory-only store loses the reload survival that is the whole point and is equally readable by injected script running in the page. The genuine mitigations are CSP + input sanitization — deferred by OQ-6, which is exactly why P1-4 matters. Two smaller notes:

- `JSON.parse` handling is genuinely defensive: `try/catch`, shape validation, and it correctly rejects `null`, `42`, unknown provider strings, and missing fields. I confirmed `isByokProvider` is load-bearing (mutation: accept any string → caught). What is *not* validated is the *lexical* content of the strings — see P2-1.
- **Cross-tab isolation is weaker than T11 states.** T11 says `sessionStorage` is "per-tab, cleared on tab close". In practice Chrome/Firefox **copy** `sessionStorage` into a duplicated tab and into a tab opened via "Duplicate"/some middle-click paths, and Chrome's "continue where you left off" session restore **repopulates** it after a browser restart. So the credential can outlive the tab the user thinks they closed. That is a threat-model text correction (P3), not a code change — but the card should not claim tab-close is a hard boundary.

---

### P2 findings

**P2-1 — header values are stored and emitted with no hygiene; the failure mode is an opaque, sticky turn failure.** I probed the real `Headers` semantics:

| stored value | outcome |
|---|---|
| trailing `\n` (the classic paste artifact) | silently trimmed — benign |
| embedded `\r\n` | **TypeError: invalid header value** |
| non-Latin-1 (e.g. a full-width or Japanese character in the model name) | **TypeError: Cannot convert argument to a ByteString** |
| leading space | silently trimmed |

No header-splitting vulnerability — `fetch` enforces that. But the TypeError is raised deep inside the transport's header callback, so the user gets a generic chat failure, not the inline settings error AC6 mandates, and it **repeats on every turn** because the bad value stays in `sessionStorage` with no in-app way to clear it until the UI wave. A Japanese user typing a model name with an ideographic space hits this. Fix: `validateByokConfig` should trim and reject control chars / code points > U+00FF with a new error code, plus the matching `byok-i18n` string (the dict has no key for it today, so the "i18n AC done" claim will need reopening regardless).

**P2-2 — vision-flag lockstep has a hole (AC3).** `setByokVisionSupported()` writes unconditionally. Reproduced: save → `clearByokConfig()` → `setByokVisionSupported(true)` yields `config: null, vision: true` — an orphan badge with no credential. This is a real race, not a contrivance: the user clears the key while the Task 5 probe is in flight, the probe resolves afterwards. Fix: no-op when `getByokConfig() === null`; add the test.

**P2-3 — an empty API key saves cleanly.** `validateByokConfig` checks only the model, and only for openai-compatible. An empty key emits `X-BYOK-Key: ""`, which the container will read as "BYOK attempted" and fail every turn. Add a non-empty-key check (and for openai-compatible, a non-empty `baseUrl`).

**P2-4 — sequencing.** Header emission lands before Task 2 (inbound allowlist stripping) and Tasks 3/4 (consumption). `_deps.py` still calls `logfire.instrument_fastapi(...)` / `instrument_httpx()` with no BYOK stripping, and `worker/app.ts` forwards `X-BYOK-*` untouched on both the anonymous and authenticated paths. So today a saved credential is transmitted to a container that ignores it and **continues the turn on the platform key** — the silent fallback T9/T4-AC4 exists to forbid. It is harmless only because no UI can write a config yet, which is luck, not design. Recommend holding the `use-chat-session.ts` spread (the 1-line change) until T2 + T4 land, or landing it with the UI wave; the storage module and i18n can go now.

**P2-5 — repo standards.** `tests/unit/chat/use-chat-session.test.tsx` is now **238 lines**, over CLAUDE.md's 200-line test-file cap — split the BYOK describe into `use-chat-session-byok.test.tsx`. `features/chat/i18n.ts` is now **335 lines**, over the 300-line file cap (already over at 328 before this PR; the real base solved it by splitting `error-states-i18n.ts` out, which the rebase will bring). `byokStorage.test.ts` is at 198 — right at the edge.

### P3 findings

- No size cap on stored values; a multi-KB pasted `base_url` produces a 431 at the edge rather than an inline message.
- PR body's "(currently empty) chat components dir" is inaccurate (19 files) — it makes the guard look stronger than P1-3 shows it is.
- The top-level-`window` guard test (`!/^\s/` + `!startsWith("}")` heuristic) works but is brittle; it would also flag a `window.` occurrence inside a top-level template string.

---

### Scope-division audit (spec Task 6, 9 ACs)

| AC | claimed | my read |
|---|---|---|
| 1 no direct `sessionStorage` (spy + grep) | ✅ | **overclaimed** → partial (P1-3; and the render-spy half needs the panel) |
| 2 `sessionHeaders()` header set / `{}` | ✅ | ✅ genuinely — but tested against a `sessionHeaders` that does not exist on the real base (P1-1) |
| 3 vision badge lockstep | ✅ storage half | **partial** — orphan-write hole (P2-2) |
| 4 model required / default pre-fill | ✅ | ✅ for the storage half; the pre-fill+override render is honestly UI |
| 5 SSR-safe, corrupt → null, `{}` | ✅ | ✅ against the AC text; the module's broader "never throws" claim is false (P1-2) |
| 6 inline error, no raw key in DOM | ⏳ | **correct deferral** — needs the component |
| 7 no silent platform-key fallback | ⏳ | **correct deferral** as a browser AC, but note it is also unsatisfiable until Task 4 exists (P2-4) |
| 8 i18n ja/zh/en, no leaked keys | ✅ | ✅ — will need reopening for the P2-1 error code and possibly #452's "could not verify" vs "rejected" split |
| 9 design-token grep | ⏳ | **correct deferral** (no CSS touched; running it now would be vacuous) |

Honest split: **4 done, 2 overclaimed-as-done, 3 correctly deferred** — the deferrals are legitimate (all three genuinely require the component/CSS), the two overclaims are the ones to fix.

### Mutation reproductions

1. Dropped `removeItem(VISION_KEY)` from `saveByokConfig` → **caught**, exactly one failure ("clears the vision flag when the config is overwritten with a new key", `byokStorage.test.ts:114`).
2. Source-guard escape matrix above — mutation **caught** in the flat components dir, **survives** in a subdirectory or any other directory.

Working tree left clean; all mutations reverted (`git status --porcelain` empty).

lifeodyssey (commented):
## Second-seat final round (Opus) — probes re-run against `f7b96837`

Verdict: **approve**, with one P2 that must ride the UI-wave card (details below). All four P1s from my first pass are verifiably closed; I re-ran every original probe rather than reading the diff.

Gates, my own run in a fresh detached worktree at `f7b96837`: `typecheck` exit 0, `oxlint --type-aware --deny-warnings` exit 0, `vitest run --coverage` **170 files / 1090 tests passed**, All files 98.7 / 95.13 / 99.16 / 99.56 — **exactly matching the PR body's numbers** (last round's 993-vs-974 discrepancy is gone). `byokStorage.ts` verified from `coverage/lcov.info`: lines 57/57, branches 49/49, functions 19/19. Zero suppressions in the new files.

---

### 1. SecurityError — CLOSED

My original jsdom probe (redefine `window.sessionStorage` with a throwing getter, then call `byokHeaders()`):

| | before | now |
|---|---|---|
| probe output | `threw: SecurityError: The operation is insecure.` | `threw: NO` |
| return value | — | `byokHeaders() = {}`, `getByokConfig() = null` |

`trySessionStorage()` puts the *accessor evaluation* inside the `try`, which is the part that mattered; `safeGet`/`safeSet`/`safeRemove` each catch independently. The new test reproduces both layers, and its comment about jsdom ignoring an instance-level `.getItem` override (hence the `Storage.prototype` spy) is correct and non-obvious — good catch by the implementer.

One residual, **P3-1**: the method-call layer only exercises a throwing `getItem`. A throwing `setItem` (Safari private mode quota error) is unexercised, and because `safeSet`'s catch body is empty, 100% line/branch coverage **cannot** prove it — istanbul has nothing to count there. Add the case rather than trusting the number.

### 2. Recursive grep — CLOSED

Re-planted all my escape points against the full-`src/`-tree walk. Every one now dies:

| planted in | first round | now |
|---|---|---|
| `features/chat/components/ChatInput.tsx` | caught | **caught** |
| `features/chat/components/ErrorStates/BudgetExhausted.tsx` | survived | **caught** |
| `src/lib/byok/nested/leak.ts` | survived | **caught** |
| `src/lib/chat/byokLeak.ts` | survived | **caught** |
| `src/routes/zzleak.tsx` | (untested) | **caught** |

AC1's grep half is now honestly done. The rendering-spy half still belongs to the UI wave, which the AC table states.

### 3. Key hygiene — PARTIALLY closed; **P2-1 (new)**

The `apiKey` leg is fully fixed. Clean-storage matrix (each row: `validateByokConfig`, then if accepted `new Headers(byokHeaders())`):

| stored | validate | Headers() |
|---|---|---|
| key with CR-LF | `key_invalid` | rejected at save |
| key non-Latin-1 | `key_invalid` | rejected at save |
| key Latin-1 accent | `key_invalid` | rejected at save |
| key trailing newline | `key_invalid` | rejected at save |
| key blank | `key_required` | rejected at save |
| openai-compatible model non-Latin-1 | `model_required` | rejected at save |

Distinct error codes with ja/zh/en copy (`apiKeyRequired` / `apiKeyInvalid`) — the sticky-TypeError class is genuinely eliminated **on the key**.

But `HEADER_SAFE` is applied to `model` **only when the provider is openai-compatible**, and to `baseUrl` **never**. Same matrix, same clean run:

| stored | validate | saved | Headers() |
|---|---|---|---|
| anthropic model in Japanese | `{ok:true}` | yes | **THROWS TypeError** (ByteString) |
| anthropic model with CR-LF | `{ok:true}` | yes | **THROWS TypeError** (invalid header value) |
| gemini model in Japanese | `{ok:true}` | yes | **THROWS TypeError** |
| openai `baseUrl` with an IDN host | `{ok:true}` | yes | **THROWS TypeError** |
| openai `baseUrl` with CR-LF | `{ok:true}` | yes | **THROWS TypeError** |

These are not exotic inputs — they are precisely the fields the spec tells users to edit. AC4 requires the model field to be "visible and user-overridable" **for Anthropic/Gemini**, i.e. the two families where the charset check is skipped. And the non-goals state "self-hosted vLLM / a user's own relay is the core BYOK use case", with Task 1's own AC expecting an **A-label** host — so an IDN `base_url` is an anticipated input, and today it saves cleanly and then crashes every turn (chat *and* photo search) with no in-app recovery path.

Still no injection risk — `Headers()` enforces — so P2, not P0. Fix is ~3 lines: run `HEADER_SAFE` over `model` for **all** families, and over `baseUrl` when present (or punycode it), each with its own error code plus copy. **This must ride the UI-wave card**, since the panel is what first lets a user type into those two fields; it cannot bite before then, which is exactly why it is safe to approve now and unsafe to forget.

### 4. Orphan vision flag — CLOSED

My race probe (save, then `clearByokConfig()`, then `setByokVisionSupported(true)` — an in-flight probe resolving after the user cleared) now yields `config: null vision: null`; previously `vision: true`. The `if (getByokConfig() === null) return;` guard plus its doc comment cover both the after-clear and the never-saved case. AC3's storage half is now genuinely done.

### 5. Follow-up issues — P1-4 released

Both exist and are OPEN with the right scope (CSP for apps/web, spec OQ-6 precondition; SSRF egress-guard retrofit, OQ-7), and the PR body links both. The spec's substantive condition is met.

**P3-2**: the wording is "must exist **and be linked here**" — "here" being the spec/threat model. The spec branch is still at `efc0abaa` and contains zero occurrences of either issue reference (I checked; my first-round grep hits at L340/L520 were false positives — the IPv6 literal `2606:4700::1` contains the digit substring). Bookkeeping, not a blocker: add the two links next to the "Explicitly accepted residual risk" paragraph.

### 6. Photo-search — ruling accepted, with two card-level carry-overs

The ruling is recorded in three places (the `session-headers.ts` doc comment, the PR body, and a dedicated `describe` in `photo-search-client.test.ts` asserting both directions — full header set present with a credential saved, null with none). **I agree with it.** It is coherent rather than convenient: Task 5's vision probe/badge exists specifically so a BYOK user's *image* turns run on their own key, and T9 forbids a BYOK turn from silently falling back to the platform key — so suppressing the headers on photo search would have created the very fallback the threat model bans. Deciding it explicitly, in a comment plus a test, is the right resolution of my P1-1.

Two things the ruling implies that are **not** this PR's to fix but must not be lost:

- **P3-3**: `/v1/photo-search` is in neither `AUTH_RATE_LIMITED_EXACT` nor the `/v1/byok/` prefix in `worker/app.ts`, and Tasks 3/4's container-side consumption is scoped to `/v1/chat` plus the probe. So until those extend, a credential is transmitted to an endpoint that cannot use it and that T9's limiter does not cover. Add a line to the Task 3/4 and Task 9 cards.
- **P3-4**: Task 2's inbound header-allowlist stripping must cover the photo-search route, not just `/v1/chat` — otherwise the ruling widens the T1 sink surface. Worth one sentence in the Task 2 card.

---

### Standards re-check

`i18n.ts` 276 lines (under the 300 cap — the rebase brought #445's `error-states-i18n.ts` split), `use-chat-session.test.tsx` back to 189, `session-headers.test.ts` 71, `photo-search-client.test.ts` 160, `byokStorage.ts` 213. **P3-5**: `byokStorage.test.ts` is 202 lines, 2 over CLAUDE.md's 200-line test-file cap — the validation cases already moved to their own file, so trimming two lines or moving the vision-lockstep describe finishes the job.

### Final AC ledger

**6 of 9 genuinely done** (AC1 grep half, AC2, AC3, AC4 storage half, AC5, AC8), **3 correctly deferred** to the UI wave (AC6, AC7, AC9 — all require the component or CSS). The two overclaims I flagged (AC1, AC3) are now real. Scope division is honest.

Approve. Carry P2-1 (model and `baseUrl` charset validation) into the UI-wave card as a required item, and the five P3s wherever they belong.

Issue #446, opened 2026-07-28
photo-search follow-ups from #445 review: confirm-signal telemetry mismatch + missing transport headers


lifeodyssey:
From the Fable-side dual review of PR #445 (S1.3, #260). Approve verdict; these are follow-ups deliberately kept out of that PR because the correct fix needs a contract change.

> **2026-07-28 update:** several items below were pulled back into PR #445 during the second-seat (Opus) review round and are now fixed there — struck through, with the fixing commit noted. Remaining open: P2-1 (contract change), quota pruning/shared storage, gemini httpx client reuse, clarify-card `data-state`.

## P2-1 — confirm-signal telemetry mismatch (TRUE bug, data integrity)
`apps/web/src/features/chat/components/PhotoSearchUpload.tsx:67` — `confirmSignals` hardcodes `query_type='anime_screenshot'` and derives `layer_hit` from intent only ('1'/'none'), while backend `photo_search.py` computes `real_world_photo` (no titles), layer '2' (GPS nearby merge), and layer '1' for ambiguous-resolve clarify. Every `user_confirmed=true` event is systematically mis-bucketed — corrupting exactly the SD-26 layer×confirm funnel that gates DD-11/layer-3. `photo-search-upload.test.tsx:96` currently enshrines the wrong values.
**Fix direction:** echo the backend-computed `PhotoSearchSignals` (or an opaque token) in the response envelope; client replays it verbatim in the confirm ping. Needs `packages/contract` change. Test: integration.

## P2-2 — photo-search transport sends only Content-Type
~~`apps/web/src/features/chat/photo-search.ts:65` — no `x-locale` (backend `_locale` falls back to 'ja' for zh/en users), no auth token / `x-session-id` (signed-in users metered as anon; member quota tier and BYOK route unreachable; anon quota key degrades to `request.client.host`, which behind the Worker may collapse into one shared edge-IP bucket). Latent while quotas default to None — **must land before ops sets quota values**. Fix: pass locale + session/auth headers like the chat transport does. Test: unit.~~
**FIXED in #445** (`ed104c86` client headers via shared `session-headers.ts` + `x-locale`; `4404be9c` quota key switched to the edge-asserted `X-User-Id`, never `x-session-id`).

## P3s
[...]

PR #472, opened 2026-07-28
feat(agent): per-identity anonymous daily message quota (#282 backend)

lifeodyssey:
## Summary

Backend half of #282 (S1.10): container-ingress **per-identity daily message quota** for anonymous visitors — a fairness/UX mechanism, not a defense line. Complements the existing global dollar breaker (#274/#438).

- New durable counter (`anon_daily_message_count`, Atlas migration, PK `(usage_date, anon_id)`) — a single atomic UPSERT...RETURNING increments and returns today's count for one `anon_id`. DB-backed rather than in-process, per #446's lesson that in-process counters don't survive multi-instance deploys.
- `chat.py` gates anonymous turns: the global budget check (D11) runs first — it's the more severe, systemic concern and wins ties — then falls through to the per-identity quota (D12) only if the budget allowed the turn. Logged-in traffic is never touched by either gate. The quota check runs *before* request-body validation, so a malformed request that goes on to 422 still counts as an attempt (same reasoning a rate limiter uses).
- 403 payload: `{"error": {"code": "anon_quota_exhausted", "message": ..., "action": "login", "data": {"quota_resets_at": "<next UTC midnight, ISO-8601 Z>"}}}`, matching the frontend's (#468, now merged) D12 wire contract.
- Container re-validates the `anon_id` itself against `^anon_[0-9a-f]{32}$` (mirrors #460's `_ANON_ID_PATTERN` precedent, now also merged) — a malformed identity is neither counted nor rejected.
- Retention: `AnonQuotaRepository.purge_older_than()`, a `--dry-run` CLI script, a standalone nightly GitHub workflow, and `Settings.anon_daily_message_count_retention_days` (default 30).
- `wrangler.toml`'s three env blocks all forward `ANON_DAILY_MESSAGE_QUOTA` (default 20); `CONTAINER_ENV_KEYS`/`buildContainerEnvVars` were extracted out of `entry.ts` into a new `worker/containerEnv.ts` so the forwarding is unit-testable under plain `node --test` (`entry.ts` imports `@cloudflare/containers`, whose ESM build only resolves under workerd).
- `packages/contract/src/error-registry.ts`: the wire codes, `AnonQuotaExhaustedData`, `AnonLimitErrorEnvelope`, and `readQuotaResetsAt()` (the one blessed extraction site for `error.data.quota_resets_at`) all live in the file #468 established as the single source — folded here rather than in a competing standalone module. Python pins its own string constants against this file's text (it can't import TypeScript directly), the same three-mirror pattern used for the catalog error registry. `worker/costBreaker.ts` deliberately keeps its own hand-mirror of the budget code rather than importing from the contract — `packages/contract/AGENTS.md`'s documented pitfall says Workers must not pull Zod into their bundle for a single string constant.

## Design notes

- **Storage**: a new table, not a reuse of `daily_usage` — that table's PK `(usage_date, scope)` is a global aggregate across the whole anonymous population; this feature needs `(usage_date, anon_id)` granularity, a different shape entirely. Mirrors the existing `UsageRepo`/`UsageRepository` port+repo pattern (`AnonQuotaRepo` protocol, `AnonQuotaRepository`, wired onto `SupabaseClient.anon_quota`). No dead `DEFAULT 0` (the only writer always supplies an explicit value) and no redundant secondary index (the PK already leads with `usage_date`).
- **Reset semantics**: UTC calendar day, consistent with `daily_usage`'s existing `utc_today()` convention.
- **0/None semantics**: both disable the quota — the same convention as the budget breaker's `0` (an earlier draft had `0` mean "reject every message", inverting that convention).
- **Fail-open**: a read/write error on the counter never blocks the anonymous surface (matches the budget breaker's own contract).
- **Malformed-body semantics** (decided, tested): the quota counter increments before body validation, so a request that goes on to 422 still spends one message of the visitor's allowance.

## AC coverage

| AC | Type | Test |
|---|---|---|
| Anonymous identity within quota sends normally | endpoint test | `test_chat_anon_quota.py::test_the_nth_message_within_quota_still_passes` |
| A brand-new anonymous identity starts at full quota, not zero | unit | `test_anon_quota.py`, `test_chat_anon_quota.py` (both `..._starts_at_full_quota_not_zero`) |
| Exceeding quota → 403 `anon_quota_exhausted` + `action: login` + `error.data.quota_resets_at` | unit (endpoint) | `test_chat_anon_quota.py::test_the_n_plus_first_message_is_rejected`, `::test_the_rejection_carries_the_wire_contract...`, `::test_the_rejection_carries_an_iso_utc_reset_instant_under_data` |
| i18n copy | owned by #468 (merged) | n/a — backend only emits code/action/timestamp |
| Logged-in users unaffected | unit | `test_chat_anon_quota.py::test_logged_in_users_are_never_quota_checked` |
| Cross-day reset (mocked clock) | unit | `test_anon_quota.py::test_crossing_a_utc_day_boundary_resets_the_count` |
| Budget vs. quota precedence | unit | `test_chat_budget_quota_precedence.py` (3 tests) |
| Malformed anon_id neither counted nor rejected | unit | `test_anon_quota.py` (3 tests, including the `.fullmatch`→`.match` mutation-killing trailing-newline case) |
| Malformed body still consumes a message | unit | `test_chat_anon_quota.py::test_a_malformed_body_still_consumes_one_message_of_the_quota` |
| Retention purges stale rows, keeps fresh ones | integration (real Postgres) | `test_anon_quota_repo_contract.py::test_purge_older_than_removes_stale_rows_and_keeps_recent_ones` |
| GRANT actually took | integration (real Postgres) | `test_anon_quota_repo_contract.py::test_agent_svc_holds_the_grants_the_repo_needs` |
| Contract wire-code + payload-field pin | unit + vitest | `test_anon_limits_contract_pin.py`, `packages/contract/test/anon-limits.test.ts` |
| `ANON_DAILY_MESSAGE_QUOTA` actually reaches the container | vitest (worker) | `worker/entry.test.ts` (2 tests against `containerEnv.ts`) |

## Real-SQL integration evidence (issue: unit tests only mocked the pool)

`test_anon_quota_repo_contract.py` runs the actual UPSERT against Postgres via a pinned Atlas 0.30.0 binary (downloaded + checksum-verified locally since the global install was a mismatched canary build):
- Same-key increments across calls, per-identity/per-date isolation.
- **20 concurrent `asyncio.gather` callers on the same key return exactly `{1..20}`** — no lost update, the one correctness dimension a mocked-pool unit test structurally cannot prove.
- Retention purge removes a stale row and keeps a fresh one.
- `has_table_privilege` proves the migration's GRANT actually took (not just that the SQL string mentions one).

All 6 pass.

## Mutation testing evidence

No mutation-testing tool is wired into this repo, so verification was manual: deliberate one-line bugs were introduced and reverted, each confirmed to fail exactly the test(s) meant to catch it:
1. `count > quota` → `count >= quota` (off-by-one) — caught.
2. Swapped budget/quota check order in `chat.py` — caught.
3. `ON CONFLICT ... DO UPDATE` → `DO NOTHING` in the repo's SQL — caught.
4. `timedelta(days=1)` → `timedelta(days=2)` in `next_utc_midnight` — caught.
5. `QUOTA_RESETS_AT_FIELD` → a hardcoded typo — caught.
6. Nested `data` wrapper removed from the 403 payload — caught.
7. Purge's `usage_date < $1` → `<= $1` (off-by-one) — caught **only by the real-Postgres integration test**, not the mocked unit test.
8. `_ANON_ID_PATTERN.fullmatch` → `.match` — caught by a trailing-newline test case (verified empirically first: plain trailing garbage like `"_EXTRA"` does *not* actually distinguish `.match`/`.fullmatch` for a `$`-anchored pattern; a bare trailing newline is the one input that does).

(One additional mutation — removing the `auth.user_id is None` guard in `_quota_rejection` — was *not* caught by any test. Confirmed as expected: `_require_trusted_user`'s FastAPI dependency already 400s before the route body runs if `user_id` is `None`, so that branch is unreachable defensive code needed only for mypy's `str | None` narrowing.)

## Gates

- `make lint` / `make typecheck` — clean.
- `make test` — 1608 passed, 89.20% coverage (floor 87%).
- `make test-integration`'s anon-quota suite — 6 passed against real Postgres via pinned Atlas 0.30.0.
- `pnpm run test:worker` — 185 passed.
- `packages/contract`: `pnpm typecheck` clean, `pnpm test` — 60 passed.
- `apps/web`: `pnpm typecheck` clean, `pnpm test` — 1319 passed (full suite, unaffected by this backend-only PR).
- `actionlint` on the new workflow — clean.

## Notes for the reviewing seats

- This branch was rebased twice onto `feat/frontend-rebuild` as it advanced during review (#468's frontend D12 half and #460's anon-identity/retention precedent both landed mid-review). The five WIP review-round commits were squashed into one clean commit before each rebase to keep conflict resolution to a single pass.
- Reconciled with #468's actual merged shape: the wire codes + `AnonQuotaExhaustedData` + `AnonLimitErrorEnvelope` + `readQuotaResetsAt()` now all live in `packages/contract/src/error-registry.ts` (the file #468 established), not a separate `anon-limits.ts` module I'd originally created before #468 merged.
- `apps/web`'s D12 consumer (`use-chat-session.ts`, from #468) already hand-rolls the `error.data ?? error` fallback read for `quota_resets_at` — it predates `readQuotaResetsAt()`. A follow-up could swap it to use the shared helper, but it isn't required: the hand-rolled version is already correct and tested.
- Spec deviation from the issue's original `Files changed` plan (Worker KV counter, `worker/quota.ts`) → actual container-ingress Postgres implementation: commented on issue #282 directly with the rationale (atomic increment-and-compare under concurrency, single authoritative decision point).
- Photo-search quota (`photo_search_quota_anon`) is a separate, pre-existing feature and unaffected by this change — confirmed by inspection, no shared code path.

Refs #282

### DIFF
--- a/.github/workflows/purge-anon-quota-counts.yml
+++ b/.github/workflows/purge-anon-quota-counts.yml
@@ -1,0 +1,46 @@
+# Standalone top-level workflow (see .claude/rules/ci.md: codeql.yml / dependabot-agent.yml
+# precedent, mirrored from agent-eval-nightly.yml) — kept separate from ci.yml so this cron
+# never rides along with every push/PR trigger in the main CI matrix.
+name: Purge anon_daily_message_count (retention)
+
+# Issue #282 review: anon_daily_message_count is a pure per-day aggregate counter with no
+# dependent rows, so its retention policy is a nightly age-based DELETE, not a per-row purge
+# with an FK backstop (contrast the anonymous-session purge precedent).
+on:
+  schedule:
+    - cron: "37 19 * * *" # 19:37 UTC daily — off-hour, distinct minute from the eval nightly
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "Report eligible rows without deleting"
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+concurrency:
+  group: purge-anon-quota-counts
+  cancel-in-progress: false
+
+jobs:
+  purge:
+    name: Purge stale per-identity quota counters
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: apps/agent
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/setup
+        with:
+          python: "true"
+
+      - name: Purge rows past the retention window
+        run: |
+          uv run --frozen python -m agent.scripts.purge_anon_quota_counts \
+            ${{ inputs.dry_run == true && '--dry-run' || '' }}
+        env:
+          SUPABASE_DB_URL: ${{ secrets.NEON_DATABASE_URL }}
--- a/apps/agent/agent/config/settings.py
+++ b/apps/agent/agent/config/settings.py
@@ -146,6 +146,25 @@ class Settings(BaseSettings):
         ge=0,
         description="Global anonymous daily spend ceiling in USD (0 disables)",
     )
+    # Per-identity anonymous daily message quota (issue #282, S1.10) — a
+    # fairness/UX mechanism, not a security defense line: it keeps one
+    # visitor's free usage reasonable while ANON_DAILY_COST_BUDGET_USD stays
+    # open for everyone else. Distinct from that global dollar breaker: this
+    # ceiling is compared against one anon identity's own message count for
+    # today. None OR 0 disables the quota entirely — the same "0 disables"
+    # convention as the budget breaker.
+    anon_daily_message_quota: int | None = Field(
+        default=None,
+        ge=0,
+        description="Per-identity daily message cap for anonymous users (None or 0 disables)",
+    )
+    # Retention for anon_daily_message_count (issue #282 review): rows older
+    # than this many UTC days are eligible for the standalone purge script
+    # (scripts/purge_anon_quota_counts.py). 30 mirrors the analogous
+    # anonymous-session retention window used elsewhere in this codebase.
+    anon_daily_message_count_retention_days: int = Field(
+        default=30, gt=0, description="Days to keep anon_daily_message_count rows"
+    )
     model_input_cost_per_mtok_usd: float = Field(
         default=0.0, ge=0, description="Input token price per million tokens (USD)"
     )
--- a/apps/agent/agent/domain/ports.py
+++ b/apps/agent/agent/domain/ports.py
@@ -147,6 +147,12 @@ class UsageRepo(Protocol):
     async def total_cost_usd(self, *, usage_date: date, scope: str) -> float: ...
 
 
+class AnonQuotaRepo(Protocol):
+    """Per-identity anonymous daily message counter (issue #282 / S1.10)."""
+
+    async def increment_and_count(self, *, usage_date: date, anon_id: str) -> int: ...
+
+
 def get_session_repo(db: object) -> SessionRepo | None:
     """Return the session repo if *db* exposes one with async upsert_session."""
     session = getattr(db, "session", None)
@@ -195,3 +201,13 @@ def get_usage_repo(db: object) -> UsageRepo | None:
     if not asyncio.iscoroutinefunction(getattr(usage, "accumulate_usage", None)):
         return None
     return cast(UsageRepo, usage)
+
+
+def get_anon_quota_repo(db: object) -> AnonQuotaRepo | None:
+    """Return the anon-quota repo if *db* exposes one with async increment."""
+    repo = getattr(db, "anon_quota", None)
+    if repo is None:
+        return None
+    if not asyncio.iscoroutinefunction(getattr(repo, "increment_and_count", None)):
+        return None
+    return cast(AnonQuotaRepo, repo)
--- a/apps/agent/agent/infrastructure/supabase/client.py
+++ b/apps/agent/agent/infrastructure/supabase/client.py
@@ -13,6 +13,7 @@ import asyncpg
 import structlog
 
 from agent.infrastructure.supabase.client_types import AsyncPGPool, Row
+from agent.infrastructure.supabase.repositories.anon_quota import AnonQuotaRepository
 from agent.infrastructure.supabase.repositories.bangumi import BangumiRepository
 from agent.infrastructure.supabase.repositories.feedback import FeedbackRepository
 from agent.infrastructure.supabase.repositories.messages import MessagesRepository
@@ -31,7 +32,7 @@ class SupabaseClient:
 
     Access repositories via explicit typed properties:
     ``db.bangumi``, ``db.points``, ``db.session``, ``db.feedback``,
-    ``db.routes``, ``db.messages``, ``db.usage``.
+    ``db.routes``, ``db.messages``, ``db.usage``, ``db.anon_quota``.
     """
 
     def __init__(
@@ -54,6 +55,7 @@ class SupabaseClient:
         self._routes: RoutesRepository | None = None
         self._messages: MessagesRepository | None = None
         self._usage: UsageRepository | None = None
+        self._anon_quota: AnonQuotaRepository | None = None
 
     async def connect(self) -> None:
         """Create the connection pool and initialise repositories."""
@@ -102,6 +104,7 @@ class SupabaseClient:
         self._routes = RoutesRepository(pool)
         self._messages = MessagesRepository(pool)
         self._usage = UsageRepository(pool)
+        self._anon_quota = AnonQuotaRepository(pool)
 
     @property
     def bangumi(self) -> BangumiRepository:
@@ -156,3 +159,11 @@ class SupabaseClient:
         if self._usage is None:
             raise RuntimeError("UsageRepository not initialized — call connect() first")
         return self._usage
+
+    @property
+    def anon_quota(self) -> AnonQuotaRepository:
+        if self._anon_quota is None:
+            raise RuntimeError(
+                "AnonQuotaRepository not initialized — call connect() first"
+            )
+        return self._anon_quota
--- a/apps/agent/agent/infrastructure/supabase/repositories/anon_quota.py
+++ b/apps/agent/agent/infrastructure/supabase/repositories/anon_quota.py
@@ -1,0 +1,49 @@
+"""Per-identity anonymous daily message counter (issue #282 / S1.10)."""
+
+from __future__ import annotations
+
+from datetime import date
+
+from agent.infrastructure.supabase.client_types import AsyncPGPool
+
+
+class AnonQuotaRepository:
+    """Atomic per-(day, anon identity) message-attempt counter."""
+
+    def __init__(self, pool: AsyncPGPool) -> None:
+        self._pool = pool
+
+    async def increment_and_count(self, *, usage_date: date, anon_id: str) -> int:
+        """Atomically bump today's count for *anon_id* and return the new total."""
+        row = await self._pool.fetchrow(
+            """
+            INSERT INTO anon_daily_message_count (usage_date, anon_id, message_count)
+            VALUES ($1, $2, 1)
+            ON CONFLICT (usage_date, anon_id) DO UPDATE SET
+                message_count = anon_daily_message_count.message_count + 1,
+                updated_at = NOW()
+            RETURNING message_count
+            """,
+            usage_date,
+            anon_id,
+        )
+        count = 0 if row is None else row["message_count"]
+        return int(count)
+
+    async def purge_older_than(self, cutoff: date) -> int:
+        """Delete rows strictly older than *cutoff*; returns the count removed.
+
+        No FK backstop is needed here (unlike the session-purge precedent):
+        this table is a pure aggregate counter with no dependent rows, so a
+        plain age-based delete is the whole retention policy.
+        """
+        result = await self._pool.execute(
+            "DELETE FROM anon_daily_message_count WHERE usage_date < $1", cutoff
+        )
+        return _parse_delete_count(result)
+
+
+def _parse_delete_count(result: str) -> int:
+    """asyncpg's `execute` returns a command tag like ``"DELETE 3"``."""
+    parts = result.split()
+    return int(parts[-1]) if len(parts) == 2 and parts[0] == "DELETE" else 0
--- a/apps/agent/agent/interfaces/anon_quota.py
+++ b/apps/agent/agent/interfaces/anon_quota.py
@@ -1,0 +1,99 @@
+"""Per-identity anonymous daily message quota (issue #282, S1.10).
+
+Distinct from the global dollar breaker in ``usage_metering.py`` (X4, #274):
+that one asks "has the *whole* anonymous surface spent its budget for today?";
+this one asks "has *this one* anon identity spent its own message allowance
+for today?". Both read/write durable Postgres state so the check survives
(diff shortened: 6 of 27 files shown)


lifeodyssey (commented):
## Reviewer (Fable) — independent review of PR #472

**Verdict: request_changes** (1×P1; code quality otherwise high — all gates independently reproduced green, 2/5 self-reported mutations replayed and confirmed caught)

### Independent verification
- `make lint` clean, `make typecheck` clean (129 files), `make test` **1521 passed, coverage 89.16%** (floor 87) — reproduced in a fresh detached worktree at d80c889b.
- `packages/contract`: typecheck clean, 54 vitest passed.
- Mutation replay (2 of the self-reported 5, independently applied + reverted):
  - `count > quota` → `count >= quota`: killed by `test_the_nth_message_within_quota_passes` ✓
  - budget/quota order swap in `chat.py`: killed by `test_both_exhausted_yields_the_budget_code_not_the_quota_code` ✓
- "Brand-new identity starts at full quota" tests are genuine: the double increments from 0 so first call returns count=1, and the endpoint test drives a real 200 through the route with `next_count=1`. Not tautological.
- Race analysis: the single-statement `INSERT ... ON CONFLICT DO UPDATE ... RETURNING` is atomic; concurrent same-identity requests serialize on the row lock, so N+1 and N+2 arriving together get distinct counts and both correctly reject — no over-issuance window. Budget-rejected turns never touch the counter (tested). `resets_at` and the counter row key derive from the same `resolved_today`, so the UTC day boundary and `quota_resets_at` cannot disagree within a request.
- Eval suite: skipped — diff touches no agent behavior/prompts/tools (CI L0 smoke green anyway).
- All 19 CI checks green, including DB Migrations dry-run and Python integration (Neon), so the migration applies cleanly against a real Postgres.

### Findings

**P1 — deployment wiring missing: the quota can never be enabled in any deployed environment**
`worker/entry.ts` `CONTAINER_ENV_KEYS` does not include `ANON_DAILY_MESSAGE_QUOTA`, and no `wrangler.toml` `[vars]` entry exists for it (the budget got both: `ANON_DAILY_COST_BUDGET_USD` is in the forward list at worker/entry.ts:18 and in all three env blocks). Even if ops sets the var, the worker silently drops it before the container sees it, so `Settings.anon_daily_message_quota` stays `None` (disabled) in prod/preview forever. Note preview/dev envs run `ANON_ACCESS_ENABLED = "true"` with no per-identity cap. Fix: add the key to `CONTAINER_ENV_KEYS` + wrangler `[vars]` in this PR, or file an explicit deploy-wiring follow-up card and reference it here — otherwise Tester validation on a deployed env will silently exercise a disabled gate. (file: worker/entry.ts:18, wrangler.toml)

**P2 — quota units burn on requests that never become messages**
`_quota_rejection` runs before `_decode_body`, `_runtime_request` (422 validation) and `validate_session_owner` in `handle_chat` (chat.py). A malformed body, an over-length message, or a session-ownership rejection still consumes one unit of a small daily allowance — a buggy client can lock a visitor out for the day without a single message being processed. The docstring documents counting quota-rejected retries, but not 422-bound attempts. Recommend moving the quota gate after body validation (still before `runtime_api.handle`), or explicitly documenting/testing attempt-semantics including 422s. Also note the accounting deliberately differs from #438 metering (pre-run attempts vs post-run actual spend) — fine, but worth one sentence in `anon_quota.py`'s module docstring. (file: apps/agent/agent/interfaces/routes/chat.py:229)

**P2 — `quota=0` semantics diverge from the budget convention the comment claims to mirror**
`Field(ge=0)` admits `0`, which blocks ALL anonymous messages (`count=1 > 0`), while `anon_daily_cost_budget_usd`'s `0` disables the breaker. The settings comment says "None disables … the same safe default as the budget breaker's 0", which invites an operator to set `0` expecting disable and instead hard-block the anon surface. Either `ge=1`, or add a test + comment stating `0` is a deliberate kill switch. No test covers `quota=0` today. (file: apps/agent/agent/config/settings.py:150)

**P2 — the UPSERT SQL is never executed by any test**
`test_anon_quota_repo.py` string-asserts SQL fragments against an `AsyncMock`; the Neon integration job applies the migration but nothing runs `increment_and_count` against a real database, so a SQL typo or wrong RETURNING semantics would pass every gate. Precedent (`daily_usage`) shares this gap, but this counter is the enforcement primitive. Suggest a small integration contract test (increment twice → 2; `asyncio.gather` of N concurrent increments → N distinct values) plus an `agent_svc` grant assertion in `test_service_roles.py` (the new `GRANT` is untested). (file: apps/agent/agent/tests/unit/repositories/test_anon_quota_repo.py)

**P2 — no retention story for `anon_daily_message_count`**
One row per (UTC day, anon identity), unbounded over time; the `usage_date` index exists but nothing uses it. Needs a follow-up: periodic purge (e.g. `usage_date < today - 7`) or fold into the #460 cleanup. File the issue before merge so it doesn't evaporate. (file: db/migrations/20260729000001_anon_daily_message_quota.sql)

**P3 — fail-open on `Exception` for an anti-abuse gate**
Consistent with `_METER_ERRORS` and the #451/#463 fail-open precedent, and the asyncpg-derives-from-Exception rationale is real. Accepted — but sustained DB failure silently means unlimited anon messages with only a `warning` log. Suggest an ops follow-up: alert on `anon_daily_message_count_failed` frequency.

**P3 — minor**
- `message_count BIGINT DEFAULT 0` is dead — the INSERT always supplies `1`. Cosmetic.
- Contract pin `f"{QUOTA_RESETS_AT_FIELD}:" in contract_source` would also match a comment; harmless given the three-mirror precedent, but an exact `quota_resets_at: z.` pin would be tighter.
- #468 alignment verified: no file overlap in `packages/contract` (no merge conflict either order); its classifier reads only `code`+`status`, so the missing `quota_resets_at` in its MSW fixtures is compatible today — but if the frontend later parses via `AnonQuotaExhaustedData` (zod requires the field), those fixtures must add it. Fold into the already-noted post-merge rebase.

### Checklists
- SOLID: clean — `AnonQuotaRepo` port + repo mirrors `UsageRepo` exactly (OCP: new gate added without touching budget logic); no violations.
- 1-10-50 / naming / test smells: pass. Test files 142/138/109/61/55 lines, ≤5 mocks, no timing or conditional logic, verb-first behavioral names.
- SQL: parameterized ($1/$2), PK-indexed upsert, no SELECT *.
- TDD/Quality Ratchet: ac_total=7 backend ACs (i18n copy owned by #468), ac_with_test=7 — every AC row maps to a test present in the diff, verified by reading. missing: none.

```json
{"verdict":"request_changes","eval_results":{"test_eval":"skipped"},"quality_ratchet":{"ac_total":7,"ac_with_test":7,"missing":[]}}
```

lifeodyssey (commented):
## Reviewer (Fable) — final-round verdict on PR #472 (979f5ed7 + 0a41e305)

**Verdict: approve**

### Independent re-verification (fresh detached worktree at 0a41e305)
- `make lint` / `make typecheck` clean; `make test` **1530 passed, coverage 89.07%** (floor 87).
- `pnpm run test:worker` **178/178**; `packages/contract` typecheck clean + **56/56** vitest.
- Mutation replay on the new 0-semantics: reverting `if not quota` to `if quota is None` is killed by `test_a_zero_quota_also_disables_the_check_same_as_none`. Confirmed, reverted, worktree clean.
- CI on head: all green including Python integration (Neon) — the new real-SQL suite (incl. the 20-way concurrent no-lost-update test) ran against real Postgres in CI. SonarCloud red is discussed below.

### Round-1 findings — resolution audit
| Finding | Status |
|---|---|
| P1 env wiring | **Fixed properly.** `worker/containerEnv.ts` extracted with `ANON_DAILY_MESSAGE_QUOTA`, two `node --test` cases prove forwarding (and absence when unset), `wrangler.toml` carries the var in all three env blocks, `test_deploy_model_env_consistency.py` repointed at the new module, docs updated. The extraction rationale (entry.ts's `@cloudflare/containers` import chain breaking under Node) is sound. |
| P2 quota=0 divergence | **Fixed.** `None` or `0` disables, matching the budget convention; unit test added and mutation-verified by me. |
| P2 UPSERT never executed | **Fixed.** `test_anon_quota_repo_contract.py`: real increment/cross-identity/cross-date, concurrency with exact `{1..20}` set assertion (lost-update proof), and the purge boundary — green in CI's Neon lane. |
| P2 retention | **Fixed.** `purge_older_than` + `--dry-run` CLI + standalone nightly workflow (pinned checkout, `persist-credentials: false`, `contents: read`, existing `NEON_DATABASE_URL` secret) + `anon_daily_message_count_retention_days` setting + unit tests + integration boundary test. |
| P2 422-before-count | **NOT addressed — see below.** |
| P3 dead DEFAULT / pin anchor | Fixed (migration cleaned incl. dropping the redundant `usage_date` index — correct call, the PK leads with it; pin hardened to `quota_resets_at: z.`). |
| error.data nesting | Fixed end-to-end: `chat.py` nests under `error.data`, endpoint tests assert both the nested field and the absence of a top-level copy, contract adds `AnonLimitErrorEnvelope` with `data` optional (budget omits it). I verified #468's head reads `error.data.quota_resets_at` (with a flat fallback) — the shapes have converged. |
| Bonus | Container-side `anon_id` re-validation (`^anon_[0-9a-f]{32}$`, uncounted+unrejected fail-open) with both malformed-shape tests — good defense-in-depth per the #460 precedent. |

### Adjudication requested: `costBreaker.ts` hand-mirror
**ACCEPTED.** `packages/contract/AGENTS.md` explicitly documents type-only hand-mirrors so Zod stays out of Worker bundles ("Do not introduce runtime value imports into the catalog's type mirror"), and `anon-limits.test.ts` pins the mirror in lockstep by reading `costBreaker.ts` as text — the same three-mirror pattern as the catalog error registry. Importing the constant would pull `zod` (and transitively the contract package) into the edge worker's production bundle for one string. My round-1 concern is adequately answered; keep as-is.

### Remaining non-blocking items (record, don't hold the merge)
1. **422-before-count is unresolved and no pin test exists.** `_quota_rejection` still runs before `_decode_body` / `validate_session_owner`, so malformed bodies and ownership rejections burn a quota unit, and I could find no test or docstring covering the 422 case anywhere in the branch (the handoff note claimed a pinned test; the code does not contain one). Downgrading to a follow-up rather than a blocker because: the shipped default is a generous 20/day, the semantics are attempt-counting by design for quota-rejected retries, and the new anon-id validation removes one junk-counting class. Please file a follow-up: either move the quota gate after body validation, or pin the deliberate 422-burning semantics with a test + docstring sentence.
2. **SonarCloud quality gate is red** — a single MAJOR `githubactions:S8541` on the new workflow's `uv run` step (no `--no-build`). The identical pattern exists in `ci.yml` and `agent-eval-nightly.yml`; Sonar only rates new code. actionlint/zizmor both pass. Suggest resolving the Sonar finding as "safe with precedent" (or adding `--no-build` if the purge script tolerates it) so the gate goes green — but it is not merge-blocking (no branch protection requires it).
3. Minor: `_ANON_ID_PATTERN` is a re-declaration of the same regex in `interfaces/routes/_deps.py` (acknowledged in the comment as a mirror). A shared constant would be tighter; fine as-is.
4. Correction for the record: #468 is still **OPEN**, not merged, as of this review. Immaterial here — no file overlap, and its `error.data.quota_resets_at` read path already matches this PR's envelope in either merge order.

```json
{"verdict":"approve","eval_results":{"test_eval":"skipped"},"quality_ratchet":{"ac_total":7,"ac_with_test":7,"missing":[]}}
```

PR #473, opened 2026-07-28
feat(#273 T4): typed session fact ledger (pacing + scene refs)

lifeodyssey:
## Summary

Implements S1.7 (#273) **Task 4** — the typed session-memory fact ledger, per the OQ-3(c) 5→2 field-reduction ruling in the approved spec (`docs/superpowers/specs/2026-07-28-s1.7-living-document-save-login-wall-design.md`).

- **Two fields only**, per the ruling: **user hard constraint** (route pacing) and **episode/scene references** (from explicitly selected points). The other three candidates from the card's prose already live in `SessionState` (`current_anime` / `search_results` / `routes`) and are not duplicated.
- **Hard consumption gate honored**: each field reaches a named prompt-injection consumption point in `trusted_session_context()` (`agents/animichi_agent.py::_fact_ledger_context`) — a field with no consumer would be cut, per the explicit anti-`user_memory`-dead-scaffold guard.
- **Deterministic recorder, zero LLM calls** (OQ-4): `record_turn_facts()` derives records solely from `AgentResult.steps` — `plan_route`'s `pacing` argument for the hard constraint, `plan_selected`'s `ordered_points` payload (episode/name/time_seconds) for scene references. Wired into `session_facade.extract_context_delta()`, invoked exactly once per turn before persistence.
- **Append/supersede, never overwrite**: a correction appends a new record and tags the prior one `superseded_by`; prior content stays readable.
- **Bounded**: 8 records/field, 16 total (mirrors `SessionState.MAX_REFS`), oldest-superseded-evicts-first, encoded ledger stays under 8 KiB.
- **Strict boundary**: `extra="forbid"` on all ledger models, matching `_SessionModel`'s discipline.
- **Persistence**: `fact_ledger` is a field on `SessionState`, so it round-trips through the existing `sessions.state` JSONB envelope (`session_state_v2`) — no new table, no migration (OQ-2). An empty ledger adds no new top-level key to the envelope.

## Design notes (two fields / two consumers)

| Field | Source (steps) | Consumer |
|---|---|---|
| Hard constraint (pacing) | `plan_route` step, `params["pacing"]` on success | `User hard constraint: {value} pacing.` line in trusted context |
| Scene reference | `plan_selected` step, `data["ordered_points"][].{id,episode,name,time_seconds}` on success | `Referenced scene: {value}.` line(s) in trusted context |

## AC coverage (8/8, unit 7 + integration 1)

- [x] Happy path: independent append per field with timestamp + kind — unit
- [x] Correction appends + tags prior `superseded_by`, prior content intact — unit
- [x] Consumption gate: renders `trusted_session_context()` and asserts each field's live value appears — unit (mutation-verified, see below)
- [x] Recorder determinism: zero model calls, derives solely from `AgentResult.steps`, records nothing for ledger-irrelevant steps — unit
- [x] Null/empty: fresh ledger round-trips, adds no key to `sessions.state` — unit
- [x] Error path: unknown field / malformed record rejected via `extra="forbid"` — unit
- [x] Boundary: 16 total/8-per-field cap, oldest-superseded evicts first, live records survive, <8 KiB — unit
- [x] Round-trip: envelope persist/load preserves timestamps + supersession chain — integration

## Mutation verification (manual, per spec's mandatory instruction)

Confirmed each of the following passes trivially against a no-op and fails correctly against the real implementation, then restored:
- Deleting `_fact_ledger_context`'s call in `trusted_session_context` → both consumption-gate tests go red.
- No-op'ing `_record_step` → recorder-determinism tests go red.
- No-op'ing `_evict` → boundary-cap test goes red (12 records instead of capped 8).

## Test plan

- [x] `make lint` — clean
- [x] `make typecheck` — clean (128 source files, incl. new `agent/domain/fact_ledger.py`)
- [x] `make test` (`agent/tests/unit`) — 1504 passed, coverage 89.24% (floor 87%)
- [x] New tests: `test_fact_ledger.py`, `test_fact_ledger_recorder.py` (unit), `test_fact_ledger_persistence.py` (integration) — all green
- [ ] `make test-integration` full suite blocked locally by a pre-existing Atlas version mismatch (installed 1.2.4 vs pinned 0.30.0) unrelated to this change; the new integration test does not require the DB fixture and passes in isolation.

Not merging — per harness process, Reviewer/Tester gate this before merge.

### DIFF
--- a/apps/agent/agent/agents/animichi_agent.py
+++ b/apps/agent/agent/agents/animichi_agent.py
@@ -43,6 +43,7 @@ from agent.agents.runtime_models import (
 from agent.agents.session_state import SessionState
 from agent.agents.tool_outcomes import ResolveAmbiguous, ResolveNotFound
 from agent.agents.web_tools import TOOLS as WEB_TOOLS
+from agent.domain.fact_ledger import FactLedger
 from agent.infrastructure.observability import (
     record_agent_run_error,
     record_managed_prompt_resolution,
@@ -457,7 +458,19 @@ def trusted_session_context(deps: RuntimeDeps) -> str:
             f"Pending {pending.reason} revision {pending.revision}; "
             f"candidate_ids={pending.candidate_ids}."
         )
+    parts.extend(_fact_ledger_context(session.fact_ledger))
     return "[Trusted runtime context]\n" + "\n".join(parts)
+
+
+def _fact_ledger_context(ledger: FactLedger) -> list[str]:
+    """Named consumption point for both fact-ledger fields (OQ-3(c) gate)."""
+    parts: list[str] = []
+    constraint = ledger.active_hard_constraint()
+    if constraint is not None:
+        parts.append(f"User hard constraint: {constraint.value} pacing.")
+    for ref in ledger.active_scene_references():
+        parts.append(f"Referenced scene: {ref.value}.")
+    return parts
 
 
 _REPEAT_GUARD_HINT = (
--- a/apps/agent/agent/agents/animichi_agent.py
+++ b/apps/agent/agent/agents/animichi_agent.py
@@ -467,9 +467,17 @@ def _fact_ledger_context(ledger: FactLedger) -> list[str]:
     parts: list[str] = []
     constraint = ledger.active_hard_constraint()
     if constraint is not None:
-        parts.append(f"User hard constraint: {constraint.value} pacing.")
+        parts.append(
+            f"User hard constraint: {constraint.value} pacing. Apply this "
+            "pacing to every subsequent plan_route call unless the user "
+            "explicitly changes it."
+        )
     for ref in ledger.active_scene_references():
-        parts.append(f"Referenced scene: {ref.value}.")
+        parts.append(
+            f"Referenced scene: {ref.value}. The user explicitly selected "
+            "this; treat it as a durable point of interest for follow-up "
+            "questions this session."
+        )
     return parts
 
 
--- a/apps/agent/agent/agents/session_state.py
+++ b/apps/agent/agent/agents/session_state.py
@@ -6,6 +6,8 @@ from typing import Literal, NewType, Self, TypeAlias, TypeVar
 
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
+from agent.domain.fact_ledger import FactLedger
+
 MAX_REFS = 8
 
 ResultRef = NewType("ResultRef", str)
@@ -202,6 +204,7 @@ class SessionState(_SessionModel):
     geocode_staging: GeocodeStaging | None = None
     last_result_ref: ResultRef | None = None
     clarification_revision: int = 0
+    fact_ledger: FactLedger = Field(default_factory=FactLedger)
 
     @model_validator(mode="after")
     def _bound_restored_registries(self) -> Self:
--- a/apps/agent/agent/domain/fact_ledger.py
+++ b/apps/agent/agent/domain/fact_ledger.py
@@ -162,9 +162,14 @@ class FactLedger(_LedgerModel):
         return len(self.model_dump_json().encode("utf-8"))
 
 
-# A same-turn supersession doesn't yet have the successor's id when the whole
-# live set is retired up front; `_TURN_SUPERSEDED` marks "no longer live"
-# without inventing a fake id. Real ids replace it once eviction runs.
+# A whole-set turn replacement has no single "successor" id for every retired
+# record — a batch of N old points can be replaced by a batch of M new ones
+# with no 1:1 correspondence between any one retired record and any one new
+# one. `_TURN_SUPERSEDED` is therefore an intentional, permanent tombstone
+# value, not a placeholder: it is never rewritten to a real record id, and it
+# only ever means "retired by a later turn's whole-set replace" rather than
+# "specifically superseded by record X" (that stronger, id-linked chain is
+# what `append_hard_constraint`'s `_supersede` still provides).
 _TURN_SUPERSEDED = FactId("__turn_superseded__")
 
 
--- a/apps/agent/agent/domain/fact_ledger.py
+++ b/apps/agent/agent/domain/fact_ledger.py
@@ -7,24 +7,40 @@ derived solely from `AgentResult.steps` by `record_turn_facts` (OQ-4); there
 is no separate LLM extraction round. Each field's live value is surfaced by
 a named prompt-injection consumption point in `agents/animichi_agent.py`, per
 the hard consumption gate — a field with no consumer is dead scaffolding.
+
+Scene references are **turn-scoped**: each `plan_selected` step replaces the
+whole live scene-reference set with the current selection (unchecking a point
+retires it, not just accumulates a new one), which is also what keeps the
+ledger's growth bounded across many turns. The record cap and the encoded
+byte budget are both enforced here, in the write path, not only asserted by
+tests.
+
+Anonymous-session purge note: route-bearing anonymous sessions are exempt
+from the retention purge (#460) and can therefore live indefinitely; the
+per-field record cap and byte budget below are what keep such a
+long-lived session's ledger bounded regardless of session age.
 """
 
 from __future__ import annotations
 
+import re
 import uuid
 from collections.abc import Sequence
-from datetime import UTC, datetime
-from typing import Literal, NewType, Protocol, TypeVar
+from datetime import datetime
+from typing import Literal, NewType, Protocol, TypeVar, cast, get_args
 
 from pydantic import BaseModel, ConfigDict, Field
 
 MAX_RECORDS_PER_FIELD = 8
-_MAX_VALUE_LEN = 200
+MAX_LEDGER_BYTES = 8 * 1024
+_MAX_VALUE_BYTES = 96
+_MAX_ID_BYTES = 96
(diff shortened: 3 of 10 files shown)


lifeodyssey (commented):
## Fable seat — independent review of #473 (S1.7 Task 4: fact ledger)

Verdict: **request_changes** (3 P1s, all empirically reproduced at PR head 2c8c7aaa).

Gates run independently in a detached worktree: `make lint` / `make typecheck` / `make test` all green (1504 passed, coverage 89.24% >= 87 floor; the one wire-contract error was a missing-pnpm-install environment issue, not this PR). Mutation checks: (1) deleting `parts.extend(_fact_ledger_context(...))` -> constraint consumption test red; (2) deleting only the scene-ref injection lines -> scene consumption test red; (3) no-op `_supersede` -> correction test red; (4) unhooking `record_turn_facts` from `extract_context_delta` -> integration round-trip red. The consumption gate is real, prompt-content-level, and wired into the actual prompt (`animichi_runner.py:200` passes `trusted_session_context(deps)` as the first prompt part). The dead-scaffold defense — the soul of this card — genuinely holds for both fields. The P1s below are about the parts around it.

### P1-1 (tdd/spec): AC "adds no key to `sessions.state` for never-recorded sessions" is violated, and its test is a tautology
`_serialize_runtime_state` is a plain `state.model_dump(mode="json")`, so every session now persists `"fact_ledger": {"hard_constraints": [], "scene_references": []}` — verified: `_serialize_runtime_state(SessionState())` contains the key. The shipped test `test_fresh_empty_ledger_round_trips_and_adds_no_key_to_session_state` asserts `set(state.model_dump(mode="json")) == before_keys` where `before_keys` was computed from the *same post-PR model* — it can never fail and never touches the envelope. This is exactly the "passes against a no-op implementation" failure mode the spec's mutation checklist warns about.
- File: `apps/agent/agent/interfaces/session_facade.py:169` + `apps/agent/agent/tests/unit/test_fact_ledger.py:485`
- Fix: omit the key when `ledger.is_empty()` (e.g. exclude in serialization) and rewrite the test to build the envelope via `extract_context_delta` on a no-fact turn and assert `"fact_ledger" not in delta["session_state_v2"]`.
- Side benefit: never-recorded sessions stay loadable by pre-#473 code under `extra="forbid"` (today a rollback makes `_parse_runtime_state` swallow a ValidationError and silently drop the whole session context).

### P1-2 (framework/boundary): the 8-per-field / 16-total cap is not enforced for live records; the 8 KiB budget is breachable and unenforced at runtime
`_evict` only removes *superseded* records and returns when none exist. Scene references for distinct `point_id`s never supersede each other, so the list grows without bound across turns — verified: 20 distinct live points -> `len(scene_references) == 20` (3384 bytes; ~48 records crosses 8 KiB). A user who selects different point sets over several turns walks straight past the cap. The boundary test only exercises the same-point supersession path, so it passes while the cap is fictional in the distinct-point case; `encoded_size_bytes()`'s only consumer is that one test — the byte budget is asserted for a shape that can't overflow and enforced nowhere it can.
- File: `apps/agent/agent/domain/fact_ledger.py:201` (`_evict`)
- Fix: when no superseded record exists, evict the oldest live record (or supersede-then-evict FIFO), and add a test appending >8 *distinct* live scene refs asserting the cap and byte budget hold.

### P1-3 (framework/data): the recorder guards against the wrong missing-value representation — `-1` sentinels leak into the prompt
`_record_one_scene` skips only `episode is None`, but the real step shape (`build_route_payload` -> `PilgrimagePoint.model_dump`) never yields `None`: `episode: int = -1`, `time_seconds: int = -1` are contract sentinels (`catalog_client.py:81-82`). Verified with a real `PilgrimagePoint`: the ledger records `"Episode -1 — X @ -1s"`, which the consumption point then injects as `Referenced scene: Episode -1 — X @ -1s.` — trusted-context pollution for every point without episode metadata. The recorder tests use hand-built dicts that diverge from the production payload shape, so nothing catches it.
- File: `apps/agent/agent/domain/fact_ledger.py:257-273`
- Fix: treat `episode`/`time_seconds` values `< 0` (or `== -1`) as absent; add a recorder test that feeds `build_route_payload(...)` output (or `PilgrimagePoint.model_dump()`) rather than a synthetic dict.

### P2-1 (clean-code): side effect hidden in a query-shaped function
`extract_context_delta` now mutates `result.session_state.fact_ledger` before serializing. Single call site today (`public_api.py:463`) keeps "exactly once" true, but the guarantee lives in a docstring, not structure; a second caller silently double-records (same-value re-appends create spurious supersession chains). Suggest a named `record_and_serialize`-style seam or moving `record_turn_facts` to the caller so the write is visible at the call site.

### P3 (notes, no action required to merge)
- `record_turn_facts` reads `datetime.now(UTC)` internally instead of accepting `now=` like the append methods — inconsistent with the repo's mock-the-clock rule; recorder tests just avoid asserting time.
- `test_recorder_performs_zero_model_calls` patches only `httpx.AsyncClient.request`; it's closer to theater than proof, though the determinism AC is adequately carried by the no-relevant-steps test.
- Recording `pacing="normal"` (the schema default, when the model passes it explicitly) as a "user hard constraint" slightly overstates user intent; harmless with supersession but worth a thought.
- Concurrency with #460 on the same `sessions.state` row is the pre-existing read-modify-write envelope pattern; this PR adds no new hazard.
- Token cost of the injection is bounded and modest (1 constraint line + <=8 scene lines inside the existing `[Trusted runtime context]` block) — consistent with post-#349 prompt discipline.

### 1-10-50 / typing
All new functions <=10 lines, `FactLedger` class ~47 lines, file 212 < 300, no `Any` (`object` + `isinstance` narrowing throughout), mypy strict green. Naming clean (`is_empty`, `active_hard_constraint`, verb-first appends).

### Quality Ratchet
ac_total 8 / ac_with_test 6. Missing-or-false: AC-5 "adds no key" (tautological test, behavior violated — P1-1); AC-7 boundary cap (only the supersession path tested; cap does not hold for distinct live records — P1-2).

```json
{"verdict":"request_changes","eval_results":{"test_eval":"skipped (no prompt/tool behavior change beyond deterministic injection; unit-verified)"},"quality_ratchet":{"ac_total":8,"ac_with_test":6,"missing":["AC: empty ledger adds no key to sessions.state","AC: record cap holds (distinct live records case)"]}}
```

lifeodyssey (commented):
## Fable seat — final verification round, HEAD a41b4ac0

Verdict: **approve**. All three P1s independently re-verified as fixed; no new blockers.

### Re-verification (detached worktree, fresh at a41b4ac0)
- Gates: `make lint` / `make typecheck` green; `make test` **1517 passed**, coverage 89.10% (>= 87 floor); the PR's integration test passes.
- **Repro A (empty-key envelope)**: `_serialize_runtime_state(SessionState())` no longer contains `fact_ledger`; the tautological test is replaced by real envelope assertions (`test_fresh_empty_ledger_adds_no_key_to_the_serialized_envelope` + positive twin).
- **Repro B (unbounded live growth)**: 6 rounds x 8 distinct points driven through the *real* recorder path (`build_route_payload` -> `record_turn_facts`) -> 8 records total, 8 live, 1444 bytes, under budget. Turn-scoped replace + write-path `_bound` (count then bytes) makes growth structurally impossible.
- **Repro C (-1 sentinel)**: a real `PilgrimagePoint` with default `episode=-1/time_seconds=-1` records nothing (`is_empty()` true); recorder tests now consume the real producer shape instead of hand-rolled dicts.
- **Rollback tolerance**: valid session + malformed `fact_ledger` recovers with the rest of the typed state intact; `{"unknown": True}` still returns `None` — the tolerance is genuinely narrow (allowlist-only), not a blanket strip.

### Mutation spot-checks (all red as required)
1. Delete the turn-scoped retire loop (`record.superseded_by = _TURN_SUPERSEDED`) -> retire/recorder tests fail.
2. Revert the sentinel guard (`episode < 0` -> type-check only) -> sentinel test fails.
3. Delete the empty-ledger `dumped.pop("fact_ledger")` -> envelope test fails.
(Round-1 mutations on the consumption gate, supersede, and recorder wiring were already verified red at the prior HEAD; the same assertions are still present and stronger.)

### P3 notes, non-blocking
- The `_TURN_SUPERSEDED` comment says "Real ids replace it once eviction runs" — eviction pops records, it never rewrites ids; doc-only inaccuracy worth a follow-up touch.
- The scene-ref instruction sentence ("The user explicitly selected this; treat it as a durable point of interest...") repeats once per live ref (up to 8x); emitting it once for the block would save tokens.
- The unconditional trailing trim in `_evict_by_count` is an intentionally untestable backstop (retired records are always superseded today); fine as defensive code.
- True rollback to pre-#473 code still drops the typed state of sessions that *did* record facts — inherent; the empty-key fix confines the blast radius to fact-bearing sessions and the new narrow tolerance protects the forward direction.

### Quality Ratchet
ac_total 8 / ac_with_test 8 — the two previously missing/false ACs (empty-key, boundary cap) now carry genuine tests, plus extra coverage the spec didn't ask for (injection hygiene, mock-the-clock, narrow rollback tolerance, real-producer shapes).

```json
{"verdict":"approve","eval_results":{"test_eval":"skipped (deterministic injection only; unit + mutation verified)"},"quality_ratchet":{"ac_total":8,"ac_with_test":8,"missing":[]}}
```

Issue #465, opened 2026-07-28
flaky: web suite intermittent single failure (~1-in-6) under coverage load


lifeodyssey:
Observed during PR #445 round-2 review: first full `pnpm --filter web test` (with coverage) run gave 1 failed | 984 passed; five subsequent runs (plain and with coverage) were all green. Failing test name not captured. Likely in the timer-heavy chat suites. Suggest a `--repeat` sweep on feat/frontend-rebuild to isolate before it starts flaking CI. Not attributed to #445 (merged).

lifeodyssey:
## Isolated. Same bug as the flake note in #437 — one issue, not two.

Ran the `--repeat`-style sweep this issue asked for: 6× full `pnpm --filter web test` (coverage on) against `feat/frontend-rebuild` @ `be2832b4`. Reproduced at **1-in-6**, matching the rate reported here and in #437:

```
run 5 exit=1  Tests  1 failed | 1243 passed (1244)   (runs 1-4, 6 green)
```

**The test** (the piece nobody had captured):

```
FAIL tests/unit/chat/save-wall-dismissal.test.tsx
  > closing the wall after a link is sent is 'going to read the email'
  > keeps the intent so create-on-login can still replay it
  expect(localStorage.getItem(DEFERRED_SAVE_KEY)).toContain("a")   // was null
```

Your guess was right — timer-heavy chat suite, but the mechanism is not a timer. `LoginForm` announced a dispatched magic link from a **passive effect**, so `onSent` landed a scheduler tick *after* the "sent" banner hit the DOM. The test helper waits on the banner, then dismisses the modal; under load the dismissal beat the notification, the P5 save gate read it as a **cancellation**, and it cleared the deferred intent.

Same window exists in production for a user who clicks 閉じる the moment the confirmation appears — it silently drops their save intent. Third instance of the #433 "only fails when the machine is busy" pattern.

**Fixed in #475**: `onSent` now fires in the send continuation next to the status commit (React batches them into one commit, so the notification cannot trail the banner), plus a regression test that resolves the send promise and asserts the callback fired after a single microtask — deterministically red against the old code. Confirmation sweep in the PR.

Suggest closing this as fixed once #475 lands.

lifeodyssey:
Duplicate of the race fixed in #475 (merged): rate (1-in-6 under coverage load) and shape match exactly; the sweep found exactly one flaky test (save-wall-dismissal). Reopen if a different 1-in-N failure resurfaces.

Issue #437, opened 2026-07-26
Follow-ups from #261 review: untested D7 catch, one-way drill-down, unwired spot checkbox, + an unreproduced web-suite flake


lifeodyssey:
Follow-ups deferred from the Fable review of PR #435 (#261 S1.4). All three were ranked P3 — real, but not merge-blocking. Filed so they aren't lost.

### 1. The `.catch` → D7 path is real but untested
`bubbleMapController.ts` previously did `void mountBubbleMap(options).then(...)` with no catch — a genuine unhandled rejection. #435 added a `.catch` that reports `"fallback"` and drives the D7 placeholder. The path is reachable (dynamic-import failure, or the MapLibre `Map` constructor throwing without WebGL), but it lives in the coverage-excluded controller and every unit test injects a fake `attach`, so nothing exercises the real catch.

**Action:** one Tester browser scenario with WebGL disabled, asserting the D7 doodle + 「地図アプリで開く」 link render instead of a blank frame. Acceptable to leave unit-untested per the existing exclude-ledger precedent (S0.4/S5.2).

### 2. C3b drill-down is one-way
Selecting a cluster bubble sets `drill` and switches to the C3a view with no way back to the overview. The S1.4 spec only demands drill-in, so this is not an AC violation — but it is a dead end for the user.

**Action:** add a back affordance (breadcrumb or 「全体に戻る」 chip) when a drill is active.

### 3. The spot-card checkbox is wired to nothing
`SearchCard`'s per-spot checkbox is uncontrolled with no state behind it. AC1 only demands its presence and that it be properly labelled (verified by mutation in review), so #435 is compliant — the selection semantics belong to the selected-route flow.

**Action:** #273 (S1.7) owns the E2 selection tray (「N 件選択中・ルートを組み直す」). Whoever implements it must wire these checkboxes rather than introducing a second selection mechanism. Flagged here so the dangling control isn't forgotten.

---

### Also worth recording: an unreproduced flaky test in the web suite
While independently re-running gates on #435 I saw **1 test fail in 1 of 5 full-suite runs**; the following 4 runs were 807/807 and then 811/811 green. I did not capture which test failed and could not reproduce it in 3 targeted retries. This repo has already had two timing-dependent flakes fixed today (#433), so this is plausibly a third.

**Action:** if CI shows an intermittent web-suite failure, capture the test name rather than re-running — that is the missing piece here.

lifeodyssey:
[...]

Issue #261, opened 2026-07-07
S1.4 — Search result cards and static map views


lifeodyssey:
## Releasable statement
A single-cluster search result renders a spot-card group plus a static map; a multi-cluster search renders a nationwide bubble overview, and selecting a cluster drills into the single-cluster view.

## AC summary
- AC count: 5
- Test type distribution: unit=1, integration=0, browser=4, eval=0, api=0, manual/ops=0

## Dependencies
S0.4 (map spike), S1.1.

## Spec source
`docs/superpowers/specs/2026-07-06-frontend-rebuild/iter-1.md` (commit 6fc13b8)

### S1.4 Search content shapes + static map (C3a/C3b)

**User story**: As a user searching for a title's pilgrimage spots, I want a single-cluster result to show top-6 spot cards + a map, and a multi-cluster result to show a nationwide bubble overview, so I'm never overwhelmed by hundreds or thousands of pins.

**Design basis**: `spec-chat-page-states.md` §C3a/C3b; `spec-chat-page-design.md` §4/§4.1 (volume measurements; the map should be read as MapLibre per X1); `user-journey.md` §4 "the cluster-overview card".

**Releasable statement**: a single-cluster search renders a spot-card group + a static MapLibre map; a multi-cluster search renders a nationwide bubble map + cluster-card group, and selecting a cluster drills into C3a.

**AC**:
- C3a renders the top-6 spot cards sorted by popularity/photo availability (screenshot cover + episode tag + checkbox) + a static map with ≤50 pins -> browser
- C3b (≥2 clusters or a >50km envelope) renders only bubbles (area ∝ count, white numeric badge); this zoom level never draws individual pins -> browser
- Empty: 0 spots within view for the search result renders the D2 "0 pilgrimage spots" state (see S1.6), not a silently empty map -> browser
- Error: a MapLibre static-tile load failure degrades to a hand-drawn SVG placeholder (D7 state) + a 「地図アプリで開く」 external link -> browser
- i18n: cluster names and count badges render place names correctly under ja/zh/en -> unit

**Files changed**: `apps/web/src/components/chat/SpotCardGrid.tsx`, `apps/web/src/components/chat/CircleOverviewMap.tsx`, `apps/web/src/components/map/StaticMap.tsx` (MapLibre wrapper), `apps/web/src/components/chat/registry.ts`.

**Dependencies**: S0.4 (map spike), S1.1.

## Design References

[...]

PR #475, opened 2026-07-29
fix(web): S1.4 follow-ups — real D7 mount-failure test + C3b drill-back chip (#437)

lifeodyssey:
Closes the still-live parts of #437 (follow-ups deferred from the Fable review of #435 / #261 S1.4). Each item was audited against the current tree first — one of the four was already fixed by later work and is **not** redone here; another turned out to be a real product race, not just a slow test.

## Audit result

| # | Item | Verdict |
|---|---|---|
| 1 | `.catch` → D7 path untested | **Still live → fixed here** |
| 2 | C3b drill-down is one-way | **Still live → fixed here** |
| 3 | Spot-card checkbox wired to nothing | **Already covered by #461** — evidence on the issue, no rework |
| 4 | Unreproduced web-suite flake | **Reproduced, named, root-caused, fixed** — same bug as #465 |

---

## 1 — the real mount-failure → D7 path is now exercised

`attachBasemap`'s `.catch` was only reachable through a genuine dynamic-import / WebGL failure, and every other suite injects a fake `attach`, so nothing ran it. `tests/unit/chat/basemap-mount-failure.test.tsx` stubs the `maplibre-gl` module itself so its `Map` constructor throws — the *real* controller then runs, rejects, and drives the placeholder. Two assertions:

- a `SearchResult` rendered with the default (real) `attach` shows the D7 doodle + `地図アプリで開く` link carrying the spot's coordinates, and **no** `.chat-search-map__gl` frame;
- cleaning up before the mount rejects leaves `onStatus` uncalled (the `attachment.active` guard).

The coverage exclude-ledger is untouched — `bubbleMapController.ts` stays excluded; this is a behavioural test, not a coverage move.

## 2 — the drill is no longer a dead end

Selecting a cluster bubble set `drill` and swapped to the C3a view with nothing to press to get back. `SearchResult` now renders `DrilledClusterView`, which prepends a back chip; pressing it clears `drill` and restores the bubble overview, and a second cluster can then be drilled into.

Design alignment: the chip reuses `.chat-chip` outright, inheriting the 動森 cream surface, `0 3px 0 var(--shadow-3d)` press, 12px radius and 44px hit target — `.chat-drill__back` carries **layout only**, asserted in the CSS test so a future raw colour regresses loudly. Copy is trilingual and follows the design-sync `← …に戻る` pattern from `Chat 状态总览.html`: `← 全体に戻る` / `← 返回全部区域` / `← Back to all areas`.

## 4 — the flake: reproduced, then fixed at the source

A 6× full-suite sweep (coverage on) reproduced it at **1-in-6**, matching #465's reported rate:

```
run 5 exit=1  Tests  1 failed | 1243 passed (1244)   (runs 1-4, 6 green)

FAIL tests/unit/chat/save-wall-dismissal.test.tsx
  > keeps the intent so create-on-login can still replay it
  expect(localStorage.getItem(DEFERRED_SAVE_KEY)).toContain("a")   // was null
```

**Root cause.** `LoginForm` announced a dispatched magic link from a *passive effect* (`useEffect(… status === "sent" …)`). The confirmation banner reaches the DOM in the status commit, but the effect calling the caller ran a scheduler tick later. The test helper waits on the banner, then dismisses; under load the dismissal beat the notification, so the P5 save gate read it as a **cancellation** and cleared the deferred intent.

Round-2 review then found the window was **wider than the flake showed**: the gate only counted a *completed* send, so the entire request latency — not just one scheduler tick — was a period in which closing the modal destroyed the intent. That part was deterministically reproducible, not a race at all.

**Fix.** The notification is now `onSendCommitted`, fired at the **start** of the send:

```ts
setStatus("submitting");
onCommitted.current?.();
setStatus(await sendMagicLink({ … }));
```

The user's commitment is the click, not the server's reply, so there is no longer any interval — tick-sized or latency-sized — in which the modal is dismissable but the caller has not been told. A send that comes back as an error also counts as committed: a kept intent is bounded by consume-once + TTL, while a cleared one is lost work.

The callback is held in a latest-ref (the `useTurnTiming` precedent), so an inline caller callback can neither churn `submit`'s identity nor be captured stale. This also removes a latent bug in the old effect: with a non-memoized callback, its `[status, onSent]` deps re-fired on every parent render while status stayed `"sent"`.

**Focus management (P2).** The C3b drill swapped views without moving focus, stranding keyboard and screen-reader users on `<body>`. The back chip now takes focus on drill-in, the originating bubble reclaims it on back (the one actually chosen, not index 0), and the D7 placeholder is the fallback target when no bubble survives.

---

## Mutation evidence

| Mutation | Result |
|---|---|
| delete the `.catch` from `attachBasemap` | D7 test fails (`expected null not to be null`) |
| drop the `if (attachment.active)` guard | cleanup test fails (`onStatus` called once) |
| make `onBack` a no-op | 2 drill tests fail |
| delete the back chip from the JSX | 3 drill tests fail |
| drop `align-self` from `.chat-drill__back` | CSS token test fails |
| notify on the send's reply instead of at dispatch | 4 tests fail (banner-ordering, in-flight, wall-dismissal, error-send) |
| always `clearDeferredSave()` on dismiss | 3 wall-dismissal tests fail |
| drop the back chip's autofocus | drill-in focus test fails |
| refocus bubble 0 instead of the chosen index | "not simply the first one" test fails |

The cleanup-guard test *survived* its first mutation (microtask flush too shallow to get past two dynamic imports) and was strengthened until the mutant died. The flake regression test was written red-first against the old effect-based code.

## Gates (`apps/web`)

- `pnpm run typecheck` — exit 0
- `pnpm run lint:oxlint` (type-aware, `--deny-warnings`) — exit 0
- `pnpm test` — 182 files / **1252 passed**, coverage **98.57 / 95.24 / 98.99 / 99.47**. Branches ratcheted **94 → 95** on the measured 95.24; statements/functions/lines left at 98/98/99 (they dipped a hair versus the S1.3 measurement, so raising them would be brittle). Never lowered.
- **Confirmation sweep after the flake fix: 8× full suite, all green.**

No suppressions; production functions stay ≤10 lines and every touched test file stays well under 200.

### DIFF
--- a/apps/web/src/components/auth/LoginForm.tsx
+++ b/apps/web/src/components/auth/LoginForm.tsx
@@ -1,4 +1,3 @@
-import { useEffect } from "react";
 import type { ChangeEvent } from "react";
 import type { Dict } from "../../i18n/dictionaries";
 import { useDict } from "../../i18n/context";
@@ -59,22 +58,15 @@ function FormFields({ form }: { form: MagicLinkForm }) {
   </>);
 }
 
-/** Announce a dispatched link exactly once: the caller uses it to tell
- * "closed to go read the email" apart from "cancelled". */
-function useSentOnce(status: FormStatus, onSent: (() => void) | undefined): void {
-  useEffect(() => {
-    if (status === "sent") onSent?.();
-  }, [status, onSent]);
-}
-
 export interface LoginFormProps {
+  /** Announced once per dispatched link, from the send continuation itself, so
+   * the caller can tell "closed to go read the email" from "cancelled". */
   readonly onSent?: () => void;
 }
 
 export function LoginForm({ onSent }: LoginFormProps = {}) {
-  const form = useMagicLinkForm();
+  const form = useMagicLinkForm(onSent);
   const auth = useDict().auth;
-  useSentOnce(form.status, onSent);
   return (
     <form className="login-form" onSubmit={form.onSubmit} noValidate aria-label={auth.title}>
       <FormFields form={form} />
--- a/apps/web/src/components/auth/LoginForm.tsx
+++ b/apps/web/src/components/auth/LoginForm.tsx
@@ -1,7 +1,7 @@
 import type { ChangeEvent } from "react";
 import type { Dict } from "../../i18n/dictionaries";
 import { useDict } from "../../i18n/context";
-import { type FormStatus, type MagicLinkForm, type ValidationKey, useMagicLinkForm } from "./useMagicLinkForm";
+import { type FormStatus, type MagicLinkForm, type SendCommitted, type ValidationKey, useMagicLinkForm } from "./useMagicLinkForm";
 
 type Auth = Dict["auth"];
 
@@ -59,13 +59,13 @@ function FormFields({ form }: { form: MagicLinkForm }) {
 }
 
 export interface LoginFormProps {
-  /** Announced once per dispatched link, from the send continuation itself, so
-   * the caller can tell "closed to go read the email" from "cancelled". */
-  readonly onSent?: () => void;
+  /** Announced once the send is dispatched, so a caller can tell "closed to go
+   * read the email" from "cancelled" — see `SendCommitted` for the timing. */
+  readonly onSendCommitted?: SendCommitted;
 }
 
-export function LoginForm({ onSent }: LoginFormProps = {}) {
-  const form = useMagicLinkForm(onSent);
+export function LoginForm({ onSendCommitted }: LoginFormProps = {}) {
+  const form = useMagicLinkForm(onSendCommitted);
   const auth = useDict().auth;
   return (
     <form className="login-form" onSubmit={form.onSubmit} noValidate aria-label={auth.title}>
--- a/apps/web/src/components/auth/LoginModal.tsx
+++ b/apps/web/src/components/auth/LoginModal.tsx
@@ -2,13 +2,14 @@ import { useEffect } from "react";
 import type { Dict } from "../../i18n/dictionaries";
 import { useDict } from "../../i18n/context";
 import { LoginForm } from "./LoginForm";
+import type { SendCommitted } from "./useMagicLinkForm";
 
 interface LoginModalProps {
   open: boolean;
   onClose: () => void;
-  /** Fires when a magic link was dispatched, so a caller can tell a
-   * "closed to go read the email" dismissal apart from a cancellation. */
-  onSent?: () => void;
+  /** Fires once the send is dispatched, so a caller can tell a "closed to go
+   * read the email" dismissal apart from a cancellation. */
+  onSendCommitted?: SendCommitted;
 }
 
 function useEscapeToClose(open: boolean, onClose: () => void): void {
@@ -22,25 +23,25 @@ function useEscapeToClose(open: boolean, onClose: () => void): void {
   }, [open, onClose]);
 }
 
-function LoginDialog({ auth, onClose, onSent }: { auth: Dict["auth"]; onClose: () => void; onSent?: () => void }) {
+function LoginDialog({ auth, onClose, onSendCommitted }: { auth: Dict["auth"]; onClose: () => void; onSendCommitted?: SendCommitted }) {
   return (
     <div className="login-modal" role="dialog" aria-modal="true" aria-label={auth.title} onClick={(event) => { event.stopPropagation(); }}>
       <button className="login-modal__close" type="button" aria-label={auth.close} onClick={onClose}>×</button>
       <h2 className="login-modal__title">{auth.title}</h2>
       <p className="login-modal__subtitle">{auth.subtitle}</p>
-      <LoginForm onSent={onSent} />
+      <LoginForm onSendCommitted={onSendCommitted} />
     </div>
   );
 }
 
 /** Magic-link login modal wired to the Neon Auth (Better Auth) client. */
-export function LoginModal({ open, onClose, onSent }: LoginModalProps) {
+export function LoginModal({ open, onClose, onSendCommitted }: LoginModalProps) {
   const auth = useDict().auth;
   useEscapeToClose(open, onClose);
   if (!open) return null;
   return (
     <div className="login-modal__mask" role="presentation" onClick={onClose}>
-      <LoginDialog auth={auth} onClose={onClose} onSent={onSent} />
+      <LoginDialog auth={auth} onClose={onClose} onSendCommitted={onSendCommitted} />
     </div>
   );
 }
--- a/apps/web/src/components/auth/useMagicLinkForm.ts
+++ b/apps/web/src/components/auth/useMagicLinkForm.ts
@@ -25,12 +25,17 @@ function callbackUrl(): string {
   return `${window.location.origin}/auth/callback`;
 }
 
-function useSendMagicLink(email: string): [FormStatus, () => Promise<void>] {
+/** `onSent` fires in this continuation, alongside the status commit — never from
+ * a passive effect, which would land a tick after the banner is observable and
+ * lose a race with an immediate dismissal (issues #437 / #465). */
+function useSendMagicLink(email: string, onSent?: () => void): [FormStatus, () => Promise<void>] {
   const [status, setStatus] = useState<FormStatus>("idle");
   const submit = useCallback(async () => {
     setStatus("submitting");
-    setStatus(await sendMagicLink({ email: email.trim().toLowerCase(), callbackURL: callbackUrl() }));
-  }, [email]);
+    const result = await sendMagicLink({ email: email.trim().toLowerCase(), callbackURL: callbackUrl() });
+    setStatus(result);
+    if (result === "sent") onSent?.();
+  }, [email, onSent]);
   return [status, submit];
 }
 
@@ -45,10 +50,10 @@ function useSubmit(email: string, setValidation: SetValidation, submit: () => Pr
   }, [email, setValidation, submit]);
 }
 
-export function useMagicLinkForm(): MagicLinkForm {
+export function useMagicLinkForm(onSent?: () => void): MagicLinkForm {
   const [email, setEmail] = useState("");
   const [validation, setValidation] = useState<ValidationKey | null>(null);
-  const [status, submit] = useSendMagicLink(email);
+  const [status, submit] = useSendMagicLink(email, onSent);
   const onSubmit = useSubmit(email, setValidation, submit);
   return { email, status, validation, setEmail, onSubmit };
 }
--- a/apps/web/src/components/auth/useMagicLinkForm.ts
+++ b/apps/web/src/components/auth/useMagicLinkForm.ts
@@ -1,4 +1,5 @@
-import { useCallback, useState } from "react";
+import { useCallback, useRef, useState } from "react";
+import type { RefObject } from "react";
 import { type MagicLinkResult, sendMagicLink } from "../../lib/auth/neonAuth";
 
 export type ValidationKey = "email_required" | "email_invalid";
@@ -25,17 +26,31 @@ function callbackUrl(): string {
   return `${window.location.origin}/auth/callback`;
 }
 
-/** `onSent` fires in this continuation, alongside the status commit — never from
- * a passive effect, which would land a tick after the banner is observable and
- * lose a race with an immediate dismissal (issues #437 / #465). */
-function useSendMagicLink(email: string, onSent?: () => void): [FormStatus, () => Promise<void>] {
+/**
+ * Announced once per dispatched request, at the **start** of the send — the
+ * user's commitment is the click, not the server's reply. A caller that keys
+ * off dismissal (the P5 save wall) must count an in-flight request as
+ * committed, or the entire request latency is a window in which closing the
+ * modal destroys the intent. Never announced from a passive effect either:
+ * that lands a tick after the banner is observable (issues #437 / #465).
+ */
+export type SendCommitted = () => void;
+
+/** Latest-ref so a caller's inline callback cannot churn `submit`'s identity or
+ * be captured stale, mirroring `useTurnTiming`'s `reportRef`. */
+function useLatest(callback: SendCommitted | undefined): RefObject<SendCommitted | undefined> {
+  const ref = useRef(callback);
+  ref.current = callback;
+  return ref;
+}
+
+function useSendMagicLink(email: string, onCommitted: RefObject<SendCommitted | undefined>): [FormStatus, () => Promise<void>] {
   const [status, setStatus] = useState<FormStatus>("idle");
   const submit = useCallback(async () => {
     setStatus("submitting");
-    const result = await sendMagicLink({ email: email.trim().toLowerCase(), callbackURL: callbackUrl() });
-    setStatus(result);
-    if (result === "sent") onSent?.();
-  }, [email, onSent]);
+    onCommitted.current?.();
+    setStatus(await sendMagicLink({ email: email.trim().toLowerCase(), callbackURL: callbackUrl() }));
+  }, [email, onCommitted]);
   return [status, submit];
 }
 
@@ -50,10 +65,10 @@ function useSubmit(email: string, setValidation: SetValidation, submit: () => Pr
   }, [email, setValidation, submit]);
 }
 
-export function useMagicLinkForm(onSent?: () => void): MagicLinkForm {
+export function useMagicLinkForm(onSendCommitted?: SendCommitted): MagicLinkForm {
   const [email, setEmail] = useState("");
   const [validation, setValidation] = useState<ValidationKey | null>(null);
-  const [status, submit] = useSendMagicLink(email, onSent);
+  const [status, submit] = useSendMagicLink(email, useLatest(onSendCommitted));
   const onSubmit = useSubmit(email, setValidation, submit);
   return { email, status, validation, setEmail, onSubmit };
 }
--- a/apps/web/src/features/chat/components/SearchMap.tsx
+++ b/apps/web/src/features/chat/components/SearchMap.tsx
@@ -8,6 +8,7 @@ import type { BasemapStatus, MountBasemapOptions } from "../../bubble-map/bubble
 import { clusterName, spotCountBadge } from "../search-copy";
 import type { ChatDict } from "../i18n";
 import { MapFallback } from "./ErrorStates/MapFallback";
+import { useAutoFocus } from "./useAutoFocus";
 
 /** Injectable mount so tests (and D7 simulations) never touch WebGL. */
 export type AttachBasemap = (options: MountBasemapOptions) => () => void;
@@ -105,11 +106,12 @@ function bubbleStyle(placement: BubblePlacement): CSSProperties {
   return { width: size, height: size, ...percentStyle(placement) };
 }
 
-type BubbleProps = Readonly<{ placement: BubblePlacement; dict: ChatDict; onClick: () => void }>;
+type BubbleProps = Readonly<{ placement: BubblePlacement; dict: ChatDict; onClick: () => void; refocus: boolean }>;
 
-function ClusterBubble({ placement, dict, onClick }: BubbleProps) {
+function ClusterBubble({ placement, dict, onClick, refocus }: BubbleProps) {
+  const ref = useAutoFocus<HTMLButtonElement>(refocus);
   return (
-    <button type="button" className="chat-map-bubble" style={bubbleStyle(placement)} onClick={onClick}>
+    <button ref={ref} type="button" className="chat-map-bubble" style={bubbleStyle(placement)} onClick={onClick}>
       <span className="chat-map-bubble__name">{placement.region}</span>
       <span className="chat-map-bubble__count">{spotCountBadge(placement.count, dict)}</span>
     </button>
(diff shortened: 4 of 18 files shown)


lifeodyssey:
## Post-fix confirmation sweep: 8/8 green

```
post-run 1 exit=0  Tests  1246 passed (1246)
post-run 2 exit=0  Tests  1246 passed (1246)
post-run 3 exit=0  Tests  1246 passed (1246)
post-run 4 exit=0  Tests  1246 passed (1246)
post-run 5 exit=0  Tests  1246 passed (1246)
post-run 6 exit=0  Tests  1246 passed (1246)
post-run 7 exit=0  Tests  1246 passed (1246)
post-run 8 exit=0  Tests  1246 passed (1246)
```

Same command and machine that produced the 1-in-6 failure on the base tree (`pnpm --filter web test`, coverage on) — 8 consecutive clean runs, past the point where the pre-fix sweep broke (run 5).

Green runs are the weaker half of the evidence, though. The stronger half is that the mechanism is gone from the source: `onSent` no longer travels through a passive effect, so there is no longer a tick in which the banner is observable but the caller has not been told. The regression test asserts exactly that ordering and is deterministically red against the old code.

lifeodyssey (commented):
## Independent review — Fable seat (dual review, no reference to the other seat)

**Verdict: approve.** Independent gates in a detached worktree at `origin/fix/437-s14-followups`: typecheck 0, type-aware oxlint 0, **182 files / 1246 tests green**, coverage 98.56 / 95.31 / 98.99 / 99.47 — all matching the self-report.

### 1. onSent race fix (the core item) — correct, and verified

- The move from the passive `useSentOnce` effect into the send continuation is the right fix, not a mask. `markLinkSent` is a **ref** write, so it now lands *synchronously before* React commits the "sent" banner — the "banner observable but gate not yet told" window is structurally gone for the P5 consumer, independent of batching.
- **Mutation replayed**: I faithfully restored the old effect-based code (`useEffect(... status === "sent" ...)` in `LoginForm` + removed the continuation call). The new timing test in `tests/unit/login-form.test.tsx` goes deterministically red (`called 0 times` after one microtask), while all four #462 quadrant tests (close before/after send × button/Esc in `save-wall-dismissal.test.tsx`) stay green — confirming the quadrant semantics are unchanged and the flake was purely a commit-vs-effect ordering bug.
- The old `[status, onSent]` deps double-fire (non-memoized `onSent` re-firing on every parent render while status stayed `"sent"`) is indeed eliminated: `onSent` now fires exactly once per dispatched link.
- Flake re-verification: **3× additional full-suite sweeps (coverage on), all green** on top of my two earlier full runs — 5/5, consistent with the implementer's 8/8.

### 2. D7 catch test — real behaviour, correctly scoped

Stubbing the `maplibre-gl` module so the **real** `attachBasemap` runs is strictly better than injecting a fake `attach`: the `.catch` → `reportMountFailure` → D7 placeholder chain is now exercised end-to-end, including the coords in the 地図アプリで開く link and the absence of `.chat-search-map__gl`. **Mutation replayed**: dropping the `attachment.active` guard fails the cleanup test with `onStatus` called once (`"fallback"`) — the strengthened mutant is dead. Keeping `bubbleMapController.ts` in the §0.6 coverage exclude-ledger is right: the live GL paths (`wireEvents`/`onMapLoad`/`fitBounds`) still cannot run under jsdom; this is a behavioural test, not a coverage move.

### 3. Drill-down back chip — matches the design pattern

`← 全体に戻る` follows the `← …に戻る` pattern in `docs/design/2026-07-06-design-sync/Chat 状态总览.html` (`← チャットに戻る`); trilingual copy present with a distinctness + `←`-prefix i18n test; `.chat-drill__back` is layout-only with a CSS test pinning `background`/`color` to null so the cream chip tokens stay single-source. The C3a negative test (no back chip on a natively single-cluster result) is a good boundary. **Mutation replayed**: neutralizing the back handler fails exactly the 2 claimed drill tests.

### 4. #461 item-3 evidence — accurate

Verified against the tree: `src/features/chat/selection/useSpotSelection.tsx` exists, `SpotPick` in `SearchResult.tsx` is controlled (`checked={selected.has(spot.id)}` + `toggle`), and the tray tests (`selection-tray.test.tsx`, `chat-page-selection.test.tsx`) exist. The issue comment is accurate; no rework needed.

### Findings

- **P2 (stale comment, trivial fix)** — `apps/web/src/features/chat/save/useSaveGate.ts` lines 84–88: the `useWall` doc still justifies the ref with "React batches **the form's effect** with the dismissal click". That effect no longer exists — this PR is precisely the change that removed it. The comment documents the invariant that just caused a data-loss bug, so it being wrong is worth a one-line follow-up (the ref is still correct; the reason is now "the continuation sets it before the banner ever commits").
- **P2 (a11y, follow-up ok)** — no focus management on the drill view swaps: pressing the back chip removes the focused element and focus drops to `body`; same on drill-in (pre-existing). Suggest: focus the back chip on drill-in, and the drilled bubble (or the map container) on back.
- **P3 (a11y nit)** — the `←` is part of the button's accessible name, so screen readers announce "leftwards arrow 全体に戻る". Consider an `aria-hidden` arrow span; note the i18n test's `startsWith("←")` assertion would need restructuring.
- **P3 (test hygiene)** — the continuation-timing test resolves the send promise outside `act()`; if React act-warnings ever become errors this will surface. Wrapping the resolve + microtask await in `act` keeps it clean.

### #465 disposition

**Closable as duplicate/fixed by this PR.** #465's signature (1-in-6, single failure, coverage load, first-run) matches the reproduced `save-wall-dismissal` failure exactly, and the 6-run sweep surfaced exactly one flaky test at that rate. Caveat for the closing comment: #465 never captured the failing test name, so identity is inferred from rate + shape, not proven — reopen if a different 1-in-N failure resurfaces.

### Quality ratchet

Item 1 → `basemap-mount-failure.test.tsx` (2 tests) · Item 2 → 4 drill tests + C3a negative + CSS + i18n · Item 4 → 2 timing tests, red-first verified by mutation. `ac_total == ac_with_test` for every live item. Coverage floors untouched and never lowered. `make test-eval`: skipped — no agent behaviour/prompt/tool changes.

lifeodyssey (commented):
## Review — Opus, second seat (independent; no other review consulted)

Verified in a detached worktree at `70b2f59f`, deps installed fresh. Verdict: **approve**, with one P1 follow-up that must be tracked before this lands and one sentence in the PR body that needs softening.

### Independent gates (my run, not CI's)

| Gate | Result |
|---|---|
| `pnpm run typecheck` | exit 0 |
| `pnpm run lint:oxlint` (type-aware, `--deny-warnings`) | exit 0 |
| `pnpm test` | exit 0 — 182 files / 1246 passed |
| Coverage | **98.56 / 95.31 / 98.99 / 99.47** — matches the PR body to the digit |
| Suppressions in diff | none (`grep` for `ts-ignore` / `eslint-disable` / `istanbul ignore` / `skip` / `any` → 0) |
| Test-file sizes | 59–125 lines, all well under 200 |
| Prod file sizes | SearchResult 96, LoginForm 75, useMagicLinkForm 59 |
| React act warnings | 0 in the full run |

### Mutation re-runs (2, done myself)

- **cleanup guard** — dropped `if (attachment.active)` from `reportMountFailure` → `basemap-mount-failure.test.tsx:57` fails (`Number of calls: 1`). Mutant dies. I also probed the *margin*: with `settleMount` cut from 5 ticks to **1**, the mutant still dies — so the drain has ~5× headroom and is not the load-sensitive kind. Concern downgraded to P3.
- **passive-effect `onSent`** — reconstructed the old `useEffect([status, onSent])` shape → exactly **one** test fails, the new continuation-timing test. `save-wall-dismissal.test.tsx` stayed green under the reverted code, which is precisely why the bug was invisible in isolation.
- Third, mine: **half-fix** `queueMicrotask(() => onSent?.())` → the timing test also fails. See P2-2 — that mutant would have been product-*safe*, so the test is stricter than the invariant.

### 1 — `onSent` timing semantics: the ordering argument holds

The fix is stronger than "React batches both into one commit". `onSent?.()` runs **synchronously** in the continuation, immediately after `setStatus` merely *schedules*; the commit that paints the banner happens in a later scheduler task. So the notification is **strictly before** the banner is observable, not merely in the same commit. Any consumer that waits on `role="status"` and then acts is now provably safe — and it works for both consumer flavors: `markLinkSent` writes a ref synchronously, and a hypothetical `setState` consumer would batch with `setStatus` in the same microtask.

`#462`'s `linkSent` ref: I traced every read point. Only `useCloseLogin` (`useSaveGate.ts:106-110`) reads it, only writers are `openLogin` (reset) and `markLinkSent`, and the only `onSent` wiring in the tree is `TimedItinerary.tsx:142`. No stale-closure exposure there — `markLinkSent` is `useCallback([], …)`. Error path preserved: the `result === "sent"` guard is intact and covered by the "stays silent when the send did not produce a link" test.

### 2 — the product race: eliminated in one window, wide open in the other

**Eliminated**, for the banner-then-dismiss window — by the ordering argument above, not by narrowing.

**But the fix does not cover the larger window**, and this is my P1. Dismissing while the send is still **in flight** still drops the deferred intent, even though the link does go out. I reproduced it deterministically with a throwaway probe (since deleted) against this branch's HEAD:

```
click 送信  →  click 閉じる  →  send resolves "sent"
expect(localStorage.getItem(DEFERRED_SAVE_KEY)).toContain("a")
  ✗ received null
```

`useCloseLogin` clears on `!linkSent.current`; nothing sets that flag until the network round-trip returns, and nothing aborts or awaits the in-flight send. The window here is the **whole request latency** — hundreds of milliseconds — versus the sub-millisecond one this PR closed. Same user-visible symptom (create-on-login silently loses the save), same subsystem, no test.

Not introduced here, and I don't think it should block the merge. But "the race is gone from the source, not merely unobserved" reads as covering this and doesn't. Please scope that sentence to the post-banner window and file the follow-up.

Suggested shape: set `linkSent.current = true` optimistically at submit time (an attempted send is already "went to read the email" territory), or have `closeLogin` skip the clear while a send is outstanding.

### 3 — maplibre stub: no leakage

`vi.mock` is module-graph-scoped per test file and vitest isolates per file by default (no `isolate: false` / `singleThread` in `vitest.config.ts`). `basemap-mount-failure.test.tsx` is the **only** file under `tests/` that references maplibre at all. The module-level `protocolReady` singleton in `bubbleMapController.ts` is mutated by this test, but it lives in that file's own module registry, so nothing else sees it. Clean.

D7 assertion strength is adequate — doodle present, `d7Message` text, the map-app link's `href` carrying `34.89,135.8`, and `.chat-search-map__gl` absent. The `.catch` is genuinely exercised (`createMap` throws synchronously inside the async `mountBasemap` → rejection → `reportMountFailure`), which my mutation confirmed.

### 4 — flake claim: corroborated

Your raw sweep artifacts are still on this machine (`/tmp/sweep-1..6.log`, timestamped today). `sweep-5.log` reads `Tests 1 failed | 1243 passed (1244)` on `save-wall-dismissal.test.tsx:77` — matches the PR body verbatim and matches #465. Good, that's an artifact and not a narrative.

My own 5× sweep on the *reverted* (passive-effect) tree came back **0/5** — but I ran without coverage instrumentation, which is the load your repro depended on, so this is a non-reproduction at low sample size, not a contradiction. Combined 1-in-11 is consistent with the reported rate. The deterministic guard is the timing test, and I proved it kills the mutant, which matters more than the sweep either way.

**Other passive-effect notification points** — I walked all 25 `useEffect` sites in `apps/web/src`. Exactly one other fires a caller-supplied callback from a passive effect: `useTurnTiming` (`use-turn-timing.ts:52-56`). Benign — fire-and-forget analytics recorder with no DOM-observe-then-act consumer, and it already uses the latest-ref pattern. Your "third instance of #433" framing checks out; there is no fourth.

### 5 — coverage floors

Locally reproduced to the digit. The dip cannot come from this diff — `SearchResult.tsx` reports 100% statements/lines/functions, so every production line added here is covered; the statements/lines/functions delta versus the S1.3 mark is base drift. Holding floors is defensible. One disagreement, below.

---

## Findings

**P1 — in-flight dismissal still drops the deferred save intent** · `src/features/chat/save/useSaveGate.ts:94-110` + `src/components/auth/useMagicLinkForm.ts:33-38` · reproduced deterministically (above). Wider window than the one fixed. *Fix:* optimistic `linkSent` at submit, or gate the clear on an outstanding send. *Also:* soften "the race is gone from the source" to name the post-banner window. **test_needed: unit.**

**P2-1 — `onSent` in the `useCallback` deps reintroduces a narrow stale-closure** · `useMagicLinkForm.ts:38` · an in-flight `submit` holds the render-time `onSent`; a parent that re-renders mid-send with a fresh inline callback gets the stale one — which the old effect (`[status, onSent]`) did not. Benign today because the sole caller memoizes, but `LoginFormProps.onSent` documents no memoization contract, and `submit`/`onSubmit` identity now churns per parent render. *Fix:* latest-ref, exactly as `use-turn-timing.ts:52-55` already does with `reportRef` — that also drops the dep and re-stabilizes `submit`. **test_needed: unit.**

**P2-2 — the timing test asserts structure, not the invariant** · `tests/unit/login-form.test.tsx:78-88` · "fires within one microtask" is stricter than "fires before the banner": my `queueMicrotask` mutant is product-safe yet fails the test, so a future safe refactor breaks it. *Fix:* keep it, and add the behavioral half — have `onSent` capture `document.querySelector('[role="status"]')` at call time and assert it is `null`. That pins the actual ordering guarantee. **test_needed: unit.**

**P2-3 — ratchet branches** · `vitest.config.ts` thresholds · branches rose 94.57 → **95.31** while the other three dipped on base drift. AGENTS.md says floors ratchet up only; it doesn't say four independent dimensions move together. Branches 94 → 95 is earned and free. *Fix:* raise branches only, leave the other three.

**P2-4 — fixed-tick drain behind a negative assertion** · `basemap-mount-failure.test.tsx:31-35` · `settleMount`'s 5×`setTimeout(0)` is the shape AGENTS.md bans ("mock the clock"). My margin probe says it isn't a flake risk here (1 tick suffices), so this is hygiene: make the stubbed `Map` throw off a deferred you resolve, then await that, and the drain disappears.

**P3-1 — the "←" is inside the accessible name** · `search-i18n.ts` · SRs announce the arrow. Prefer a decorative `aria-hidden` span and copy of just `全体に戻る`; note `search-copy-i18n.test.ts:42-46` currently asserts the glyph *is* in the string, entrenching it.

**P3-2 — no focus management on drill / back** · `SearchResult.tsx:64-70` · pressing back unmounts the focused chip and focus falls to `<body>`. No keyboard test for the C3b→C3a→C3b journey.

**P3-3 — `drill` survives a `spots` change** · `SearchResult.tsx:89` · the held `SpotCluster` comes from a previous `spots` prop. Pre-existing, but the back chip makes drilled state stickier. A `key` on the caller or a reset would settle it.

**P3-4 — class-name assertions** · `search-card-c3b.test.tsx:80,86` and `basemap-mount-failure.test.tsx:47,52` · `className.toContain("chat-drill__back")` and `.chat-map-bubble` / `.chat-map-fallback__doodle` selectors are on the repo's own test-smell list. The chip already has a role+name; the CSS test covers the class contract separately.

**P3-5 — a *rejecting* `sendMagicLink` is still uncovered** · `useMagicLinkForm.ts:34` · only the resolved-`"error"` path is tested; a throw leaves status stuck on `"submitting"` and surfaces as an unhandled rejection out of `void submit()`. Same for an `onSent` that throws, which now escapes into that rejection instead of React's error boundary. Pre-existing.

**P3-6 — TDD order isn't provable from history** · two of three commits carry test + implementation together, so there is no red commit to point at. The mutation table (two of which I reproduced) is a decent substitute; noting it for the ratchet record, not asking for a rewrite.

---

**Verdict: approve.** The `onSent` fix is correct for the reason stated and then some — strictly-before, not merely same-commit. The D7 test reaches real code, the mutation evidence survives independent re-running, the flake claim has artifacts behind it, and my gates are green. Merge once P1 has a tracked follow-up issue and the "gone from the source" sentence is scoped.

lifeodyssey (commented):
## Delta confirmation — Opus, second seat · `c1457cbb`

Re-verified in a fresh detached worktree, deps reinstalled. **No regressions. Semantic change: accepted. Final verdict: approve — merge it.**

### Gates (my run)

| Gate | Result |
|---|---|
| `pnpm run typecheck` | exit 0 |
| `pnpm run lint:oxlint` | exit 0 |
| `pnpm test` | exit 0 — 182 files / **1252** passed (+6) |
| Coverage | 98.57 / **95.24** / 98.99 / 99.47 vs floors 98 / **95** / 98 / 99 |
| act warnings | 0 |
| Suppressions in delta | none |
| Sizes | biggest touched test 128 lines, biggest touched source 178 |
| Stale identifiers | `onSent` / `markLinkSent` / `linkSent` → **0 hits** across `src/` and `tests/` |

### ① P1 — closed, and closed structurally

`onCommitted.current?.()` now fires synchronously inside the click handler, before `sendMagicLink` is even called. There is no longer a window to lose: the flag is set before any await exists. My never-resolving-promise scenario is now a passing test at `save-wall-dismissal.test.tsx:87-95`, and `login-form.test.tsx` has the component-level twin.

**Mutation (mine):** reverted the body to `const result = await …; setStatus(result); if (result === "sent") onCommitted.current?.()` — i.e. exactly the code I reviewed last round. **4 tests fail**, including *"keeps the intent when the wall is closed while the request is still in flight"* and *"still reports a send that came back as an error"*. Red→green on my reproduction, confirmed by re-running it myself rather than taking the table's word.

The banner-ordering test is also genuinely behavioral now — `seenBanner` capturing `queryByRole("status")`/`("alert")` at call time and asserting `[null, null]` is the invariant I asked for, not the "within one microtask" structural proxy. It dies under the same mutant.

### ② P2s — all three landed as recommended

- **latest-ref**: `useLatest` mirrors `useTurnTiming`'s `reportRef` exactly, and `submit`'s dep list is now `[email, onCommitted]` where `onCommitted` is a stable ref object. Identity churn and the stale-capture path are both gone.
- **banner-ordering assertion**: above.
- **branches 94 → 95**: correct dimension, correct direction, and the comment records the measurement. One note below on margin.

Bonus: P3-2 (focus management) was also taken, and taken well. `useAutoFocus` is 18 lines with a comment that explains *why* a view swap needs it. Three focus tests, and the third — "returns focus to the bubble actually chosen, not simply the first one" — is the one that makes the other two worth having.

**Mutations (mine):** `useAutoFocus` made a no-op → 3 tests fail. `back()` changed to `setRefocus(drill ? 0 : null)` → **exactly one** test fails, the "not simply the first one" one. The test names are honest about what they pin.

### ③ Semantic ruling: **accept** — an errored send counts as committed

Agreed with the coordinator, and I'd go further: this is not merely acceptable, it is the **only coherent** semantics once the in-flight window is closed.

The decisive argument is structural, not preferential. "Errors don't count" requires knowing the reply. The reply arrives after the latency window. So you can have at most two of: *(a)* no in-flight window, *(b)* errors don't count, *(c)* no retroactive un-commitment. The alternative to what you shipped is committing at dispatch and then *un*-committing on an error reply — which reopens the same class of bug in a nastier form: the flag would flip back to `false` under a modal the user may have already closed, and the ordering of that flip against the dismissal is precisely the race #465 was about. Commit-at-dispatch with no take-backs is the one arrangement that has no window at all.

The cost asymmetry points the same way, and I verified both bounds rather than assuming them:

- **consume-once** — `createOnLogin.ts:34-35` reads then immediately `clearDeferredSave()`. Real.
- **TTL** — `DEFERRED_SAVE_TTL_MS = 30 min`, enforced in `readDeferredSave` *and* swept by `pruneDeferredSave` on mount. Real.

So: cost of keeping a stale intent = at worst one unexpected — and reversible — saved route, within 30 minutes, on a login the user has to actually complete. Cost of clearing a live one = silent loss of the user's work, no signal, no recovery path. That asymmetry isn't close. And note the error case adds no new *class* of risk: the already-accepted "sent, then dismissed, then never clicked the link" path has the identical exposure.

Two things that make me comfortable signing this rather than just tolerating it:

1. **The rename does the honest work.** `linkSent` → `committed`, `onSent` → `onSendCommitted`, `SendCommitted`. If the flag still said "link sent" while being set for a failed send, the next reader would file it as a bug. It doesn't, so they won't.
2. **The inversion is visible.** The old test wasn't deleted, it was rewritten to assert the opposite with a comment saying why. That's the right way to change a pinned semantic.

Practical upside worth naming: the errored send is the case where the user is *most* likely to come back. Clearing there means "hit an error → close → log in later via the D8 banner → save silently doesn't replay", which is the exact failure mode this whole issue is about.

### New findings (all P3, none blocking)

**P3-a — `refocusIndex` is never cleared once consumed** · `SearchResult.tsx:85-91` · `back()` sets it and only `select()` clears it, so it stays set indefinitely after the first back-out. If the basemap later flips `ready → fallback` (WebGL loss), `BubbleMapFallback` mounts with `refocus={true}` and pulls focus to the placeholder — an unsolicited, non-user-initiated focus move (WCAG 3.2.1 territory). Narrow, but free to fix: make it one-shot by clearing after the focus lands.

**P3-b — `not_configured` commits although nothing was dispatched** · `neonAuth.ts:35-37` returns early without a request, yet the click already committed. Bounded by TTL + consume-once, and a misconfigured build can't complete a login anyway, so the practical exposure is nil. If you want it airtight, skip committing when `!isNeonAuthConfigured()`.

**P3-c — branch floor margin is thin** · measured 95.24 against a floor of 95, i.e. roughly four uncovered branches of headroom. That's the ratchet working as designed (CI fails, you add tests), and it was my ask — just flagging so nobody is surprised when a future PR trips it.

**P3-d — `useLatest` writes the ref during render.** Matches the existing `use-turn-timing.ts` precedent exactly, and the callback is only ever read from event handlers and async continuations, never during render, so there's no concurrent-mode hazard here. Noting only so that if you ever move to `useEffectEvent`, both sites move together.

Still open from round one and unchanged, both still P3: `drill` survives a `spots` change, and the `←` glyph sits inside the accessible name.

---

**Verdict: approve.** The P1 is gone structurally rather than narrowed, all three P2s landed in the form I asked for, four mutants died under my own hand (including a discriminating one I wrote myself), the gates are green on my machine and not just CI's, and the semantic change is the correct call for a reason stronger than the one offered. Ship it.

PR #478, opened 2026-07-29
docs(ops): #284 Task 7 — egress network policy spike + fallback acceptance

lifeodyssey:
## Summary

- Closes **#284 Task 7** (container-level egress network policy) per the OQ-5 spike-first ruling in `docs/superpowers/specs/2026-07-28-284-byok-design.md`.
- **Spike conclusion: confirmed unavailable.** Cloudflare Containers do not support `NET_ADMIN` / iptables-style egress policy — official docs state containers run without root and "do not support iptables manipulation." No capability/security-context field exists on `[[containers]]` in the Wrangler config schema.
- Fallback (pre-authorized by the spec) is documented in `docs/ops/cloudflare-hardening.md` §6: the application-layer guard (`egress_guard` + guarded `httpx` factory, Task 1/#458) is the sole enforced control; residual risk (T12: future code path bypasses the guarded factory) is recorded, not silently accepted.
- Evaluated a process-level Python `socket` guard as an optional stand-in; **declined** — reasoning documented (global blast radius across every outbound call incl. Postgres/catalog/model providers, no credible way to verify the exact bypass scenario without a real container network namespace, and coupling to the still-unmerged T1/T3 branches risks two guards drifting apart).
- Recorded the upgrade path (re-open this section, add the AC + integration test) if Cloudflare ships real network-layer control later.
- No permanently-skipped test was left behind — Task 7's three ACs are satisfied by this documented acceptance, per the OQ-5 ruling, not by a shipped-or-skipped runtime test.

## Test plan

- [x] Docs-only change; pre-commit hooks (trailing-whitespace, secret scan) passed clean
- [x] No markdownlint tooling configured in this repo (verified — no config, package not installed); skipped
- [ ] Not merging — leaving open per instructions for review

## Summary by CodeRabbit

* **Documentation**
  * Expanded the Cloudflare Hardening Runbook with documented egress network policy controls.
  * Clarified that application-level safeguards block restricted outbound destinations when network-layer controls are unavailable.
  * Documented credential redaction, proxy-bypass prevention, rate limiting, residual risks, and evaluated alternatives.
  * Added guidance for revisiting network-layer enforcement and integration testing if platform capabilities change.

### DIFF
--- a/apps/agent/AGENTS.md
+++ b/apps/agent/AGENTS.md
@@ -84,7 +84,17 @@ Anitabi (`api.anitabi.cn`) + Bangumi (`api.bgm.tv`) share Bangumi.tv subject IDs
 
 - **httpx only** — aiohttp is retired (F7). **One shared `httpx.AsyncClient` per client**, created
   lazily and closed via the FastAPI lifespan `aclose()` (`agent/interfaces/fastapi_service.py`) — never per-request.
-  Leave `trust_env` at httpx's default (`True`) so proxy/CA env vars are respected.
+  Leave `trust_env` at httpx's default (`True`) for this **shared lifespan client** so proxy/CA env
+  vars are respected. **This does not apply to BYOK/egress-guarded clients** (#284 Task 1/T13):
+  those must be built via `egress_transport.build_guarded_async_client`, which sets
+  `trust_env=False` and no `mounts`/proxy on purpose — a proxy env var would silently defeat the
+  connect-time IP pinning that closes the DNS-rebinding/TOCTOU window (T4) and the redirect-bypass
+  window (T5). **T12 note:** the guarded factory is the *only* sanctioned way to build an
+  outbound HTTP client for a user-influenceable destination (BYOK `base_url`, and any future
+  user-controlled egress) — any new outbound call site that constructs its own `httpx.AsyncClient`
+  instead of going through the factory recreates the SSRF/T13 bypass this convention exists to
+  close. See `docs/ops/cloudflare-hardening.md` §6 for why this code-review convention, not a
+  container network policy, is the actual enforcement point for T12.
 - **Status-based retry** — classify by **status code, never by URL/substring**: 5xx, transport errors,
   and transient 4xx (408/429) retry with backoff; other 4xx raise immediately (`agent/clients/catalog_client.py`).
 - **Observability = logfire only** (F8). Never hand-roll OpenTelemetry or add `opentelemetry-api|sdk`
--- a/apps/agent/AGENTS.md
+++ b/apps/agent/AGENTS.md
@@ -93,8 +93,11 @@ Anitabi (`api.anitabi.cn`) + Bangumi (`api.bgm.tv`) share Bangumi.tv subject IDs
   outbound HTTP client for a user-influenceable destination (BYOK `base_url`, and any future
   user-controlled egress) — any new outbound call site that constructs its own `httpx.AsyncClient`
   instead of going through the factory recreates the SSRF/T13 bypass this convention exists to
-  close. See `docs/ops/cloudflare-hardening.md` §6 for why this code-review convention, not a
-  container network policy, is the actual enforcement point for T12.
+  close. See `docs/ops/cloudflare-hardening.md` §6 — `RuntimeContainer.deniedHosts` covers T12 at
+  the Worker's URL-hostname layer for plain-HTTP requests naming a denied IP/hostname literal
+  directly (not DNS rebinding, and not HTTPS); this code-review convention is the enforcement
+  point for everything that layer does not reach — HTTPS to a private/link-local/CGNAT address,
+  and any hostname that only *resolves* to one.
 - **Status-based retry** — classify by **status code, never by URL/substring**: 5xx, transport errors,
   and transient 4xx (408/429) retry with backoff; other 4xx raise immediately (`agent/clients/catalog_client.py`).
 - **Observability = logfire only** (F8). Never hand-roll OpenTelemetry or add `opentelemetry-api|sdk`
--- a/docs/ops/README.md
+++ b/docs/ops/README.md
@@ -10,7 +10,7 @@ Use this directory for:
 
 Current canonical docs:
 - `deployment.md` — Cloudflare Workers + Containers deployment runbook (topology, auth flow, env boundaries, rollback)
-- `cloudflare-hardening.md` — WAF rate limiting, prompt-injection filtering, rollback for edge rules
+- `cloudflare-hardening.md` — WAF rate limiting, prompt-injection filtering, rollback for edge rules, container-level egress network policy (#284 Task 7)
 - `preview.md` — per-PR preview environments (label-gated CF preview URLs + Neon branch-per-PR + auto-teardown)
 - `anonymous-session-purge.md` — scheduled anonymous-session retention sweep (cron cadence, required secret, reading a run)
 
--- a/docs/ops/cloudflare-hardening.md
+++ b/docs/ops/cloudflare-hardening.md
@@ -209,112 +209,222 @@ no live deploy was performed — deploys stay CI/CD-only per the repo's no-local
 **This clause holds: NET_ADMIN / iptables-style egress policy is confirmed unavailable on
 Cloudflare Containers today.**
 
-### Spike, clause 2 (correction — the first pass of this doc got this wrong): a declarative CIDR denylist IS available, and does not require NET_ADMIN
+### Spike, clause 2 (corrected twice now — see both corrections below): a platform-enforced URL-hostname denylist IS available, and does not require NET_ADMIN — but it is a glob matcher, not a CIDR parser
 
-An earlier version of this section additionally claimed the platform's only egress-shaping
-primitive was HTTP(S)-scoped Worker request routing (`outboundByHost`), and concluded from that
-that clause 2 of OQ-5 was also unmet. **That conclusion was wrong** and is corrected here after
-independent re-verification against `https://developers.cloudflare.com/containers/platform-details/outbound-traffic/`
-(checked 2026-07-29) and the `@cloudflare/containers@0.3.7` type definitions actually vendored in
-this repo (`node_modules/@cloudflare/containers/dist/lib/container.d.ts` — checked directly, not
-inferred from docs prose):
+An earlier version of this section claimed the platform's only egress-shaping primitive was
+HTTP(S)-scoped Worker request routing (`outboundByHost`), and concluded that clause 2 of OQ-5 was
+also unmet. **That was wrong.** The next revision corrected it to "`deniedHosts` accepts IP CIDR
+ranges" — citing the official docs' own example, `deniedHosts = ["some-nefarious-website.com",
+"141.101.64.0/18"]` — and shipped a `string[]` of literal CIDR strings (`"10.0.0.0/8"`, etc.).
+**That was also wrong, and worse: it shipped as a silent no-op.**
 
-- **`deniedHosts`** is a `Container`-class instance property (same class `RuntimeContainer`
-  already extends) that accepts **hostnames, hostname globs, and IP CIDR ranges** — the official
-  example is literally `deniedHosts = ["some-nefarious-website.com", "141.101.64.0/18"]`.
-- Per the vendored type declaration's own doc comment: *"Denied hosts are blocked unconditionally,
-  even when `enableInternet` is `true` or a catch-all outbound handler is set"* — i.e. this is
-  enforced **before** any outbound handler runs, unconditionally, and does **not** require
-  `enableInternet: false` (which would additionally restrict egress to ports 80/443 + DNS and
-  break the direct `asyncpg` Postgres hop — confirmed as a real constraint, which is why the
-  `deniedHosts`-only variant, not the `enableInternet: false` variant, is the one implemented
-  below).
-- `allowedHosts` is separately documented as *"a deny-by-default allowlist"* when set — i.e.
-  default-deny-then-allow is expressible on this platform; we are not using `allowedHosts` here
-  (an allowlist would need to enumerate every legitimate provider hostname and would be a much
-  larger, riskier change than a denylist of well-known non-routable ranges), but its existence is
-  further confirmation that clause 2 of OQ-5 is met.
+A second review round read the vendored **implementation**, not just the type declaration or the
+docs prose (`node_modules/@cloudflare/containers/dist/lib/container.js`,
+`@cloudflare/containers@0.3.7`), and found:
 
-**Corrected conclusion: OQ-5's clause 2 is met. The platform can express an egress policy (a
-declarative, platform-enforced CIDR denylist) without NET_ADMIN. The pre-authorized fallback
-("document that the application layer is the sole control") does *not* trigger, because its own
-precondition — the platform cannot express an egress policy at all — is false. Task 7 ships as
-an actual implemented policy, not a documented-acceptance closure.**
+```js
+function simpleGlobMatch(pattern, value) {
+    const parts = pattern.split('*');
+    if (parts.length === 1) return pattern === value;
+    // ... prefix/suffix/substring matching on `parts`, no numeric parsing at all
+}
+function matchesHostList(hostname, patterns) {
+    return patterns.some(pattern => simpleGlobMatch(pattern, hostname));
+}
+```
+
+`deniedHosts`/`allowedHosts` are matched against the **request URL's hostname string** with a
+plain `*`-wildcard string matcher — there is no `/`-suffix (CIDR) parsing anywhere in this
+function, and it never resolves DNS, so it cannot match a hostname's *resolved* IP either. The
+literal string `"10.0.0.0/8"` only ever matches a hostname that is exactly the eleven characters
+`10.0.0.0/8` — it does not match `10.0.0.1`, `10.5.3.200`, or anything else. The previous revision
+of `worker/entry.ts` shipped exactly that: five CIDR-notation strings that matched nothing,
+ever — a complete no-op, caught only because a reviewer decided that "the type is `string[]`"
+does not establish "the semantics are CIDR" and went and read the matcher's actual source instead
+of trusting the docs' prose example. **The type of a config field is not proof of what a runtime
+does with it — the implementation is the ground truth,** and that lesson is recorded here on
+purpose, not smoothed over.
+
+The corrected, verified claim: the Container-runtime egress control that exists without NET_ADMIN
+is a **string/glob denylist matched against the request URL's hostname**, not a CIDR-aware network
+filter. It is real, it is enforced before any outbound handler
+(`container.js:203`, `if (deniedHosts && matchesHostList(hostname, deniedHosts)) return new
+Response('Origin is disallowed', { status: 520 })` — see the "What actually happens on a match"
+note below), and it does not require `enableInternet: false`. But its coverage claim has to be
+stated in glob terms, not CIDR terms — see "What is implemented" below for exactly what that
+buys.
+
+**Corrected conclusion: OQ-5's clause 2 is met — the platform can express *a* URL-hostname-layer
+egress policy without NET_ADMIN — but it is narrower than "an egress policy" reads in the abstract.
+It is not a CIDR/network-layer filter, and it does not see resolved IPs.** The pre-authorized
+fallback ("document that the application layer is the sole control") still does not fully trigger,
+because this is a real, additional, platform-enforced control layer — but the application-layer
+guard (`egress_guard`/`GuardedAsyncTransport`) remains the layer that actually understands
+IP/network semantics (DNS resolution, `ipaddress` classification, DNS-rebinding defense). Task 7
+ships as an implemented (glob-based, URL-hostname-layer) policy plus the pre-existing
+application-layer guard — not a documented-acceptance closure, and not a claim of network-layer
+enforcement either.
 
 ### What is implemented
 
-`RuntimeContainer.deniedHosts` (`worker/entry.ts`) is set from `DENIED_EGRESS_CIDRS`
+`RuntimeContainer.deniedHosts` (`worker/entry.ts`) is set from `DENIED_EGRESS_HOSTS`
 (`worker/containerEnv.ts`, split out for the same Node-import-chain reason as
-`buildContainerEnvVars` — see that file's header comment):
+`buildContainerEnvVars` — see that file's header comment). It is a set of dotted-decimal glob
+prefixes and exact hostnames — **not CIDR strings** — chosen to be the glob-equivalent of the
+spec's target ranges when a request URL's hostname is already a bare IPv4 literal (the common
+shape for cloud-metadata SSRF, and the exact shape of the spec's Task 7 AC):
 
 ```ts
-export const DENIED_EGRESS_CIDRS = [
-  "10.0.0.0/8",       // RFC1918
-  "172.16.0.0/12",    // RFC1918
-  "192.168.0.0/16",   // RFC1918
-  "169.254.0.0/16",   // link-local (AWS/Azure/GCP IMDS)
-  "100.64.0.0/10",    // CGNAT (Alibaba/Tencent metadata)
+export const DENIED_EGRESS_HOSTS = [
+  "10.*",                       // RFC1918 10.0.0.0/8 — glob-exact equivalent
+  "172.16.*", ..., "172.31.*",  // RFC1918 172.16.0.0/12 — 16 generated entries, one per octet
+  "192.168.*",                  // RFC1918 192.168.0.0/16 — glob-exact equivalent
+  "169.254.*",                  // link-local 169.254.0.0/16 — glob-exact equivalent (AWS/Azure/GCP IMDS)
+  "100.64.*", ..., "100.127.*", // CGNAT 100.64.0.0/10 — 64 generated entries, one per octet
+  "100.100.100.200",            // Alibaba/Tencent metadata (exact; already covered above, kept explicit)
+  "192.0.0.192",                // Oracle Cloud Infrastructure metadata
+  "metadata.google.internal",   // GCP metadata hostname literal — glob ranges above cannot match a hostname
 ];
 ```
 
-Pinned by `worker/containerEnv.test.ts` (run via `pnpm run test:worker`), which asserts both the
-exact CIDR list and that it covers every address the spec's Task 7 AC error-path names
-(`169.254.169.254`, `100.100.100.200`, `10.0.0.1`) while not covering a public address
-(`1.1.1.1`).
+The 16 and 64 per-octet entries are generated (`Array.from({ length: 16 }, (_, i) => \`172.${16 +
+i}.*\`)`, etc.), not hand-typed, so the source stays short even though the effective list is long;
+this is what makes full-range coverage (rather than a metadata-IP-only subset) the practical
+choice here.
 
-### Scope and an honest limit: HTTP is covered today; HTTPS interception is a deferred, evaluated cost
+**What actually happens on a match:** `ContainerProxy.fetch` (the `WorkerEntrypoint` that
+performs this check — see the `ContainerProxy` export note below) returns a synthesized
+**`520 Origin is disallowed`** HTTP response. Nothing is "refused at the network layer" in the
+sense of a socket-level RST or connection timeout — the platform's egress proxy answers the
+request itself, and the container never gets a TCP connection to the target. Functionally
+equivalent for SSRF purposes (no bytes reach the target, no bytes come back), but a future
+engineer debugging a `520` from inside the container should know this policy is what produced it.
 
-Per the same vendored type declaration, `deniedHosts` filtering is applied as part of the
-Container runtime's outbound-HTTP interception, which is unconditional for **plain HTTP**. HTTPS
-traffic is only decrypted/inspected for filtering when `interceptHttps: true` is additionally set
-(`applyOutboundInterception`'s doc comment: *"When `interceptHttps` is enabled, also applies
-HTTPS interception"*). We evaluated turning this on and **did not**, for one concrete, evidenced
-reason: `interceptHttps` MITMs all outbound HTTPS with an ephemeral per-instance CA
-(`/etc/cloudflare/certs/cloudflare-containers-ca.crt`), which the container's Python process must
-be configured to trust (Python's `httpx`/`certifi` bundle does not trust arbitrary system CAs by
-default). Flipping this on without first wiring that trust — and verifying it in a real deploy,
-which this task cannot do — would silently break **every** production HTTPS call (MiMo, the
-DeepSeek fallback, any BYOK provider), not just close a gap. That is a materially worse failure
-mode than the gap it would close, given:
+Pinned by `worker/containerEnv.test.ts` (run via `pnpm run test:worker`), which ports the real
+`simpleGlobMatch`/`matchesHostList` algorithm (rather than a hand-rolled IP-in-CIDR helper — the
+mistake from the previous revision) and asserts against it: every spec Task 7 AC error-path
+address is matched, the two extra metadata endpoints are matched, three representative public
+addresses are not matched, the full 16- and 64-entry octet ranges are matched at their exact
+boundaries (and the addresses just outside those boundaries are not), and an emptied denylist
+fails the coverage assertions (a mutation guard against the exact silent-no-op failure mode this
+correction exists because of). A canary test additionally asserts the vendored
+`container.js` source still contains the exact `simpleGlobMatch`/`matchesHostList` function
+signatures this port is based on, so a package upgrade that changes the algorithm is flagged
+rather than silently invalidating the port.
 
-- the exact AC error-path scenario in the spec uses `http://169.254.169.254/` (plain HTTP) — the
-  shipped `deniedHosts` list already blocks this, and real cloud-metadata IMDS endpoints
-  (AWS/Azure/GCP `169.254.169.254`, Alibaba/Tencent `100.100.100.200`) are HTTP-only in practice,
-  so the highest-value SSRF targets are covered without `interceptHttps`;
-- the residual gap is an HTTPS request to a private/link-local/CGNAT IP specifically — a narrower
-  bypass than the one this section closes, and one still caught by the application-layer guard
-  (`egress_guard`/`GuardedAsyncTransport`) for the BYOK path specifically.
+**Required for the policy to run at all — `ContainerProxy` export.** `applyOutboundInterception`
+(`container.js:1207`) hard-throws at container start if `ctx.exports.ContainerProxy` is undefined:
 
-**Follow-up, not required for Task 7's closure:** turning on `interceptHttps` needs its own
-spec'd rollout (Python CA-trust wiring + a staging verification pass), tracked as a Task 7
-follow-up rather than bundled into this docs/config PR.
+```
+if (ctx.exports.ContainerProxy === undefined) {
+    throw new Error('ctx.exports.ContainerProxy is undefined, export ContainerProxy from the
+    containers package in your worker entrypoint');
+}
+```
+
+`worker/entry.ts` now re-exports it (`export { ContainerProxy } from "@cloudflare/containers";`),
+matching the pattern in Cloudflare's own `Container` class reference example. This is not a
+regression introduced by this change — the repo already exercised `outboundByHost` for
+`catalog.internal` before this PR, which also depends on the same interception machinery — but it
+was never verified as present, and `deniedHosts` makes it load-bearing for the first genuinely
+security-relevant use of this mechanism. `worker/entry.test.ts` asserts (via a source-text match,
+since `entry.ts`'s `@cloudflare/containers` import chain cannot be loaded under plain `node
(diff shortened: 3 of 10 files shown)


lifeodyssey (commented):
## Fable seat review — PR #478 (#284 Task 7, docs-only)

Verified against spec rev `efc0abaa`, base `feat/frontend-rebuild`, and merged PRs #458/#474/#477/#451. Codecov: N/A (doc-only). Evals: skipped (no agent behavior change).

### Verified accurate
- T12/T13 quotes match the spec threat table verbatim; OQ-5 ruling (spike-first, fallback pre-authorized, "no permanently-skipped test") faithfully reproduced.
- The three Task 7 ACs are correctly mirrored in the upgrade path, including the exact test path `apps/agent/agent/tests/integration/test_egress_network_policy.py`.
- `egress_guard.validate_base_url()` exists (`apps/agent/agent/infrastructure/egress_guard.py:238`); route-level 400 verified by `test_byok_ssrf_route.py:59`.
- Task 9 rate limiting is merged (#451) — claim accurate. `wrangler.toml` `outboundByHost["catalog.internal"]` comments and Dockerfile non-root `appuser` both exist as cited. Cloudflare no-iptables/no-root evidence consistent with the platform docs.
- Upgrade path is executable: concrete changelog URLs, concrete file targets, concrete test shape, and the T12 downgrade step.

### P1 findings

**P1-1 (docs / dangling reference).** `docs/superpowers/specs/2026-07-28-284-byok-design.md` — the section's primary reference — does **not exist on the base branch**. The spec lives only on `docs/spec-284-byok-v2` / `feat/284-t68-byok-ui-journey`; no open PR delivers it to `feat/frontend-rebuild`. If #478 merges first, the audit entry point's central citation is a dead path. Fix: land the spec on base first, or note the interim location.

**P1-2 (tdd/security — T12 mitigation incomplete vs spec).** Spec residual-risk table (line 1151) prescribes the T12 mitigation as Task 7's second line **"plus an architecture note in `apps/agent/AGENTS.md`"**. With the second line now void, that note is load-bearing — and it does not exist. Worse, `apps/agent/AGENTS.md` (lines 85–87) currently instructs the opposite: *"Leave `trust_env` at httpx's default (`True`) so proxy/CA env vars are respected."* A future contributor following the agent guide would recreate exactly the T13 bypass this section claims is closed. Since "PR review is the enforcement point" is the doc's entire thesis, either record this gap in §6's residual-risk paragraph or file the AGENTS.md follow-up before closing Task 7.

**P1-3 (docs / stale claim).** "Coupling to an unmerged dependency" states Task 1/Task 3 are "still in flight on unmerged branches (`feat/284-t1-ssrf-guard`, `feat/284-t3-byok-model`) at the time of this spike." #458 (Task 1) merged 2026-07-28T17:13Z and #477 (Task 3) merged 2026-07-29T04:06Z — both before/at the spike's own evidence date (2026-07-29). Decline-reason 3 partially rests on a false premise (reasons 1–2 stand on their own). Reword or drop reason 3.

### P2 findings

**P2-1 (naming/accuracy — factory misattributed).** The guarded factory is `build_guarded_async_client` in `apps/agent/agent/infrastructure/egress_transport.py`; `byok_models.py` is a consumer that wires all three provider families to it. `trust_env=False` is set inside the transport module, not in `byok_models.py`. As an audit entry, point at the real file/symbol.

**P2-2 (completeness — control list undersells what is enforced).** `GuardedAsyncTransport` also enforces: pinned-IP connection with Host-header rewrite (DNS-rebinding defense), redirect refusal (`REDIRECT_REFUSED`), and opt-out from global Logfire/OTel httpx instrumentation (#474, folded into the transport via `_exclude_from_httpx_instrumentation`). None appear in the "What actually enforces…" list. Add them — the strongest runtime controls should not be invisible to a future auditor.

**P2-3 (minor accuracy).** (a) State explicitly that all three families (OpenAI-compatible, Anthropic, Gemini — #477) route through the guarded factory; that's what makes "only sanctioned way" complete. (b) "both live model providers" / "two live provider hops": prod is currently MiMo-only (DeepSeek fallback disabled); the spec's Task 7 null/empty AC says "the MiMo provider call". Align the count or say "configured providers".

### Verdict

```json
{
  "verdict": "request_changes",
  "eval_results": { "test_eval": "skipped" },
  "quality_ratchet": { "ac_total": 3, "ac_with_test": 0, "missing": [], "note": "All three Task 7 ACs closed via OQ-5 documented-acceptance ruling; test-free closure is spec-authorized, upgrade path preserves the original AC test shape." }
}
```

P1-1 and P1-3 are line edits; P1-2 needs either a residual-risk sentence or a tracked follow-up. None require code.

— Fable seat (independent; no other reviewer's conclusions consulted)

lifeodyssey (commented):
## Review: request_changes

Docs-only PR, but the document is a **decision record that closes a security task and accepts a residual risk (T12)**. The load-bearing spike conclusion does not survive independent verification.

---

### P0 — The spike answered half the OQ-5 question; the fallback precondition is not met

`docs/ops/cloudflare-hardening.md:185-218`

OQ-5 asks whether the runtime "grants `NET_ADMIN` **/ can express an egress policy at all**". The doc verifies the first clause (correctly) and then declares the second clause unavailable on the basis that outbound interception is "HTTP(S)-scoped request interception … not a network-layer firewall … configured per-hostname (allow-list of one), not a default-deny-then-allow egress policy."

That is contradicted by Cloudflare's current docs (`https://developers.cloudflare.com/containers/platform-details/outbound-traffic/`, last updated 2026-04-21 — verified today):

- **`deniedHosts`** on the `Container` class accepts **hosts *and IP CIDRs*** — the official example is literally `deniedHosts = ["some-nefarious-website.com", "141.101.64.0/18"]`. A declarative `["10.0.0.0/8","172.16.0.0/12","192.168.0.0/16","169.254.0.0/16","100.64.0.0/10"]` is exactly the Task 7 scope, expressible today, additive, with no NET_ADMIN.
- **`allowedHosts`** is documented as **"a deny-by-default allowlist"** — i.e. default-deny-then-allow, the precise thing the doc says does not exist.
- **`enableInternet = false`** — "Set to `false` for sandboxed environments where you want to intercept or block all outbound traffic"; when false "only traffic you explicitly allow … through `allowedHosts` or outbound handlers can leave the container."
- `deniedHosts`/`allowedHosts` are **evaluated before any outbound handler**, so this is enforcement, not routing.

Consequently the sentence "NET_ADMIN / iptables-style egress policy is confirmed unavailable" is true but **not the finding that authorizes the fallback**; "the OQ-5 pre-authorized fallback triggers" (line 217) does not follow, and Task 7's three ACs are therefore **not** closed by documented acceptance.

Required: re-run the spike arm on `deniedHosts` / `allowedHosts` / `enableInternet` and either (a) implement the CIDR denylist + the AC integration test, or (b) document a concrete, evidenced reason it is not viable. Honest costs to weigh in (b), not to hand-wave: `interceptHttps = true` + `ContainerProxy` export + container trust of `/etc/cloudflare/certs/cloudflare-containers-ca.crt` are prerequisites for HTTPS interception; `enableInternet=false` restricts egress to ports 80/443 + DNS, which would break the direct `asyncpg` Postgres hop (so the `deniedHosts` variant, not the `enableInternet` variant, is the candidate); and whether `deniedHosts` covers non-HTTP egress is undocumented and needs its own probe.

---

### P1 — Decline reason #3 is factually stale

`docs/ops/cloudflare-hardening.md:274-282` — "still in flight on unmerged branches (`feat/284-t1-ssrf-guard`, `feat/284-t3-byok-model`)". Both are merged into `feat/frontend-rebuild`: **#458** (2026-07-28T17:13Z), **#474** (2026-07-29T02:27Z), **#477** (2026-07-29T04:06Z). One of the three stated reasons for declining the socket guard has evaporated; the decline needs re-justification on the two surviving reasons, or the paragraph rewritten to the current tree state.

### P1 — Dangling spec reference (DOCS_POLICY)

`docs/superpowers/specs/2026-07-28-284-byok-design.md` is cited three times but **does not exist on the base branch** `feat/frontend-rebuild`, nor on this PR's head. It exists only on the unmerged, PR-less branch `docs/spec-284-byok-v2`. A runbook in `docs/ops/` must not be the only merged artifact pointing at an unmerged path. Land the spec first, or drop the path and inline the OQ-5 ruling text.

### P1 — Wrong file attributed to the guarded factory

`docs/ops/cloudflare-hardening.md:232` — "The guarded `httpx` client factory (`byok_models.py`, Task 3)". The `trust_env=False`, no-`mounts` client is built in `apps/agent/agent/infrastructure/egress_transport.py` (lines 59 and 157; the T13 rationale comment is at line 145). `byok_models.py` is the model-resolution layer. Since this line names the single sanctioned construction path that is the whole T12 mitigation, pointing a future contributor at the wrong file defeats the mitigation. Fix to `apps/agent/agent/infrastructure/egress_transport.py`.

### P2 — Wrangler schema enumeration is presented as exhaustive and is not

`docs/ops/cloudflare-hardening.md:198`. The real `[[containers]]` schema today is `image`, `class_name`, `instance_type`, `max_instances`, `name`, `image_build_context`, `image_vars`, `rollout_active_grace_period`, `rollout_step_percentage`, `ssh`, `wrangler_ssh` (deprecated), `authorized_keys`, `constraints.regions`, `constraints.jurisdiction`. The load-bearing claim (no capability / security-context field) holds; the list does not. Suggested replacement:

> — the schema exposes only image/build, placement, rollout, SSH and `constraints` settings (`image`, `class_name`, `instance_type`, `max_instances`, `name`, `image_build_context`, `image_vars`, `rollout_*`, `ssh`, `authorized_keys`, `constraints.*`); **no capability, security-context, or network-policy field exists** (`https://developers.cloudflare.com/workers/wrangler/configuration/#containers`, checked 2026-07-29).

### P2 — Misquote inside quotation marks

`docs/ops/cloudflare-hardening.md` quotes *"Containers operate without root privileges"*. The FAQ says **"Containers run without root privileges"**. The iptables quote on the same bullet is verbatim-correct. In an evidence section, quoted material must be verbatim — change `operate` to `run`.

### P2 — Control inventory omits the T4/T5 exemption semantics

The four-item "what actually enforces this today" list stops at T1/T2/T3/T9. Item 4 claims Task 9 rate limiting "bounds the blast radius … independent of whether that target is blocked" — that claim is materially conditioned by the T4/T5 BYOK quota-exemption work (`feat/284-t45-exemption-probe`, unmerged). Either scope item 4 ("for server-key traffic; BYOK exemption semantics land in T4/T5 and will change this bound") or add a fifth item once T4 lands.

### P2 — Docs index not updated

`docs/ops/README.md:13` still describes this file as "WAF rate limiting, prompt-injection filtering, rollback for edge rules". Add the egress-policy section to that one-liner.

### P2 — CI: 6 red checks, none content-related

`Analyze (actions|javascript-typescript|python)`, `gitleaks`, `osv-scanner`, `changes` all fail with `Resource not accessible by integration` (token/permission, not the diff). Not caused by this PR, but they need a rerun or an explicit waiver before this can land.

---

### What is right

- The NET_ADMIN / iptables claim itself is **verified correct and verbatim-accurate** against the Containers FAQ, and the "no capability field on `[[containers]]`" claim is correct.
- The no-permanently-skipped-test constraint is honored — nothing was stubbed or `skip`-marked.
- The socket-guard decline reasons #1 (global monkeypatch blast radius across `asyncpg`/catalog/providers) and #2 (a local-pytest test would validate a mock, not the container network namespace) are sound on their own, as is the pointer to scoping any future guard inside `egress_guard` as an `httpx` transport hook rather than a global patch.
- Upgrade-path section is actionable and correctly names the changelog sources to re-check.

**Verdict: request_changes** — blocked on the P0 (re-spike `deniedHosts`/`allowedHosts`/`enableInternet` before closing Task 7 as accepted risk) plus the three P1s.

lifeodyssey (commented):
## Re-review (HEAD `4932aa94`): approve

All four P0/P1s and all three P2s from the first round are resolved, and the correction is done honestly — the doc names its own earlier error rather than quietly rewriting it. Verified each claim independently against the vendored `@cloudflare/containers@0.3.7` source, not the docs prose.

### Verified

**Denylist contents** (`worker/containerEnv.ts:DENIED_EGRESS_CIDRS`) — `10/8`, `172.16/12`, `192.168/16`, `169.254/16`, `100.64/10`. Correct and complete against the Task 7 scope; covers all three spec AC error-path addresses (`169.254.169.254`, `100.100.100.200`, `10.0.0.1`) and excludes public space.

**Type-declaration citations — accurate, verbatim.** `container.d.ts:123` reads *"Denied hosts are blocked unconditionally, even when `enableInternet` is true or a catch-all outbound handler is set"* — the doc quotes it correctly. Confirmed in the implementation too: `container.js:202-205` places the `deniedHosts` check as step 1 of `ContainerProxy.fetch`, ahead of `allowedHosts`, `outboundByHost` overrides, static `outboundByHost`, and both catch-all paths. "Enforced before any outbound handler, independent of `enableInternet`" is a true statement about this build.

**`enableInternet: true` retained — correct call.** `enableInternet = false` restricts egress to 80/443 + DNS, which would kill the direct `asyncpg` hop. Choosing the `deniedHosts`-only variant is right, and the reasoning in `worker/entry.ts` matches the runtime.

**`interceptHttps` deferral — the reasoning holds.** `applyOutboundInterception` (`container.js:1207-1211`) only calls `interceptOutboundHttps('*', fetcher)` when `interceptHttps` is set, so plain HTTP is covered and HTTPS is not — the doc's scoping is exact, not hedged. The deferral rationale is sound on the merits: enabling MITM interception without first wiring Python CA trust for `/etc/cloudflare/certs/cloudflare-containers-ca.crt` would break every production HTTPS call rather than close a gap, and `certifi` will not pick that CA up on its own. Deferring a change whose failure mode is "all provider traffic dies" until it can be verified in staging is the correct order of operations, and it is documented as a tracked follow-up rather than dropped.

**AC disposition — honestly bounded.** Marking all three ACs "satisfied by design / by unit pin, not yet verified against a live deployed instance", and the error path explicitly "satisfied for plain HTTP, not for HTTPS", is the right disposition given the no-local-deploy rule. No fabricated closure, no permanently-skipped test.

**P1/P2 carry-over — all landed.** Stale-branch reason removed with an explicit note (#458/#474/#477 merged) and the decline re-grounded on two reasons that stand alone; both specs now in tree so the `2026-07-28-284-byok-design.md` reference resolves; factory correctly re-attributed to `egress_transport.py` with `GuardedAsyncTransport`/`build_guarded_async_client` named; schema enumeration corrected and no longer exhaustive-by-implication; the "run without root" quote fixed to verbatim; T4/T5 conditioning added to control-inventory item 5; `docs/ops/README.md:13` updated. The `apps/agent/AGENTS.md` `trust_env` carve-out is a good catch beyond what I asked for — the prior text would have actively instructed a contributor to recreate T13.

---

### P1 — `ContainerProxy` is not exported from the Worker entrypoint; add it to the top of the live-verification checklist

`deniedHosts` is not enforced by a kernel filter — it is enforced *inside* `ContainerProxy.fetch` (`container.js:196-205`), a `WorkerEntrypoint` that must be reachable via `ctx.exports`. `applyOutboundInterception` hard-throws when it is not:

```
if (ctx.exports.ContainerProxy === undefined) {
    throw new Error('ctx.exports.ContainerProxy is undefined, export ContainerProxy from the containers package in your worker entrypoint');
}
```

and that promise **is** awaited (`container.js:1154-1155`). `worker/entry.ts` exports `DOQueueHandler`, `DOShardedTagCache`, `EdgeGuard`, and `default` — no `ContainerProxy`. Cloudflare's outbound-traffic page carries the same requirement as a Note attached to the `allowedHosts`/`deniedHosts` section.

Not treating this as a merge blocker, because it is **not a regression from this PR**: `container.js:301-308` already sets `usingInterception = true` when `ctor.outboundByHost !== undefined`, which is true today via `RuntimeContainer.outboundByHost["catalog.internal"]`. So the repo either already satisfies this some other way (compat-date-provided `ctx.exports`) or already has a latent problem on the catalog hop — pre-existing either way.

But it is decisive for whether the shipped denylist enforces anything, so: make **"confirm `ContainerProxy` is exported / reachable via `ctx.exports`, and that container start does not throw"** step 1 of the live-verification pass, ahead of the three ACs. If it turns out to be missing, the fix is one line in `worker/entry.ts`:

```ts
export { ContainerProxy } from "@cloudflare/containers";
```

### P2 — `deniedHosts` also promotes the container from per-host to intercept-all mode; worth one sentence

`shouldInterceptAllOutbound()` (`container.js:1121-1126`) returns true whenever `effectiveDeniedHosts !== undefined`, and `hasInterceptAllRegistration` makes the promotion sticky "until the instance restarts". Today the container runs in per-host mode (only `outboundByHost`); after this change every plain-HTTP outbound takes a Workers-runtime hop through `ContainerProxy` instead of going direct. `asyncpg` is unaffected (non-HTTP, `interceptAllOutboundHttp` only), so the `worker/entry.ts` comment's Postgres claim is right — but the change is broader than "additive denylist" reads. Suggested addition to the "What is implemented" section:

> Setting `deniedHosts` also promotes the container from per-host interception to intercept-all for plain HTTP (`shouldInterceptAllOutbound`), so every plain-HTTP outbound now routes through `ContainerProxy` in the Workers runtime. Non-HTTP egress (`asyncpg`'s Postgres connection) is unaffected, and is correspondingly **not** covered by the denylist.

That second clause is worth stating on its own: a raw non-HTTP socket to a denied CIDR is not blocked by this control.

### P2 — "refused at the network layer" vs. what actually happens

The spec AC says the connection is "refused at the network layer". In practice `ContainerProxy.fetch` synthesizes `new Response('Origin is disallowed', { status: 520 })` — no connection is refused; an HTTP error is returned. Equivalent for SSRF purposes (nothing reaches the target, no data returns), but a future reader debugging a 520 should not have to rediscover this. Suggest adding to the AC disposition:

> Note: enforcement returns HTTP `520 Origin is disallowed` from the platform egress proxy rather than refusing the connection at the socket layer — the spec AC's "refused at the network layer" is satisfied in effect (nothing reaches the target), not literally. A `520` from a container HTTP call to a private range is this policy firing.

### P2 — one stale sentence in `apps/agent/AGENTS.md`

The added block ends: "See `docs/ops/cloudflare-hardening.md` §6 for why this code-review convention, **not a container network policy**, is the actual enforcement point for T12." That was true of the previous revision and is now contradicted by the shipped `deniedHosts`. Suggest:

> See `docs/ops/cloudflare-hardening.md` §6 — `RuntimeContainer.deniedHosts` covers T12 at the platform layer for plain HTTP; this code-review convention is the enforcement point for the remaining HTTPS-to-private-IP residual.

---

**Verdict: approve.** Task 7 closes as an implemented policy with an honestly-bounded AC disposition and a named follow-up, which is the right outcome. The P1 is a verification-ordering item for the staging pass, not a merge blocker; the three P2s are wording/scoping precision. CI red is the known token-permission infra issue on your side.

lifeodyssey (commented):
## Fable seat re-review — PR #478 @ 4932aa94 (direction change: real deniedHosts implementation)

### Prior findings: all resolved
- P1-1: Both approved specs (#284 rev efc0abaa content, #273) are now in the tree; the §6 citation resolves.
- P1-2: apps/agent/AGENTS.md trust_env section correctly split (shared lifespan client True / guarded client False), with the T12 convention note and pointer to §6. Matches the spec's residual-risk table requirement.
- P1-3: Stale "unmerged branches" claim explicitly corrected; socket-guard decline rewritten to the two reasons that stand independently.
- P2-1/2/3: egress_transport.py attribution fixed; control inventory now includes pinned-IP + Host rewrite (T4), redirect refusal (T5), instrumentation exclusion (#474); three-family (#477) wiring stated; MiMo-only wording corrected.

The doc is also honest where it matters: AC disposition marks live verification as "not yet verified", and the HTTP-vs-HTTPS (interceptHttps) limit is correctly derived from the vendored types.

### New P0 — the shipped CIDR denylist is inert in the vendored library

I verified the actual enforcement path in node_modules/@cloudflare/containers/dist/lib/container.js (v0.3.7, which npm view confirms is also the latest published version):

1. Setting the deniedHosts instance property flips interception to intercept-all (shouldInterceptAllOutbound() returns true when effectiveDeniedHosts !== undefined; interceptAllOutboundHttp(fetcher) is registered). The wiring itself is valid — deniedHosts?: string[] is a public Container instance property.
2. But enforcement is ContainerProxy.fetch -> matchesHostList(hostname, deniedHosts) -> simpleGlobMatch(pattern, value), which supports exact string equality and `*` globs only. There is NO CIDR parsing anywhere in the package — I grepped the entire dist/ for cidr/mask/ip-int logic: zero hits.
3. Therefore simpleGlobMatch("169.254.0.0/16", "169.254.169.254") -> no `*` in pattern -> strict equality -> false. Check 1 never fires; the request falls through to check 8 (enableInternet -> fetch(request)).

**Consequence:** `http://169.254.169.254/` — the spec's literal Task 7 error-path AC — is NOT blocked by the shipped config. All five CIDR entries match nothing, ever. The policy silently blocks zero traffic.

The doc's §6 says the CIDR claim was "checked directly, not inferred from docs prose" — but what was checked directly was the *type* (string[]), and a type cannot establish CIDR semantics. The implementation refutes them for the version actually vendored. If the online containers/platform-details/outbound-traffic page shows the "141.101.64.0/18" example, that page is ahead of (or wrong about) the published SDK — the only matcher that exists in this tree is a glob matcher.

**P0-2 (test anti-pattern, same root):** worker/containerEnv.test.ts implements its OWN ipInCidr() and asserts the list against it. That tests the test's private CIDR matcher, not the library's matching semantics — exactly the "testing our own mock, not the real bypass scenario" failure mode the previous revision of this very doc warned about. A test exercising the library's matcher (or ContainerProxy behavior) would have caught the no-op.

### Fix options (descending preference)
- (a) **Glob patterns the matcher actually supports:** "10.*", "192.168.*", "169.254.*", "172.16.*".."172.31.*" (16 entries), "100.64.*".."100.127.*" (64 entries). Clunky but functional for literal-IP URLs; the unit test should then replicate simpleGlobMatch semantics (or exercise the proxy), not an ipInCidr the library doesn't have.
- (b) **Gate closure on the live staging probe** the doc already commits to — which, as designed, would catch this (curl http://169.254.169.254/ from the container and observe it is NOT refused). Requires reverting the "clause 2 met / Task 7 ships as implemented policy" conclusion until the probe passes.
- (c) **Revert to documented acceptance** until the SDK ships real CIDR support; keep the wiring behind a follow-up.

### P1 — scope of the mechanism is hostname-pattern, not resolved-IP
Even with working patterns, deniedHosts matches the request URL hostname at the proxy (new URL(request.url).hostname), never the resolved address. http://rebind.attacker.com/ resolving to 169.254.169.254 passes any IP-shaped denylist. §6 calls this a "platform-enforced CIDR denylist" / "network policy" — it is an HTTP-proxy hostname filter. One sentence scoping this (the hostname-to-private-IP case remains app-guard-only) keeps the doc audit-accurate.

### P2 — intercept-all side effects undocumented
Before this PR, only catalog.internal was intercepted (per-host). Setting deniedHosts switches to intercept-all: EVERY plain-HTTP outbound from the container now transits the ContainerProxy Worker (subrequest semantics, latency, limits), and the whole HTTP path now depends on the intercept-all registration (ctx.exports.ContainerProxy). Worth one paragraph in §6 and an explicit item in the staging verification pass. (Note worker/entry.ts does not export ContainerProxy; if the existing per-host interception works in prod today this is evidently satisfied some other way, but the intercept-all switch grows the blast radius of that assumption from one host to all HTTP — verify in staging.)

### Verdict

```json
{
  "verdict": "request_changes",
  "findings": [
    {"priority": "P0", "category": "security", "file": "worker/containerEnv.ts", "line": 34, "issue": "CIDR strings in deniedHosts never match: @cloudflare/containers@0.3.7 enforcement is simpleGlobMatch (exact/glob only), no CIDR parsing exists in the package; shipped policy blocks nothing", "fix": "glob patterns (a), gate on live probe (b), or revert to documented acceptance (c)", "test_needed": "integration"},
    {"priority": "P0", "category": "tdd", "file": "worker/containerEnv.test.ts", "line": 22, "issue": "test asserts the list against its own ipInCidr(), not the library's matcher — tests the test's mock, masking the no-op", "fix": "assert via the library's matching semantics", "test_needed": "unit"},
    {"priority": "P1", "category": "security", "file": "docs/ops/cloudflare-hardening.md", "issue": "mechanism is URL-hostname matching, not resolved-IP filtering; hostname-to-private-IP case must be scoped as app-guard-only", "test_needed": "none"},
    {"priority": "P2", "category": "framework", "file": "docs/ops/cloudflare-hardening.md", "issue": "per-host to intercept-all switch (all plain-HTTP via ContainerProxy) undocumented; add to staging verification", "test_needed": "none"}
  ],
  "quality_ratchet": {"ac_total": 3, "ac_with_test": 2, "missing": ["Error-path AC: covered only by a self-referential CIDR test that does not reflect library semantics"]}
}
```

The doc scholarship in this revision is genuinely good — the failure is one unverifiable-from-types assumption at the exact point the platform hands off to the SDK. Options (a)+(b) together would make this the real Task 7 closure the Opus seat aimed for.

— Fable seat (independent verification against vendored dist, not docs prose)

lifeodyssey (commented):
## Final review (HEAD `77a6fd50`): approve

**First, my error.** I approved a functional no-op. I verified the *contract* — the type declaration's doc comment, the docs' `141.101.64.0/18` example, the ordering of the `deniedHosts` check inside `ContainerProxy.fetch` — and never read the matcher those three things feed into. `matchesHostList` (`container.js:121-123`) is `patterns.some(p => simpleGlobMatch(p, hostname))`, and `simpleGlobMatch` (104-120) on a pattern with no `*` collapses to `pattern === value`. `"10.0.0.0/8" === "169.254.169.254"` is false; there is no `/`-parsing anywhere in that file. Five entries, zero matches, ever. "The type is `string[]` and the doc comment says CIDR" is not evidence that the matcher parses CIDR, and I treated it as such. Good catch by Fable.

### ① Glob expansion completeness — verified exhaustively, correct

I did not trust the ported matcher in the test file. I extracted `simpleGlobMatch`/`matchesHostList` from the vendored `container.js` source itself, bound them as live functions, and ran `DENIED_EGRESS_HOSTS` (86 entries) through a full 0-255 second-octet sweep for all five families:

```
octet sweep mismatches: NONE
```

Every `172.{16..31}.*` matches and no other `172.*` does; every `100.{64..127}.*` matches and no other `100.*` does; all 256 `10.*` match; exactly `192.168.*` and exactly `169.254.*` match. Both requested boundary pairs confirmed on the real matcher:

```
172.15.255.255 -> false     172.16.0.0    -> true
172.31.255.255 -> true      172.32.0.0    -> false
100.63.255.255 -> false     100.64.0.0    -> true
100.127.255.255 -> true     100.128.0.0   -> false
```

Range math is right: `/12` → second octet 16–31 (`length: 16`, `16 + i`), `/10` → 64–127 (`length: 64`, `64 + i`). No off-by-one at either end.

The trailing dot in each pattern is load-bearing and correct — `172.160.1.1` and `110.0.0.1` both correctly miss, because `"172.16.*"` requires the literal `"172.16."` prefix. Public probes (`1.1.1.1`, `8.8.8.8`) miss; all three spec AC error-path addresses plus `192.0.0.192` and `metadata.google.internal` hit.

**Test quality is now real.** Porting the actual algorithm instead of hand-rolling `ipInCidr()` is the correct fix — the previous test's helper validated a semantics the platform never had, which is precisely how the no-op hid behind a green check. I diffed the ported copy against the vendored source: character-identical. The canary + the empty-list mutation guard close the regression path.

### ② ContainerProxy export and the three wordings — all landed

- `export { ContainerProxy } from "@cloudflare/containers";` in `worker/entry.ts`, with the throw site cited. `worker/entry.test.ts` guards it by source-text match, and the comment honestly explains why a runtime-shape assertion isn't possible under `node --test`.
- Intercept-all promotion documented as limit 3 with the exact call site (`container.js:1121`), the sticky-until-restart caveat, and the correct scoping that `asyncpg` is unaffected *and correspondingly uncovered*.
- The `520` semantics are stated plainly — "Nothing is 'refused at the network layer' in the socket sense" — so a future engineer debugging a 520 lands on this section.
- `apps/agent/AGENTS.md` no longer says "not a container network policy"; it now scopes the convention to exactly what the platform layer does not reach.

### ③ Staging checklist — correctly ordered

`ContainerProxy` reachability is step 1 and labelled a precondition, ahead of the ACs; the intercept-all latency check on the MiMo hop is step 2. AC disposition still separates "verified today" from "needs live check" without fabricating closure.

---

### P2 — IPv6 literals are unmatched; worth a fourth limit

Verified against the real matcher: `[::1]`, `[fd00::1]`, and `[::ffff:a9fe:a9fe]` (the IPv4-mapped form of `169.254.169.254`) all return `false`. `url.hostname` yields the bracketed, colon-compressed form, which no dotted-quad glob can match. Suggest adding to the limits list:

> 4. **IPv6 literals are not covered.** `url.hostname` renders IPv6 in bracketed compressed form (`[::1]`, `[fd00::1]`, `[::ffff:a9fe:a9fe]` — the IPv4-mapped form of `169.254.169.254`), which no dotted-quad glob matches. ULA (`fd00::/8`), IPv6 link-local (`fe80::/10`), and IPv4-mapped IPv6 are therefore out of scope for this layer and remain the application-layer guard's job (`egress_guard` already handles IPv4-mapped-IPv6).

### P2 — the encoded-IPv4 bypass is already closed for free; say so, so nobody "fixes" it

`ContainerProxy.fetch` does `new URL(request.url)` before matching, and the WHATWG host parser normalizes numeric hosts. Verified:

```
http://2852039166/          -> hostname "169.254.169.254"
http://0xA9FEA9FE/          -> hostname "169.254.169.254"
http://0251.0376.0251.0376/ -> hostname "169.254.169.254"
http://METADATA.GOOGLE.INTERNAL/ -> hostname "metadata.google.internal"
```

So decimal/octal/hex-encoded IPv4 and uppercase hostnames are covered without extra entries. Worth one sentence in "What is implemented", because otherwise a future contributor either believes there's a bypass and adds junk entries, or believes there is one and doesn't check.

### P2 — prefix globs over-match subdomains (fail-closed)

`10.0.0.1.evil.com` matches (`"10.*"` is a prefix glob, not an IP anchor). Same for `169.254.x.attacker.com`, `172.16.foo.com`, etc. This errs toward blocking, so it is safe, but it does mean a legitimate hostname beginning `10.`/`172.16.`–`172.31.`/`192.168.`/`169.254.`/`100.64.`–`100.127.` would be blocked. Worth a half-sentence noting the over-match is deliberate and fail-closed.

### P2 — canary pins signatures, not the whole body

The canary asserts three lines (`function simpleGlobMatch(pattern, value)`, `const parts = pattern.split('*');`, `function matchesHostList(hostname, patterns)`). A semantic change to the body — the final `pos <= value.length - parts[...].length` comparison, or the package *adding* CIDR support — would slip through green. Consider pinning the full normalized function body, or asserting a hash of it, so a version bump that changes semantics fails loudly rather than silently re-diverging the port.

---

**Verdict: approve.** ① passes exhaustively against the real vendored matcher with correct boundaries on both requested ranges; ② and ③ are fully landed with accurate source citations. The four P2s are additive precision, none of them affect whether the shipped control works. Good correction — the doc now records both wrong turns instead of quietly overwriting them, which is what makes the section worth trusting.

lifeodyssey (commented):
## Final review (HEAD `77a6fd50`): approve

**First, my error.** I approved a functional no-op. I verified the *contract* — the type declaration's doc comment, the docs' `141.101.64.0/18` example, the ordering of the `deniedHosts` check inside `ContainerProxy.fetch` — and never read the matcher those three things feed into. `matchesHostList` (`container.js:121-123`) is `patterns.some(p => simpleGlobMatch(p, hostname))`, and `simpleGlobMatch` (104-120) on a pattern with no `*` collapses to `pattern === value`. `"10.0.0.0/8" === "169.254.169.254"` is false; there is no `/`-parsing anywhere in that file. Five entries, zero matches, ever. "The type is `string[]` and the doc comment says CIDR" is not evidence that the matcher parses CIDR, and I treated it as such. Good catch by Fable.

### ① Glob expansion completeness — verified exhaustively, correct

I did not trust the ported matcher in the test file. I extracted `simpleGlobMatch`/`matchesHostList` from the vendored `container.js` source itself, bound them as live functions, and ran `DENIED_EGRESS_HOSTS` (86 entries) through a full 0-255 second-octet sweep for all five families:

```
octet sweep mismatches: NONE
```

Every `172.{16..31}.*` matches and no other `172.*` does; every `100.{64..127}.*` matches and no other `100.*` does; all 256 `10.*` match; exactly `192.168.*` and exactly `169.254.*` match. Both requested boundary pairs confirmed on the real matcher:

```
172.15.255.255 -> false     172.16.0.0    -> true
172.31.255.255 -> true      172.32.0.0    -> false
100.63.255.255 -> false     100.64.0.0    -> true
100.127.255.255 -> true     100.128.0.0   -> false
```

Range math is right: `/12` → second octet 16–31 (`length: 16`, `16 + i`), `/10` → 64–127 (`length: 64`, `64 + i`). No off-by-one at either end.

The trailing dot in each pattern is load-bearing and correct — `172.160.1.1` and `110.0.0.1` both correctly miss, because `"172.16.*"` requires the literal `"172.16."` prefix. Public probes (`1.1.1.1`, `8.8.8.8`) miss; all three spec AC error-path addresses plus `192.0.0.192` and `metadata.google.internal` hit.

**Test quality is now real.** Porting the actual algorithm instead of hand-rolling `ipInCidr()` is the correct fix — the previous test's helper validated a semantics the platform never had, which is precisely how the no-op hid behind a green check. I diffed the ported copy against the vendored source: character-identical. The canary + the empty-list mutation guard close the regression path.

### ② ContainerProxy export and the three wordings — all landed

- `export { ContainerProxy } from "@cloudflare/containers";` in `worker/entry.ts`, with the throw site cited. `worker/entry.test.ts` guards it by source-text match, and the comment honestly explains why a runtime-shape assertion isn't possible under `node --test`.
- Intercept-all promotion documented as limit 3 with the exact call site (`container.js:1121`), the sticky-until-restart caveat, and the correct scoping that `asyncpg` is unaffected *and correspondingly uncovered*.
- The `520` semantics are stated plainly — "Nothing is 'refused at the network layer' in the socket sense" — so a future engineer debugging a 520 lands on this section.
- `apps/agent/AGENTS.md` no longer says "not a container network policy"; it now scopes the convention to exactly what the platform layer does not reach.

### ③ Staging checklist — correctly ordered

`ContainerProxy` reachability is step 1 and labelled a precondition, ahead of the ACs; the intercept-all latency check on the MiMo hop is step 2. AC disposition still separates "verified today" from "needs live check" without fabricating closure.

---

### P2 — IPv6 literals are unmatched; worth a fourth limit

Verified against the real matcher: `[::1]`, `[fd00::1]`, and `[::ffff:a9fe:a9fe]` (the IPv4-mapped form of `169.254.169.254`) all return `false`. `url.hostname` yields the bracketed, colon-compressed form, which no dotted-quad glob can match. Suggest adding to the limits list:

> 4. **IPv6 literals are not covered.** `url.hostname` renders IPv6 in bracketed compressed form (`[::1]`, `[fd00::1]`, `[::ffff:a9fe:a9fe]` — the IPv4-mapped form of `169.254.169.254`), which no dotted-quad glob matches. ULA (`fd00::/8`), IPv6 link-local (`fe80::/10`), and IPv4-mapped IPv6 are therefore out of scope for this layer and remain the application-layer guard's job (`egress_guard` already handles IPv4-mapped-IPv6).

### P2 — the encoded-IPv4 bypass is already closed for free; say so, so nobody "fixes" it

`ContainerProxy.fetch` does `new URL(request.url)` before matching, and the WHATWG host parser normalizes numeric hosts. Verified:

```
http://2852039166/          -> hostname "169.254.169.254"
http://0xA9FEA9FE/          -> hostname "169.254.169.254"
http://0251.0376.0251.0376/ -> hostname "169.254.169.254"
http://METADATA.GOOGLE.INTERNAL/ -> hostname "metadata.google.internal"
```

So decimal/octal/hex-encoded IPv4 and uppercase hostnames are covered without extra entries. Worth one sentence in "What is implemented", because otherwise a future contributor either believes there's a bypass and adds junk entries, or believes there is one and doesn't check.

### P2 — prefix globs over-match subdomains (fail-closed)

`10.0.0.1.evil.com` matches (`"10.*"` is a prefix glob, not an IP anchor). Same for `169.254.x.attacker.com`, `172.16.foo.com`, etc. This errs toward blocking, so it is safe, but it does mean a legitimate hostname beginning `10.`/`172.16.`–`172.31.`/`192.168.`/`169.254.`/`100.64.`–`100.127.` would be blocked. Worth a half-sentence noting the over-match is deliberate and fail-closed.

### P2 — canary pins signatures, not the whole body

The canary asserts three lines (`function simpleGlobMatch(pattern, value)`, `const parts = pattern.split('*');`, `function matchesHostList(hostname, patterns)`). A semantic change to the body — the final `pos <= value.length - parts[...].length` comparison, or the package *adding* CIDR support — would slip through green. Consider pinning the full normalized function body, or asserting a hash of it, so a version bump that changes semantics fails loudly rather than silently re-diverging the port.

---

**Verdict: approve.** ① passes exhaustively against the real vendored matcher with correct boundaries on both requested ranges; ② and ③ are fully landed with accurate source citations. The four P2s are additive precision, none of them affect whether the shipped control works. Good correction — the doc now records both wrong turns instead of quietly overwriting them, which is what makes the section worth trusting.

Issue #464, opened 2026-07-28
edge: normalizeV1Path does not percent-decode — /v1/%63hat skips the auth rate limiter while the container decodes it to /v1/chat


lifeodyssey:
From PR #451 round-3 Fable-seat review (P3, pre-existing class, not a Task 9 AC).

`URL.pathname` in the Worker does not percent-decode, so `POST /v1/%63hat` misses `AUTH_RATE_LIMITED_V1` matching at the edge — but ASGI hands Starlette the decoded path, so the container routes it to the chat handler. One uncounted request per encoding variant.

Reachability depends on whether Cloudflare's edge normalizes percent-encoding before the Worker sees the path — unverified either way.

Fix options:
1. `decodeURIComponent` inside `normalizeV1Path` behind try/catch + assertion in `authRateLimitScope.test.ts`, or
2. verify CF edge normalization empirically on staging and record the finding in the code comment.

Related: #451 (auth rate limiting), #284 Task 9.

lifeodyssey:
时效核验(2026-08-04):**已修,关闭。**

`workers/edge/routing-policy.ts:35-49`(C2 由 app.ts 拆出)已实现 `decodedForRouting()`:路由判定前 `decodeURIComponent`,且**malformed `%` escape 返回 null → 调用方 fail-closed**(视为受限路径)——比本卡当初要求的更严:evasion 尝试产生的不可解析路径不会落进「不受限」分支。注释块直接引用 #464 与 #479 P1-1,附带完整攻击叙述(`/v1/%62yok/probe` 绕过 checkRateLimit 却命中容器 handler)。

线程判定: 无未处理 findings,凭上述代码证据关闭。

PR #479, opened 2026-07-29
feat(#284 T4+T5): BYOK quota exemption regression locks + vision/probe endpoint

lifeodyssey:
## Summary

Implements #284 Task 4 (quota boundary regression locks) + Task 5 (`POST /v1/byok/probe`), building on the merged T1/T2/T3/T6/T9 work.

**Task 4** — the login-gated dollar-budget/anon-quota exemption is already correct by construction (T3, #472, #477): a logged-in caller never reaches either anonymous guard on either tier. This PR adds the regression locks that pin it so a future breaker refactor can't silently start charging BYOK users, plus the two review follow-ups from #477 (P3-c, P3-d):

- Container: an authenticated BYOK turn never reads `daily_usage`, even with the budget/quota actually configured on (not the default-disabled values, which would make the test pass vacuously).
- Edge: an authenticated `/v1/chat` request is unaffected by a latched anonymous budget (`worker/byokBudgetExemption.test.ts`) — latches the **real current UTC day** (`utcDayKey(Date.now())`) and asserts directly on the guard's own call trace, not just the response status (see Review round 2 below for why the first version of this test was insufficient).
- The login gate's teeth: an anonymous `X-BYOK-*` request is rejected with `403 byok_requires_login` *before* either anonymous guard's repo read runs — the header spends nothing. (Note: this deliberately diverges from the design spec's literal AC4 wording — see the deviation note in `test_byok_guards_still_apply.py`'s module docstring.)
- A completed BYOK turn banks under `UsageScope "byok"` at `cost_usd == 0.0` with non-zero token counts, and today's `anon` spend is unaffected.
- The injection preflight and the singleton (validated) `animichi_agent` still gate/dispatch a BYOK turn — T8 guard retention.
- P3-c: a direct unit test on `_require_nonblank_key`.
- P3-d: a web_tools-level test proving `translate_anime_title` actually calls the injected `title_translator` (the D18 boundary's real callsite), not just that `RuntimeAPI` passes one.

**Task 5** — new `agent/interfaces/routes/byok.py`: `POST /v1/byok/probe`, authenticated-only (not in the worker's `ANON_V1`, so an unauthenticated call 401s at the edge before the container — `worker/byokProbeAuth.test.ts`). Validates a BYOK credential and detects vision support with one 1×1 PNG probe call:
- Failure taxonomy: only 401/403 are `byok_credential_rejected`; only 400/422 mean "reachable, model rejects the image part"; every other status (404/429/5xx) and every non-HTTP exception collapse to `provider_unreachable` via a bare `except Exception` — no path escapes to the generic 500 handler.
- Fixed ≤5s timeout + ≤64 KiB response cap, installed at client-construction time (`build_byok_model(..., transport_wrapper=...)`), contain the probe's use as a reachability oracle for a caller-chosen endpoint. A residual timing side-channel below the 5s ceiling is accepted and tracked in #481, not silently declared solved.
- `openai-compatible`'s caller-chosen `base_url` is pre-validated against the SSRF guard (own `egress_blocked` code) before any model is built.
- Rate-limited (already-merged Task 9 `AUTH_RATE_LIMITED_PREFIX = "/v1/byok/"`, now decode-aware — see Review round 2) and edge-login-gated (already-merged, `worker/byok.test.ts`).

## Review round 2 (Opus + Fable, both `request_changes` on the first version)

Both reviewers ran mutation testing against the first version of this PR and found real gaps, all fixed in a follow-up commit:

- **P1 (Opus): rate-limiter bypass via percent-encoding.** `isAuthRateLimited` matched on the raw `URL.pathname`, which never decodes `%XX` — but the container's ASGI router does. `/v1/%62yok/probe` read as "not `/v1/byok/`" (zero limiter calls) while still executing the real probe in the container: an authenticated caller could burst unlimited outbound calls one percent-encoded letter at a time. Fixed with a decode-before-match step that fails **closed** (rate-limited) on a malformed escape; added a real `app.request` regression (not just a pure-function test) in `worker/authRateLimitScope.test.ts`, closing the related #464 follow-up.
- **P1 (Opus): error-code collapse missed a 500 path.** `_run_probe`'s exception tuple didn't include `UnexpectedModelBehavior`/`AgentRunError`, so some provider responses could 500 through the generic handler — a fourth, distinguishable outcome. Replaced with a bare `except Exception`, and narrowed "vision:false, reachable:true" from "any non-401/403" to exactly 400/422 so 404/429/5xx correctly read as `provider_unreachable` instead of masquerading as a legitimate no-vision answer.
- **P1 (both, mutation-verified): containment constraints had zero real coverage.** Removing the cap wrapper, widening the timeout, or nulling `error_code` all left the original 11-test suite green. Added oversized-response (header-lie + streaming, each with a control proving the fixture is a genuine, otherwise-successful body), connection-failure, and mock-clock timeout tests, plus direct constant-value pins — every one of the three named mutations is now caught (re-verified after the fix).
- **P1 (Fable): T5-AC5 had no test despite being claimed covered**, and **the edge budget-exemption lock latched a fixed test day** while a hypothetical `Date.now()`-driven regression would read the real clock and pass anyway. Both fixed: `worker/byokProbeAuth.test.ts` added; `byokBudgetExemption.test.ts` now latches `utcDayKey(Date.now())` and asserts on the guard's call trace directly.
- **P2:** the probe's response cap now installs via a new `transport_wrapper` parameter on `build_guarded_async_client`/`build_byok_model` at construction time, never as a post-hoc `client._transport =` reassignment; functions split per the 1-10-50 guideline; opened #481 for the accepted timing-side-channel residual instead of leaving it undocumented.
- **P3:** added anthropic/gemini probe coverage (previously openai-compatible only); fixed the usage-metering test double's `total_cost_usd`, which ignored its `scope` argument and always returned `0.0` (decorative — would pass even on a real scope-bleed bug) and its accompanying date-bucket mismatch; corrected a test docstring's false citation of #472.

## Test plan

- [x] `make lint` — ruff check + format, clean
- [x] `make typecheck` — mypy, clean (137 files)
- [x] `make test` — 1651 passed, coverage 88.22% (floor 87%)
- [x] `pnpm run test:worker` — 194 passed
- [ ] `make test-integration` — blocked in this sandbox by a pre-existing, unrelated environment gap: the installed Atlas CLI (`v1.2.4`) doesn't match the pinned `0.30.0` the DB-container integration tests require. All 131 non-DB integration tests pass (22 skipped, unaffected by this change); the DB-container tests error identically on `main` before this PR's changes.
- [x] BYOK-scoped suite in isolation: `pytest -k byok` (unit + integration) → 120 passed
- [x] Mutation testing: every specific mutation named by either reviewer (percent-encoding bypass, error-code collapse, all three containment constraints, edge budget-exemption clock mismatch, usage-metering scope-bleed) manually re-verified caught after the fix

Not merging — for review per the harness workflow.

### DIFF
--- a/apps/agent/agent/agents/byok_models.py
+++ b/apps/agent/agent/agents/byok_models.py
@@ -35,7 +35,10 @@ from agent.config.byok_defaults import (
 )
 from agent.infrastructure.egress_errors import EgressBlocked
 from agent.infrastructure.egress_guard import validate_base_url
-from agent.infrastructure.egress_transport import build_guarded_async_client
+from agent.infrastructure.egress_transport import (
+    TransportWrapper,
+    build_guarded_async_client,
+)
 
 ByokProvider = Literal["openai-compatible", "anthropic", "gemini"]
 BYOK_PROVIDERS: Final[frozenset[str]] = frozenset(
@@ -194,9 +197,11 @@ async def _validate_openai_base_url(base_url: str | None) -> None:
         ) from exc
 
 
-async def _build_openai_compatible(credential: ByokCredential) -> ByokModel:
+async def _build_openai_compatible(
+    credential: ByokCredential, *, transport_wrapper: TransportWrapper | None
+) -> ByokModel:
     await _validate_openai_base_url(credential.base_url)
-    client = build_guarded_async_client()
+    client = build_guarded_async_client(transport_wrapper=transport_wrapper)
     sdk_client = AsyncOpenAI(
         base_url=credential.base_url,
         api_key=credential.key,
@@ -208,8 +213,10 @@ async def _build_openai_compatible(credential: ByokCredential) -> ByokModel:
     return ByokModel(model=model, client=client)
 
 
-def _build_anthropic(credential: ByokCredential) -> ByokModel:
-    client = build_guarded_async_client()
+def _build_anthropic(
+    credential: ByokCredential, *, transport_wrapper: TransportWrapper | None
+) -> ByokModel:
+    client = build_guarded_async_client(transport_wrapper=transport_wrapper)
     sdk_client = AsyncAnthropic(
         api_key=credential.key, http_client=client, max_retries=0
     )
@@ -218,7 +225,9 @@ def _build_anthropic(credential: ByokCredential) -> ByokModel:
     return ByokModel(model=model, client=client)
 
 
-def _build_gemini(credential: ByokCredential) -> ByokModel:
+def _build_gemini(
+    credential: ByokCredential, *, transport_wrapper: TransportWrapper | None
+) -> ByokModel:
     # `GoogleProvider.__init__` falls back to `os.getenv("GOOGLE_API_KEY")`
     # only when the `api_key` it receives is falsy (P1-2) — `_require_nonblank_key`
     # below is the structural guarantee that never happens here, so this call
@@ -226,7 +235,7 @@ def _build_gemini(credential: ByokCredential) -> ByokModel:
     # Retry behaviour is left at the `google-genai` SDK default (unlike the
     # openai-compatible family's explicit `max_retries=0`, T3-AC2 names only
     # `AsyncOpenAI`) — pinned here as a deliberate choice, not an oversight.
-    client = build_guarded_async_client()
+    client = build_guarded_async_client(transport_wrapper=transport_wrapper)
     provider = GoogleProvider(api_key=credential.key, http_client=client)
     model: Model = GoogleModel(credential.model, provider=provider)
     return ByokModel(model=model, client=client)
@@ -242,16 +251,28 @@ def _require_nonblank_key(credential: ByokCredential) -> None:
         raise ByokError("invalid_request", "BYOK credential key must not be blank.")
 
 
-async def build_byok_model(credential: ByokCredential) -> ByokModel:
+async def build_byok_model(
+    credential: ByokCredential,
+    *,
+    transport_wrapper: TransportWrapper | None = None,
+) -> ByokModel:
     """Build the per-request guarded model. Caller MUST `await client.aclose()`.
 
     Every family routes through `build_guarded_async_client` (Task 1's SSRF
     guard, Task 2's httpx-instrumentation exclusion) — there is no second,
     unguarded way to build a BYOK-bound `httpx.AsyncClient` in this module.
+
+    `transport_wrapper` (review follow-up, #479 P2): threaded straight to
+    `build_guarded_async_client` so a caller — the probe route (Task 5) —
+    can install an additional transport (its response-size cap) at
+    construction time, never via a post-construction `client._transport =`
+    reassignment.
     """
     _require_nonblank_key(credential)
     if credential.provider == "openai-compatible":
-        return await _build_openai_compatible(credential)
+        return await _build_openai_compatible(
+            credential, transport_wrapper=transport_wrapper
+        )
     if credential.provider == "anthropic":
-        return _build_anthropic(credential)
-    return _build_gemini(credential)
+        return _build_anthropic(credential, transport_wrapper=transport_wrapper)
+    return _build_gemini(credential, transport_wrapper=transport_wrapper)
--- a/apps/agent/agent/infrastructure/egress_transport.py
+++ b/apps/agent/agent/infrastructure/egress_transport.py
@@ -12,6 +12,7 @@ to a validated IP literal.
 from __future__ import annotations
 
 import ssl
+from collections.abc import Callable
 
 import httpx
 
@@ -134,11 +135,15 @@ def _exclude_from_httpx_instrumentation(transport: httpx.AsyncBaseTransport) -> 
         unwrap(transport, "handle_request")
 
 
+TransportWrapper = Callable[[httpx.AsyncBaseTransport], httpx.AsyncBaseTransport]
+
+
 def build_guarded_async_client(
     *,
     resolver: Resolver = default_resolve,
     timeout: httpx.Timeout | float | None = None,
     verify: ssl.SSLContext | str | bool = True,
+    transport_wrapper: TransportWrapper | None = None,
 ) -> httpx.AsyncClient:
     """Build the per-request BYOK transport.
 
@@ -151,9 +156,20 @@ def build_guarded_async_client(
     (#284 Task 1) and the httpx-instrumentation exclusion (#284 Task 2,
     X3/P1-1, applied in `GuardedAsyncTransport.__init__`); a hand-rolled
     `httpx.AsyncClient` for BYOK traffic would have neither.
+
+    `transport_wrapper` (review follow-up, #479 P2): applied at construction
+    time, never as a post-hoc `client._transport = ...` reassignment — the
+    probe route (#284 Task 5) uses this to install its response-size cap
+    *before* the client is ever handed to a caller, so there is no window
+    where an unwrapped, cap-free transport exists.
     """
+    transport: httpx.AsyncBaseTransport = GuardedAsyncTransport(
+        resolver=resolver, verify=verify
+    )
+    if transport_wrapper is not None:
+        transport = transport_wrapper(transport)
     return httpx.AsyncClient(
-        transport=GuardedAsyncTransport(resolver=resolver, verify=verify),
+        transport=transport,
         trust_env=False,
         follow_redirects=False,
         timeout=timeout if timeout is not None else httpx.Timeout(30.0),
--- a/apps/agent/agent/interfaces/fastapi_service.py
+++ b/apps/agent/agent/interfaces/fastapi_service.py
@@ -31,6 +31,7 @@ from agent.interfaces.routes._middleware import (
     register_observability_middleware,
 )
 from agent.interfaces.routes.bangumi import router as bangumi_router
+from agent.interfaces.routes.byok import router as byok_router
 from agent.interfaces.routes.chat import router as chat_router
 from agent.interfaces.routes.conversations import router as conversations_router
 from agent.interfaces.routes.feedback import router as feedback_router
@@ -183,6 +184,7 @@ def create_fastapi_app(
     app.include_router(health_router)
     app.include_router(runtime_router)
     app.include_router(chat_router)
+    app.include_router(byok_router)
     app.include_router(feedback_router)
     app.include_router(conversations_router)
     app.include_router(bangumi_router)
--- a/apps/agent/agent/interfaces/routes/byok.py
+++ b/apps/agent/agent/interfaces/routes/byok.py
@@ -1,0 +1,211 @@
+"""`POST /v1/byok/probe` — BYOK credential validation + vision-capability
+probe (#284 Task 5, spec D5).
+
+Deliberately dual-purpose (OQ-2 ruling): a single minimal call both validates
+the caller's credential (one upstream request) and detects whether the
+configured model accepts an image part, so configuring a key costs the user
+exactly one probe request instead of two.
+
+Authenticated route only (login-gated like the rest of BYOK): not listed in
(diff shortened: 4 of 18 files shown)


lifeodyssey (commented):
## Fable seat — independent review of PR #479 (#284 T4+T5)

Independently ran gates in a detached worktree (25 new Python tests + 187 worker tests green), re-ran the implementer's mutations plus five of my own, and probed the T5 containment constraints with throwaway tests. Spec baseline: `docs/spec-284-byok-v2` @ efc0abaa.

**Verdict: request_changes** — 3 P1s, all cheap to fix.

### What held up under mutation (verified, not taken on trust)

- **T4 container locks are mutation-solid.** All red as claimed: login-gate reordered after both guards → `test_an_anonymous_byok_header_never_touches_the_anon_budget_or_quota` fails; quota-only moved first → still fails (the `increment_and_count` canary matches the real repo method); breaker's non-anonymous early-return deleted → `test_an_authenticated_byok_turn_never_reads_daily_usage` fails; BYOK priced normally in `_record_usage` → metering test fails; `403` dropped from the probe's auth tuple → `test_provider_403_also_reports_credential_rejected` fails. The non-default `Settings(anon_daily_cost_budget_usd=5.0, anon_daily_message_quota=10)` genuinely arms both guards (defaults confirmed disabled: `0.0`/`None`) — the self-reported first-draft vacuity is fixed on the container side.
- **T5 taxonomy behaves as documented**: my own probes confirmed timeout (patched to 50ms), >64KiB body, and `httpx.ConnectError` all fold to `provider_unreachable`; mid-run `EgressBlocked` folds there too (no post-connect oracle); pre-flight SSRF → `400 egress_blocked` with `build_byok_model` never called. 401/403 → `byok_credential_rejected`, no key echo; every `ByokError` message is a fixed string (no base_url/key interpolation).
- **T5-AC6** satisfied by already-merged T9 (`AUTH_RATE_LIMITED_PREFIX = "/v1/byok/"`, `byok.test.ts` AC2 + trailing-slash pins). Vision result shape maps cleanly onto `byokStorage`'s three-state (`null` until probed / `setByokVisionSupported(bool)`); OQ-2 validate+probe-in-one honored with the exactly-one-upstream-call assertion. `forwardV1`'s inbound strip (X-User-*/X-Anon-Id/x-byok-endpoint) is shared code covering the probe path.

### P1

1. **T5-AC5 is not covered, and the PR body's claim for it is wrong.** The AC requires an unauthenticated `POST /v1/byok/probe` → 401 "asserted at the worker level, pinning that the probe is absent from the anonymous allowlist". No such test exists: every `worker/byok.test.ts` case hits the probe *authenticated*; the suite's unauthenticated-401 assertions use `/v1/chat` and `/v1/feedback` only. The PR body says this is "already-merged, `worker/byok.test.ts`" — it isn't. One test (unauthenticated `authenticate: ok:false`, path `/v1/byok/probe`, expect 401 + zero container fetches) closes it.
2. **The new edge lock (T4-AC2) is vacuous against a realistic mutant — the same vacuity class the implementer fixed on the container side, reproduced at the edge.** `byokBudgetExemption.test.ts` latches `DAY_KEY = utcDayKey(Date.UTC(2026, 6, 28))`, but `authenticatedForward` receives no clock, so any real refactor moving the breaker onto the shared path keys by `utcDayKey(Date.now())` — a latch armed for 2026-06-28 that nothing ever reads. Mutation-verified: inserting `if (await budgetLatched(env.EDGE_GUARD, utcDayKey(Date.now()))) return budgetGuidanceResponse();` at the top of `authenticatedForward` leaves **both new tests green**; it is caught only by the pre-existing `anonymous.test.ts` "the breaker does not touch logged-in callers" (which latches via a real anonymous request at the real clock). Notably the PR's mutation-evidence list contains no edge mutation. Fix: latch `utcDayKey(Date.now())` (as the pre-existing test effectively does), or follow the spec AC's letter and assert on a guard stub that records `budgetLatched` calls.
3. **Containment constraints (b) ≤5s timeout and (c) ≤64KiB cap have zero committed test coverage.** Per coverage.xml from the PR's own probe suite, `routes/byok.py` misses lines 104-105 and 110-111 (both `_ProbeResponseTooLarge` raise paths — the cap is never enforced in any test), 154-162 (the entire `TimeoutError/TransportError/EgressBlocked/…` → `provider_unreachable` fold), and 202-203 (`ByokError` branch). The two security constraints named in the route's own docstring can be deleted and the suite stays green; this also puts patch coverage on the file at roughly 89% (<95%). My throwaway probes show the behavior is correct today — the tests just need writing (transport that sleeps past a patched timeout; transport returning >64KiB; transport raising `ConnectError`).

### P2

4. **Hostile 200-with-garbage-body → 500, not `provider_unreachable`** (verified: a 200 response with a non-JSON body raises uncaught through `_run_probe` and returns the generic `internal_error` 500). No detail leaks, but a caller-chosen endpoint can mint container 500s/alert noise at will, and the 500-vs-200 split itself distinguishes "answered garbage" from "unreachable" — a distinction constraint (a) meant to close. Fold `UnexpectedModelBehavior`/response-parse errors into `provider_unreachable`.
5. **`byok_model.client._transport = _CappedResponseTransport(...)` mutates httpx's private attribute in production code.** Fragile across httpx upgrades and bypasses `_mounts` routing. Prefer installing the capped transport where the client is constructed (a `build_byok_model` knob or a probe-specific builder in `byok_models.py`).
6. `test_byok_probe.py` is 219 lines — over the repo's ≤200-line test-file guardrail (AGENTS.md). Splitting the taxonomy cases out would also give P1-3's new tests a home.
7. `test_a_byok_turn_never_moves_todays_anon_spend_total`: the verdict half is vacuous — `_UsageRepoDouble.total_cost_usd` unconditionally returns `0.0` regardless of scope, so `verdict.spent_usd == 0.0` holds no matter what was banked; only the final `repo.calls[0][1] == "byok"` line has teeth (duplicating test 1). Make the double actually sum per-(date, scope), or drop the verdict assertions.

### P3

8. **T4-AC4 spec deviation is unrecorded and the citation looks wrong.** The frozen spec's AC4 says an anon+`X-BYOK-*` request "is still counted and gated by the anonymous budget exactly as a normal anonymous request would be"; the shipped ordering (locked by this PR) rejects it *before* either guard runs — nothing is counted. AC3 and AC4 were mutually inconsistent as written and the stricter reading is defensible, but the test docstring attributes the ruling to "#472 combination, reviewed" — #472 is the anon-quota PR and contains no such ruling. Record the deviation (or the correct review source) where the AC mapping lives.
9. T4-AC6's `output_validator` half is asserted only indirectly (singleton dispatch carries it); T4-AC8 has no dedicated test (relies on the pre-existing #274 suite staying green). Both acceptable — the AC mapping should say so explicitly.
10. `_validate_openai_compatible_egress`'s `credential_base_url is None` early-return is dead on every reachable path (parse layer already requires base_url for openai-compatible; other providers never call it) — uncovered line 134 confirms. Drop the branch or type the parameter `str`.
11. `handle_byok_probe` (~27 lines) and `_run_probe` (~26) exceed the 10-line rule; consistent with `handle_chat`'s existing shape, so noting for the ledger only.

### Quality ratchet

T4: 8 ACs — AC1 ok · AC2 test present but vacuous (P1-2) · AC3 ok · AC4 deviation (P3-8) · AC5 ok · AC6 ok (indirect) · AC7 ok (merged T9) · AC8 inherited. T5: 6 ACs — AC1-AC4 ok · AC5 missing (P1-1) · AC6 ok (merged T9). Eval suite: skipped — no agent behavior/prompt/tool changes.

lifeodyssey (commented):
## Fable seat — final verification round (PR #479, rebased @ 8e438d20)

Re-ran everything independently in a fresh detached worktree: 45 new/changed Python tests + 204 worker tests green locally; `make typecheck` clean (137 files); ruff clean. Re-ran all prior mutations plus three new ones against the revised code.

### All three P1s verified closed — with mutation evidence

1. **T5-AC5** — `worker/byokProbeAuth.test.ts` now pins both the plain unauthenticated 401 (zero container fetches) and the stronger "401 even with `ANON_ACCESS_ENABLED` — absent from `ANON_V1`" case. Closed.
2. **T4-AC2 edge lock** — re-ran my exact killer mutation (`budgetLatched(env.EDGE_GUARD, utcDayKey(Date.now()))` inserted into `authenticatedForward`): **both revised tests now fail** (previously both passed). The real-clock latch plus the `/budget`-shard call-trace assertion make it a mechanism-level pin. Closed. The bonus percent-encoding fix (`/v1/%62yok/probe` limiter bypass, fail-closed on malformed escapes, end-to-end 429 regression) is a genuine catch beyond my review — verified green, and the fail-closed direction is right.
3. **Containment coverage** — new mutations all red: dropping `transport_wrapper=_CappedResponseTransport` from `_resolve_probe_model` → `test_the_cap_transport_is_installed_at_construction_never_reassigned` fails; `_PROBE_TIMEOUT_SECONDS = 500.0` → the value pin fails; widening the taxonomy back to "any non-401/403 = reachable" → the parameterized 404/429/5xx collapse test fails. The control-test pattern (every capped rejection paired with an uncapped success control on the same fixture) is exactly right. Prior M1/M5 (guard ordering, BYOK pricing) re-verified red on the rebased tree.

P2s closed as described: construction-time `transport_wrapper` (no `client._transport` reassignment in production; one test pins the kwarg), file split, metering double now sums per `(usage_date, scope)`. P1-2 from my last round (hostile-body 500) closed via the bare `except Exception` fold + the 404/429/5xx narrowing, with the residual timing-oracle honestly documented and tracked (#481). T4-AC4 deviation note now recorded without the false #472 citation.

### The one blocker: CI is red, and the red test is a new flaky-pattern control

`Python integration (Neon)` failed on `test_byok_probe_containment.py::test_without_the_cap_the_oversized_body_would_have_succeeded` — got `provider_unreachable`, expected success. The log shows a `CancelledError` propagating through the SDK's exit stack into `_run_probe`'s `except Exception`. Locally this test passes 3/3 at **0.15s**; the patched 30s ceiling being genuinely exceeded by a 200x slowdown is implausible, so this looks like a cancellation/teardown interaction, not simple runner load — i.e., the 30s widening in the docstring treats the wrong cause. Two asks before merge:

- Root-cause the cancellation (why does a preloaded 66KiB fake-transport probe get cancelled on that lane?) rather than re-running until green — a rerun-green would just merge a known-flaky test into the required lane.
- Same pass: `test_the_probe_never_exceeds_its_timeout_ceiling` asserts real wall-clock `elapsed < 1.0` — same pattern class (repo rule: mock the clock, no timing-dependent asserts). It passed this run, but it's the next one to flake on a loaded runner; consider widening the bound generously (e.g. `< transport sleep of 2.0s`) since the value pin test already guards the constant.

### Residual notes (non-blocking, P3)

- `_byok_probe_shared.py:118` carries `# type: ignore[arg-type]` — the repo's no-suppression rule; trivially avoidable by typing the `provider` parameter as `ByokProvider`. (Inert w.r.t. the gate — `make typecheck` doesn't scan `tests/integration` — but the rule is about the code, not the gate.)
- Line caps: `routes/byok.py` 320 (>300), `test_byok_probe_containment.py` 230 and `test_byok_guards_still_apply.py` 210 (>200). All overage is docstring/deviation-note mass this review itself demanded — accepted residual, note for the ratchet.
- The cap wrapper's *success* path (small body through `_CappedResponseTransport` → `_rebuild_response`) is never exercised by any test — coverage lines 127/151-152/173 miss in every suite combination, because capped tests all trip the cap and controls run uncapped. I verified directly that the path is sound (body passes through intact), so no latent bug ships, but one test (`apply_probe_cap=True` + small success body → `vision: true`) would close the only remaining hole where a wrapper bug could brick every successful prod probe under a green suite.

### Verdict

**request_changes — solely on the red Neon lane / flaky control.** Every finding from my previous round is verifiably fixed and mutation-hardened; the code itself is in the best shape it's been. Once the flaky control is root-caused and the lane is genuinely green (not rerun-green), this seat is an approve.

PR #480, opened 2026-07-29
feat(#284 T6-UI + T8): BYOK settings panel + discovery/guidance journey

lifeodyssey:
Implements **#284 Task 6 (UI half)** and **Task 8 (full journey)** per spec rev4 (`docs/superpowers/specs/2026-07-28-284-byok-design.md` @ efc0abaa).

## Task 6 — settings panel (UI half; storage half landed in #467)
- `ByokSettings` in the chat input area (gear toggle in `ChatInput`, panel docks above the composer): three-family selector, password-masked key field, model field (**required** for openai-compatible; pre-filled from `BYOK_DEFAULT_MODEL` and user-overridable for Anthropic/Gemini — OQ-1), `base_url` for openai-compatible only.
- Save is fused with `/v1/byok/probe` (OQ-2: one request validates + detects vision). **T5 is in flight in parallel** — the probe client (`byok-probe.ts`) is written against the spec-pinned response shape `{"vision","reachable","error_code"}` and exercised via MSW; live-wire validation is a Tester browser AC.
- Inline error region inside the panel maps `validateByokConfig` codes + probe taxonomy onto #467's `byok-i18n` copy; probe 401/403 renders the explicit key-not-accepted state; raw key is never rendered back (masked summary only, field cleared on save).
- Vision flag stored three-state (`null` unprobed / true / false); the badge itself renders only on a **definitive** true (#479 P2-1: a probe with any `error_code` leaves the flag unprobed). Cleared in lockstep with the credential.

## Task 8 — discovery + guidance journey (9 AC)
- **Touchpoint A (D11)**: `BudgetExhausted` gains a secondary "use your own key" affordance → opens `ByokUpsell` (value explainer: benefit / browser-only key / account required) → login.
- **Touchpoint B**: anonymous visitors get the panel's teaser state — value prop + sign-in CTA, **no key input rendered at all**.
- **Touchpoint C**: `byok_requires_login` → new classifier state **D13** → journey entry (explainer), never the D8 session-expired story; `byok_credential_rejected` → **D14** key-not-accepted strip (no retry).
- **Setup intent rides the callback URL** (T8-AC4 fresh-tab): `useMagicLinkForm` keeps `callbackURL = ${origin}/auth/callback` and appends a **validated** relative `next` (spec P2-4); `/auth/callback` honours it and `/chat?settings=byok` auto-opens the panel via `useByokPanel`.
- **T14 open-redirect guard** (`lib/auth/returnTarget.ts`): same-origin relative paths only — spec's full vector set tested (`https://evil.test/`, `//evil.test`, `http:/evil.test`, `/\evil.test`, backslash variants, control chars) with `/` fallback, applied at search parse, at navigate, and before the mailed link.
- Coexists with #462's deferred-save intent: save intent stays in storage with its own key; BYOK intent rides the URL — independent carriers, no overwrite.

## Design alignment
- Panel/journey follow the in-repo Animal Island conventions (`.chat-*` banner family geometry, cream press-shadow CTAs per DS 補全 `.btn.cream`, warning-tint left-border strips per Chat 状态总览 banner family). **No gold spent** (一屏唯一金, #462 precedent) — enforced by test.
- Tokens only: `byok-design-tokens.test.ts` is the T6-AC9 source grep (Tailwind palette classes, inline hex, non-`var(--)` colors in every `.chat-byok*` rule).
- ja/zh/en copy for every new string; enumeration + raw-code-leak tests extended.

## Verification
- Gates (apps/web): `typecheck` ✅ · `lint:oxlint --deny-warnings` ✅ · `pnpm test` **1437 passed**, coverage **98.61/95.31/98.78/99.44** vs floors 98/95/98/99 (ratchet untouched).
- **Mutation gate — 12 mutations, all red**: drop `//` rejection, drop backslash rejection, skip `next` validation in callbackURL, callback route ignores `next`, drop D13/D14 classifier mapping, no OQ-1 model prefill, vision flag not persisted, deep-link never opens panel, save ignores validation, teaser renders key field, D11 loses secondary affordance, token grep misses `bg-white`.
- Browser ACs (T6-AC6/AC7, T8-AC1, T8-AC4 live run) need the running stack with `ANON_ACCESS_ENABLED=true` (spec P3 precondition; anonymous surface tracked by #447) + a real magic-link mailbox — flagged for the Tester red-team/journey walkthrough (Verification Plan §6-7). E2E suite not run here for the same reason.

## Dependencies
- Consumes #467 (`byokStorage`, `byok-i18n`) and #462 (`useAuthCallback` deferred-save shape) — both merged.
- `/v1/byok/probe` (T5) and `byok_requires_login`/`byok_credential_rejected` wire codes (T3/T4) are in flight; literals pinned from the spec, MSW-covered.

Refs #284 (Task 6 UI, Task 8).

---
**Review round 2 (Opus request_changes) — addressed @ 740c34e9:**
- **P1-1**: anonymous teaser offers masked-summary + clear for a stored credential (lapsed session) — deletion never gated on login.
- **P1-2**: with a BYOK return intent, /auth/callback navigates to `next` even on a failed deferred-save replay (intent survives in storage for a later login); no-intent behaviour unchanged. Dual-intent coexistence tested at component + route level.
- **P2**: D14 gains an open-key-settings action (ByokPanel.show); probe runs generation-guarded (stale outcome after clear is dropped); token scan surface widened (round 1).
- **P3**: `..` segments rejected; raw-key absence on document.body for all 9 outcomes; Object.values enumeration; dead i18n key removed; cream hover/active press states; aria-controls + focus-into-panel; zh wording.
- Mutations for every new guard verified red (5/5). Gates: typecheck 0 · oxlint 0 · 1466/1466 · coverage 98.56/95.36/98.65/99.45.


lifeodyssey (commented):
## Fable independent review — PR #480 (#284 T6-UI + T8), second seat

Base: `feat/frontend-rebuild`. Spec: `origin/docs/spec-284-byok-v2` rev4, Task 6 UI-half + Task 8 all 9 AC compared line-by-line. Independent gate re-run in a detached worktree at `37150858`: **typecheck OK · lint:oxlint OK · 1437/1437 passed (198 files) · coverage 98.61/95.31/98.78/99.44 vs floors 98/95/98/99, ratchet untouched** (vitest.config.ts not in diff). 4 of the 12 self-reported mutations independently replicated, all red (drop `//` rejection; drop D13 classifier mapping; deep-link never opens panel; OQ-1 model lost on family switch). Eval suite: skipped — no agent behavior/prompt/tool changes.

### Journey triad — verified
- **A (D11)**: secondary affordance opens `ByokUpsell` (explainer), never a login form; explainer's primary action opens `LoginModal` with `returnTarget=BYOK_SETUP_TARGET`. OK
- **B (settings teaser)**: anonymous branch renders no key input, no password field, no save — asserted directly. OK
- **C (`byok_requires_login`)**: classifier D13 -> journey entry, D14 -> key-not-accepted strip; bare 403 stays D8 — all three asserted. OK
- **Fresh-tab semantics**: intent rides the callback URL only — `callbackURL = ${origin}/auth/callback?next=...` asserted from teaser AND upsell; `/auth/callback` honours `next` through the T14 guard (validated twice: parse + navigate); `?settings=byok` auto-opens via `useByokPanel`. No `sessionStorage` anywhere on the journey path. AC5 empty/whitespace-next fallback covered. T14 vector set complete vs spec (`https://`, `//`, `http:/`, `/\`, backslash variants, control chars, scheme-ish, relative-no-slash).
- **OQ-1**: openai-compatible model required blocks save *and* probe; anthropic/gemini prefill from `BYOK_DEFAULT_MODEL`, editable; hand-typed value survives switch (mutation-verified).

### Cross-PR alignment
- **#479 (T5)**: probe client `ProbeBody {vision, reachable, error_code}` matches #479's `ProbeResponse` (`ProbeErrorCode = byok_credential_rejected | provider_unreachable`) exactly; 400 `invalid_request`/`egress_blocked` + 401/403 taxonomy consistent. OK
- **#478**: zero file overlap, no conflict. OK
- **#467**: all four `ByokSaveError` codes (`key_required`/`key_invalid`/`model_required`/`base_url_invalid`) have UI copy via `ERROR_COPY`, exhaustive by `Record` type. OK

### Design comp conformance
Read `DS 补全 - Chat 桌面.html` + `Chat 状态总览.html` directly. Cream press-shadow CTAs match `.btn.cream` (`box-shadow: 0 Npx 0 var(--shadow-3d)` is the in-app convention, 2-3px throughout chat.css). D13/D14 strips follow the "inline warning 提示条" left-border convention (visual comp for D13/D14 is spec-deferred to a design pass — this is the sanctioned interpretation). Vision badge on `--color-walk-*` per ruling. No gold token spent — verified in shipped CSS; all colors `var(--...)` incl. dark-mode variants. ja copy keeps the fox voice (〜だよ/〜てね; 「キーをかくにん中…」).

### Findings

**P2-1 (test robustness, T8-AC9): the design-token grep has demonstrated blind spots.** `SOURCES` omits `TurnFailure.tsx` (hosts the D13 gate + D14 strip) and `ChatInput.tsx` (settings toggle), and `byokCssRules()`'s `\.chat-byok` regex misses `.chat-input__settings` — the gold check too. **Survivor replicated**: adding `bg-white text-red-500` to the D14 strip in `TurnFailure.tsx` plus `var(--color-gold)` to `.chat-input__settings` passes `byok-design-tokens.test.ts` 8/8. Shipped code is clean (verified), but the AC says "applied to the journey surfaces". Fix: add both sources to `SOURCES`; broaden the CSS regex to cover `.chat-input__settings` (or grep the whole appended block).
`apps/web/tests/unit/chat/byok-design-tokens.test.ts`

**P2-2 (reviewability): `return-target.test.ts` contains a raw NUL byte** (the "rejects a path smuggling a raw control character" case embeds a literal 0x00), so git flags the file as binary — its diff is unreviewable in the PR UI and text tooling breaks on it. Use the backslash-u0000 string escape instead; behavior identical.
`apps/web/tests/unit/auth/return-target.test.ts`

**P2-3 (naming): `useSaveSubmit` is not a hook** — it calls no hooks yet carries the `use` prefix and is invoked inline in `SettingsForm`'s JSX, which reads as a hook-rules violation to both humans and lint heuristics. Rename (e.g. `saveSubmitHandler(view)`) or inline the two lines.
`apps/web/src/features/chat/components/ByokSettings.tsx`

**P3-1 (weak assertion, flagged for Tester): the `innerHTML` raw-key assertion is decorative in jsdom** — React sets `<input value>` via the DOM property, which never serializes into `innerHTML`, so that line passes even when the field still holds the key. The adjacent `input.value === ""` assert is the one doing the work. Fine as-is since T6-AC6 is a browser AC — Tester should assert on the real DOM (incl. error state / probe-failure / family-switch residue).
`apps/web/tests/unit/chat/byok-settings-probe.test.tsx`

**P3-2: `useProbeRun` chains `.then` without `.catch`** — the default `runByokProbe` never rejects, but an injected `probe` that throws would strand `phase: "checking"` forever with an unhandled rejection. Cheap hardening: `.catch(() => applyOutcome({ kind: "error" }, sink))`.
`apps/web/src/features/chat/use-byok-settings.ts`

**P3-3: `byokProbeUrl` naive concat** — a trailing-slash `baseUrl` yields `...//v1/byok/probe`. Consistent with existing transport convention, so cosmetic.

**P3-4: `useByokPanel` reads `search.settings` only as `useState` initial** — correct for the fresh-tab magic-link return (full page load), but an in-app client-side nav to `/chat?settings=byok` after mount will not open the panel. No current caller does this; note for when D13's explainer someday deep-links in-app.

**P3-5: `key_invalid` / `base_url_invalid` inline copies are never rendered through the panel in tests** (only `key_required`/`model_required` are). The mapping is type-total so risk is low; a one-liner each would close it.

### Verdict

```json
{
  "verdict": "approve",
  "findings_summary": {"P1": 0, "P2": 3, "P3": 5},
  "quality_ratchet": {
    "ac_total": 13,
    "ac_with_test": 13,
    "missing": [],
    "note": "T6-AC6/AC7 + T8-AC1/AC4 are browser-annotated — unit approximations present in diff; live-run owed to Tester with ANON_ACCESS_ENABLED=true + real mailbox (spec P3 precondition)."
  },
  "gate": {"tests": "1437/1437", "coverage": "98.61/95.31/98.78/99.44", "ratchet": "untouched", "mutations_replicated": "4/4 red", "survivor_found": "1 (token-grep blind spot, P2-1)"}
}
```

None of the P2s touch shipped-code correctness or security; the T14 guard itself is solid and double-applied. Approve, with P2-1/P2-2/P2-3 requested as pre-merge or fast-follow fixes.

— Fable seat, independent (did not read the other seat's review)

lifeodyssey (commented):
## Fable seat — final verification round, HEAD `740c34e9`

Re-verified in a fresh detached worktree. Fix-round diff (`37150858..740c34e9`, 22 files) read in full.

### My P2s — all confirmed landed
- **P2-1 (token grep blind spots): FIXED and mutation-verified.** `SOURCES` now includes `ChatInput.tsx`, `LimitBanner.tsx`, `TurnFailure.tsx`; CSS regex covers `.chat-input__settings`; rule-count floor raised to >=14. **Both of my original survivors replayed and now RED**: `bg-white text-red-500` on the D14 strip -> 1 failed; `var(--color-gold)` on `.chat-input__settings` -> 1 failed.
- **P2-2 (NUL byte): FIXED.** Blob at `740c34e9` scanned byte-by-byte — zero control characters; `file` reports UTF-8 text. (The `Bin` marker in the cross-range diff is only because the *old* side was binary; future diffs will be text.)
- **P2-3: FIXED.** `useSaveSubmit` -> `makeSaveSubmit`.
- Bonus: my P3-2 (`.catch` on probe), P3-3 (`new URL` join), P3-4 (in-app `?settings=byok` nav honoured via effect) all addressed too.

### New surfaces — no regression to the journey triad or T14
- **StoredKeyNotice (P1-1)**: teaser still renders **no key input / no password field** with a credential stored (asserted); clear works without login; notice absent when nothing stored. Touchpoint B semantics intact.
- **AuthCallback `hasReturnIntent` (P1-2)**: derived from the *sanitized* `next` (`!== "/"`), so a T14-rejected `next` cannot force navigation past the blocking retry surface — the dual-intent branch only opens for an already-validated relative target. Three-way behavior tested (replay ok / replay failed + intent / replay failed, no intent).
- **D14 open-settings action (P2-1 Opus)**: `byok.show` wired ChatPage -> TurnFailure; still no generic retry; cream-family button, tokens only (now inside the grep's reach).
- **Probe `definitive` + generation guard**: vision persisted only on a clean `error_code === null` verdict; stale in-flight probes can't paint over a cleared panel; `.catch` closes the stranded-"checking" latent I flagged. Still conforms to #479's pinned `{vision, reachable, error_code}` shape and collapsed taxonomy.
- **`returnTarget` `..` rejection**: dropping the new check turns **2 tests red** (mutation-verified); full T14 vector set + honoured set (incl. `/chat?settings=byok`) re-run green — 69/69 across the journey/T14/deep-link/teaser/error-state suites.
- i18n: `errorCredentialRejected` deduped into `notAccepted` (one D14 message, good); enumeration switched to `Object.values` — future keys are auto-covered, strictly stronger than the hand-list.

### Independent gate at `740c34e9`
typecheck OK · lint:oxlint OK · **1466/1466 passed (198 files, +29)** · coverage **98.56 / 95.36 / 98.65 / 99.45** vs floors 98/95/98/99 — ratchet untouched. All 3 replayed mutations red; worktree left clean.

### Verdict

```json
{
  "verdict": "approve",
  "round": "final",
  "p2_followups": "all 3 landed and independently re-verified",
  "new_surface_regressions": "none found",
  "gate": {"tests": "1466/1466", "coverage": "98.56/95.36/98.65/99.45", "ratchet": "untouched", "mutations_replayed": "3/3 red (2 former survivors + new .. guard)"},
  "remaining_for_tester": "browser ACs unchanged: T6-AC6/AC7 live DOM raw-key assert, T8-AC1 D11 walkthrough, T8-AC4 fresh-tab magic-link live run (ANON_ACCESS_ENABLED=true + real mailbox)"
}
```

**Final: APPROVE.** Green from this seat — clear to merge once the other seat concurs.

— Fable seat, independent

PR #490, opened 2026-07-29
fix(ci): give the changes job full history so push-event path filtering works

lifeodyssey:
## What broke

The `#483` merge to `main` (S1 Iteration 1, 22 PRs) ran **zero test lanes and deployed nothing**. Run [30429636371](https://github.com/lifeodyssey/animichi/actions/runs/30429636371): `changes` failed, and all 12 test lanes plus all 8 deploy jobs reported `skipping`.

## Root cause

`dorny/paths-filter` has two modes. On `pull_request` it calls the GitHub API `listFiles` — that path works and is what `permissions: pull-requests: read` (added 2026-07-28) fixed. On **`push`** it instead diffs the event's `before` SHA against `HEAD` with local git:

```
Changes will be detected between 02cd7fa0… and main
fatal: Not a valid object name 02cd7fa0…^{commit}
git fetch --depth=1 --no-tags origin 02cd7fa0…
fatal: could not read Username for 'https://github.com': No such device or address
##[error]The process '/usr/bin/git' failed with exit code 128
```

Two settings combine: checkout's default `fetch-depth: 1` leaves `before` unreachable, and `persist-credentials: false` (zizmor hardening) denies the action's own fallback fetch. The defect was invisible on every PR because PRs never take the git path — it only fires on push to `main`/`develop`.

## Fix

`fetch-depth: 0` on that one checkout. Keeps `persist-credentials: false`, so the security hardening is not rolled back; full history makes `before` locally reachable, which is the mode paths-filter documents for push events.

`ci.yml` is the only workflow using paths-filter. `actionlint` clean.

## Verification

Cannot be reproduced on a PR by construction (PRs use the API path). The real proof is the next push to `main`: `changes` must succeed and the deploy jobs must run rather than skip. Reviewers should treat "green on this PR" as **no evidence at all** and judge the mechanism.

## Related

Directly enables #484 (the post-deploy gate cannot assert anything if deploys never run) and is a second instance of the failure class in #457 — no roll-up job, so "everything skipped" renders as a mergeable board.

## Summary by CodeRabbit

* **Chores**
  * Improved automated build validation by ensuring complete change history is available during checks.
  * Prevented valid updates from being incorrectly skipped when determining which areas require verification.
  * No changes to product functionality or user-facing behavior.

Issue #485, opened 2026-07-29
P0 ops: no rollback capability on any deploy path


lifeodyssey:
From the 2026-07-29 infra/CI audit.

Nothing in the repo uses `wrangler rollback` or `wrangler versions deploy`, and nothing runs `pulumi stack export` before `pulumi up`. A bad production deploy can only be recovered by pushing a revert through the entire pipeline (CI → staging → approval → prod).

This compounds #<POST_STAGING_ISSUE> (the promotion gate is vacuous), so bad deploys *will* reach prod.

`preview.yml` already uses `wrangler versions upload/list`, so the primitive is understood in this repo — the same mechanism gives instant rollback. Scope: (a) document/script a one-command rollback per component, (b) `pulumi stack export > backup` as a pre-`up` step, (c) note the rollback command in `docs/ops/deployment.md`.

lifeodyssey:
完成:#762 落地 rollback.yml(workflow_dispatch + production 审批 + 与部署互斥的并发组 + --env production)与 runbook CI 小节;Pulumi 备份半边此前已在 reusable-deploy-component.yml;#751 挂账的测试重构一并偿还。

PR #492, opened 2026-07-29
fix(ci): rollback capability + TURNSTILE/ANON_ID secret injection chain (#485)

lifeodyssey:
## Summary

Two related CI/ops fixes from the 2026-07-29 infra/CI audit:

**Secret injection chain (blocking staging anonymous surface)**
- `TURNSTILE_SECRET` existed in GH repo secrets and the default `animichi` Worker, but not in `animichi-staging` — none of the 14 workflows even grepped for "turnstile", so the GH→CF path didn't exist.
- `ANON_ID_SECRET` has the same gap and, per a coordinator check during this work, **isn't in GH repo secrets at all yet** — this PR's job-failure assertion will now surface that gap loudly on the next deploy instead of silently shipping a degraded Worker.
- Added a **list-driven** `post_deploy_secrets` step to `_deploy-component.yml` (and mirrored it in `deploy.yml`'s manual path) that runs a `wrangler secret put --env <target>` loop **strictly after** the Wrangler deploy step, fed from stdin. Order is safety-critical: `wrangler secret put` in non-interactive CI silently answers "yes" to "Worker does not exist, create it?", which has already produced one orphaned shell Worker in this account (confirmed with the coordinator: `animichi-staging` doesn't exist yet either, pending #490 for the `changes`-job push-event bug — so this ordering matters for the very first deploy, not just steady state).
- Any secret name in the list that resolves empty fails the job (no silent skip).
- Adding a future post-deploy secret = 3 touch points: the caller's `post_deploy_secrets` list, the reusable workflow's `secrets:` map, and one `env:` line in the push step (GitHub Actions has no dynamic `secrets.<computed-name>` lookup, so that last line is unavoidable).

**Rollback capability (#485)**
- Every deploy step is already a plain `wrangler deploy` (not `versions upload`), but Cloudflare still versions it — so `wrangler rollback` / `wrangler deployments list` work today with **no change to the deploy step itself**. Rewrote `docs/ops/deployment.md`'s Rollback section into a one-command-per-component table (root/catalog/users/web), a Pulumi restore procedure, and an explicit warning that none of this undoes Atlas/Supabase migrations.
- Added `pulumi stack export` (uploaded as a 30-day-retention workflow artifact) immediately before every `pulumi up` in `_deploy-component.yml`.
- Deliberately did **not** add a `workflow_dispatch` rollback trigger — the manual `wrangler rollback` commands already are the one-command primitive the issue asked for, and wrapping them in an unexercised dispatch workflow (component + version-id inputs) would add untested complexity to an incident-response path that most needs to be simple and trustworthy under pressure. Documented as a follow-up once the manual path has been used for real.

## Safety notes

- **No actual deploy or secret write happens in this PR** — only workflow/doc changes. Real injection happens on the next CI run (owner-controlled).
- `actionlint` and `zizmor` are clean across the full `.github/workflows/` directory (not just touched files).
- No `continue-on-error`, no suppressions.

## Test plan

- [ ] `actionlint .github/workflows/*.yml` — clean (verified locally)
- [ ] `zizmor .github/workflows/*.yml` — clean (verified locally)
- [ ] On next staging deploy: confirm `TURNSTILE_SECRET` job-failure path doesn't trigger (secret exists) and the push step succeeds
- [ ] On next staging deploy: confirm `ANON_ID_SECRET` job **does fail** (secret not yet in GH) — expected, and is the intended signal to provision it
- [ ] After #490 lands and staging deploys resume: verify `animichi-staging` gets both secrets and `wrangler secret list --env staging` shows them

Refs #485.

## Summary by CodeRabbit

- **New Features**
  - Added deployment preflight checks for required security secrets, with clear failures or warnings before deployment.
  - Added automatic post-deployment secret publishing for deployed Workers.
  - Added manual execution options for post-deployment tests across staging and production.

- **Bug Fixes**
  - Scoped database migrations to the intended public schema.
  - Improved deployment reliability by pinning Wrangler and package-manager behavior.

- **Documentation**
  - Expanded deployment, first-time migration, and rollback guidance, including infrastructure-state recovery procedures.

### DIFF
--- a/.github/workflows/_deploy-component.yml
+++ b/.github/workflows/_deploy-component.yml
@@ -91,6 +91,35 @@ jobs:
           atlas migrate apply --dir "file://db/migrations" --url "$NEON_DATABASE_URL" --revisions-schema public
         env:
           NEON_DATABASE_URL: ${{ secrets.NEON_DATABASE_URL }}
+      # Installs the Pulumi CLI onto PATH without running a command (omitting
+      # `command:` is the documented install-only mode) so the plain `pulumi`
+      # steps below can use it.
+      - name: Install Pulumi CLI
+        if: ${{ inputs.run_pulumi }}
+        uses: pulumi/actions@8e5e406f4007fca908480587cb9893c07090f58d # v7.0.0
+      - name: Pulumi stack export (rollback backup)
+        # Snapshot state BEFORE `up` mutates it, so a bad apply can be rolled
+        # back with `pulumi stack import` from the artifact below. #485.
+        if: ${{ inputs.run_pulumi }}
+        working-directory: infra
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          PULUMI_CONFIG_PASSPHRASE: ${{ secrets.PULUMI_CONFIG_PASSPHRASE }}
+          PULUMI_BACKEND_URL: ${{ secrets.PULUMI_BACKEND_URL }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
+          PULUMI_STACK: ${{ inputs.pulumi_stack }}
+        run: |
+          pulumi login "$PULUMI_BACKEND_URL"
+          pulumi stack select "$PULUMI_STACK"
+          pulumi stack export --file "$RUNNER_TEMP/pulumi-$PULUMI_STACK-backup.json"
+      - name: Upload Pulumi stack export
+        if: ${{ inputs.run_pulumi }}
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: pulumi-stack-export-${{ inputs.pulumi_stack }}-${{ github.run_id }}
+          path: ${{ runner.temp }}/pulumi-${{ inputs.pulumi_stack }}-backup.json
+          retention-days: 30
       - name: Pulumi up
         if: ${{ inputs.run_pulumi }}
         uses: pulumi/actions@8e5e406f4007fca908480587cb9893c07090f58d # v7.0.0
--- a/.github/workflows/_deploy-component.yml
+++ b/.github/workflows/_deploy-component.yml
@@ -157,21 +157,50 @@ jobs:
             # token distinguishes "bad secret" (`invalid-input-secret`) from
             # "secret is fine, token just wasn't real" (`invalid-input-response`)
             # without ever needing a real Turnstile widget solve.
+            #
+            # This is deliberately NOT treated with the same fail-closed rule
+            # as the ANON_ACCESS_ENABLED parser above: a local TOML parse
+            # failure is a semantic ambiguity that never self-heals, but a
+            # failed network call to a third party is overwhelmingly a
+            # transient blip — retry with backoff, and only `invalid-input-
+            # secret` (an unambiguous, non-transient verdict) hard-fails.
+            # Exhausting retries with no verdict either way means the
+            # validator was unreachable, which is not evidence the secret
+            # itself is wrong — warn and let the deploy proceed rather than
+            # blocking, say, an unrelated production hotfix (this same check
+            # also runs in deploy.yml's manual production path) on a
+            # siteverify outage that would leave guardTurnstile 403ing every
+            # anonymous request regardless of whether this deploy happens.
             if [ "$name" = "TURNSTILE_SECRET" ] && [ "$anon_flag" = "true" ]; then
-              turnstile_probe=$(curl -sS --max-time 10 \
-                --data-urlencode "secret=$value" \
-                --data-urlencode "response=preflight-probe-not-a-real-token" \
-                https://challenges.cloudflare.com/turnstile/v0/siteverify 2>/dev/null || true)
-              # Never echo $value or $turnstile_probe: the response carries no
-              # secret field, but print only the parsed verdict on principle.
-              if printf '%s' "$turnstile_probe" | grep -q '"invalid-input-secret"'; then
+              turnstile_verdict=""
+              attempt=1
+              while [ "$attempt" -le 3 ]; do
+                # Secret piped via stdin (`secret@-`), never as a curl argv
+                # value — argv can leak through process listings on the
+                # runner; stdin doesn't. Never echo $value or the raw
+                # response: the response carries no secret field, but only
+                # the parsed verdict is printed, on principle.
+                turnstile_probe=$(printf '%s' "$value" | curl -sS --max-time 10 \
+                  --data-urlencode "secret@-" \
+                  --data-urlencode "response=preflight-probe-not-a-real-token" \
+                  https://challenges.cloudflare.com/turnstile/v0/siteverify 2>/dev/null || true)
+                if printf '%s' "$turnstile_probe" | grep -q '"invalid-input-secret"'; then
+                  turnstile_verdict="invalid"
+                  break
+                elif printf '%s' "$turnstile_probe" | grep -q '"invalid-input-response"'; then
+                  turnstile_verdict="valid"
+                  break
+                fi
+                attempt=$((attempt + 1))
+                [ "$attempt" -le 3 ] && sleep $((attempt * 2))
+              done
+              if [ "$turnstile_verdict" = "invalid" ]; then
                 echo "::error::TURNSTILE_SECRET failed Cloudflare siteverify validation for component=$TARGET_COMPONENT, environment=$TARGET_ENVIRONMENT (error-code: invalid-input-secret). This is what happens when the stored value is a Site Key (24 chars, public, meant for the client) instead of the Secret Key (35 chars, private) — or any other string siteverify doesn't recognize as a real secret. Refusing to deploy with it. Re-provision with: gh secret set TURNSTILE_SECRET"
                 missing=1
-              elif printf '%s' "$turnstile_probe" | grep -q '"invalid-input-response"'; then
+              elif [ "$turnstile_verdict" = "valid" ]; then
                 echo "TURNSTILE_SECRET passed Cloudflare siteverify validation for component=$TARGET_COMPONENT, environment=$TARGET_ENVIRONMENT (probe token was correctly rejected as invalid-input-response, meaning the secret itself was accepted)."
               else
-                echo "::error::Could not validate TURNSTILE_SECRET against Cloudflare's siteverify endpoint for component=$TARGET_COMPONENT, environment=$TARGET_ENVIRONMENT — the probe returned neither invalid-input-secret nor invalid-input-response (network failure, endpoint outage, or an unexpected response shape). Same principle as the ANON_ACCESS_ENABLED parser above: ambiguity is fail-closed, not fail-open — refusing to guess. Re-run the deploy; if this persists, check Cloudflare's status page before assuming the secret itself is bad."
-                missing=1
+                echo "::warning::Could not reach Cloudflare's siteverify endpoint to validate TURNSTILE_SECRET after 3 attempts for component=$TARGET_COMPONENT, environment=$TARGET_ENVIRONMENT (network failure, endpoint outage, or an unexpected response shape) — treating this as validator unavailability, not evidence the secret itself is wrong. Proceeding without this specific check; the empty-value check above still applies."
               fi
             fi
           done <<< "$POST_DEPLOY_SECRETS"
@@ -206,62 +235,68 @@ jobs:
         uses: pulumi/actions@8e5e406f4007fca908480587cb9893c07090f58d # v7.0.0
       - name: Pulumi stack export (rollback backup)
         # Snapshot state BEFORE `up` mutates it, so a bad apply can be rolled
-        # back with `pulumi stack import` from the artifact below. #485.
+        # back with `pulumi stack import` from the R2 object this step writes.
+        # #485.
+        #
+        # Deliberately NOT a GitHub Actions artifact: this repo is a PUBLIC
+        # repository (verified — `gh repo view --json visibility`), and a
+        # public repo's workflow artifacts are downloadable by ANY logged-in
+        # GitHub account, not just people with repo access. `infra/index.ts`
+        # exports `cloudflareAccountId`/`cloudflareZoneId`/`webDomain` in
+        # PLAINTEXT and `catalogDatabaseUrl` as `secure:` ciphertext — an
+        # artifact would mean publishing the plaintext identifiers (a
+        # targeted-abuse/social-engineering surface, even though not
+        # credentials themselves) and an offline-crackable Neon connection
+        # string on every catalog deploy. Writing to the SAME R2 bucket the
+        # Pulumi backend already lives in (derived from `PULUMI_BACKEND_URL`,
+        # which is `s3://<bucket>?...` — see `pulumi login` below) keeps this
+        # exactly as private as the Pulumi state itself already is, with zero
+        # new exposure surface, using credentials (`R2_ACCESS_KEY_ID`/
+        # `R2_SECRET_ACCESS_KEY`) already present in this step's env.
+        #
         # NOTE: `run_pulumi` defaults to true and only `component: catalog`
         # jobs leave it at that default — web/users/root all pass
-        # `run_pulumi: false` (see ci.yml). So in practice this step (and the
-        # upload below) only ever runs under the catalog deploy jobs; look
-        # there for the artifact, not under a root/web/users run.
+        # `run_pulumi: false` (see ci.yml). So in practice this step only ever
+        # runs under the catalog deploy jobs; look there for the backup, not
+        # under a root/web/users run.
         if: ${{ inputs.run_pulumi }}
-        id: pulumi_export
         working-directory: infra
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
           PULUMI_CONFIG_PASSPHRASE: ${{ secrets.PULUMI_CONFIG_PASSPHRASE }}
           PULUMI_BACKEND_URL: ${{ secrets.PULUMI_BACKEND_URL }}
           AWS_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: auto
           PULUMI_STACK: ${{ inputs.pulumi_stack }}
+          GITHUB_RUN_ID: ${{ github.run_id }}
         shell: bash
         run: |
           # Same empty-guard pattern as the "Atlas migrate" step above: this
           # secret is `required: false` at the workflow_call boundary, so a
           # caller that omits it must not crash `pulumi login` with an empty
-          # backend URL. This is the ONLY legitimate reason the backup file
-          # doesn't exist afterward — everything below is a hard failure, not
-          # a silent skip, precisely so "Upload Pulumi stack export" below can
-          # tell the two cases apart instead of treating both as fail-open.
+          # backend URL.
           if [ -z "$PULUMI_BACKEND_URL" ]; then
-            echo "PULUMI_BACKEND_URL not set — skipping stack export"
-            echo "skipped=true" >> "$GITHUB_OUTPUT"
-            exit 0
+            echo "PULUMI_BACKEND_URL not set — skipping stack export"; exit 0
           fi
-          echo "skipped=false" >> "$GITHUB_OUTPUT"
           pulumi login "$PULUMI_BACKEND_URL"
           pulumi stack select "$PULUMI_STACK"
-          pulumi stack export --file "$RUNNER_TEMP/pulumi-$PULUMI_STACK-backup.json"
-          if [ ! -s "$RUNNER_TEMP/pulumi-$PULUMI_STACK-backup.json" ]; then
+          backup_file="$RUNNER_TEMP/pulumi-$PULUMI_STACK-backup.json"
+          pulumi stack export --file "$backup_file"
+          if [ ! -s "$backup_file" ]; then
             echo "::error::pulumi stack export produced no (or an empty) backup file despite PULUMI_BACKEND_URL being set. Refusing to silently proceed to 'pulumi up' without a rollback snapshot."
             exit 1
           fi
-      - name: Upload Pulumi stack export
-        if: ${{ inputs.run_pulumi }}
-        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
-        with:
-          name: pulumi-stack-export-${{ inputs.pulumi_stack }}-${{ github.run_id }}
-          path: ${{ runner.temp }}/pulumi-${{ inputs.pulumi_stack }}-backup.json
-          # "warn" is only reachable for the legitimate skip case (no
-          # PULUMI_BACKEND_URL) — the export step above already hard-fails
-          # (before this step even runs) if it ran but produced no file, so
-          # this no longer needs to (and must not) paper over that case too.
-          if-no-files-found: ${{ steps.pulumi_export.outputs.skipped == 'true' && 'warn' || 'error' }}
-          # Rollback windows here are hours-to-days, not weeks — 30 days was
-          # needlessly long for a state dump that (per infra/AGENTS.md) is
-          # sensitive: not `--show-secrets`'d, so config secrets stay
-          # ciphertext, but treat the artifact itself as sensitive anyway and
-          # never paste its contents into an issue/PR. This coexists with,
-          # and does not replace, the R2 backend's own state `.bak` file.
-          retention-days: 7
+          # Same bucket as the Pulumi state backend, different prefix —
+          # `s3://<bucket>?query...` -> just the bucket name, ignoring
+          # whatever query params (endpoint/region/style) happen to be in
+          # this secret's value, since we derive the R2 endpoint from
+          # CLOUDFLARE_ACCOUNT_ID instead of parsing that.
+          bucket="${PULUMI_BACKEND_URL#s3://}"
+          bucket="${bucket%%\?*}"
+          aws s3 cp --endpoint-url "https://${CLOUDFLARE_ACCOUNT_ID}.r2.cloudflarestorage.com" \
+            "$backup_file" "s3://$bucket/rollback-backups/pulumi-$PULUMI_STACK-$GITHUB_RUN_ID.json"
       - name: Pulumi up
         if: ${{ inputs.run_pulumi }}
         uses: pulumi/actions@8e5e406f4007fca908480587cb9893c07090f58d # v7.0.0
--- a/.github/workflows/_deploy-component.yml
+++ b/.github/workflows/_deploy-component.yml
@@ -173,6 +173,7 @@ jobs:
             # anonymous request regardless of whether this deploy happens.
             if [ "$name" = "TURNSTILE_SECRET" ] && [ "$anon_flag" = "true" ]; then
               turnstile_verdict=""
+              shape_seen="false"
               attempt=1
               while [ "$attempt" -le 3 ]; do
                 # Secret piped via stdin (`secret@-`), never as a curl argv
@@ -184,6 +185,13 @@ jobs:
                   --data-urlencode "secret@-" \
                   --data-urlencode "response=preflight-probe-not-a-real-token" \
                   https://challenges.cloudflare.com/turnstile/v0/siteverify 2>/dev/null || true)
+                # Shape sentinel: siteverify's response always has a
+                # top-level "success" boolean today. If we never see it
+                # across every retry, that's not (only) a transient network
+                # blip — it's a sign Cloudflare's response shape changed
+                # under us, which would otherwise degrade this whole check to
+                # a permanently silent warning with no way to notice.
+                printf '%s' "$turnstile_probe" | grep -q '"success"' && shape_seen="true"
                 if printf '%s' "$turnstile_probe" | grep -q '"invalid-input-secret"'; then
                   turnstile_verdict="invalid"
                   break
@@ -199,8 +207,10 @@ jobs:
                 missing=1
               elif [ "$turnstile_verdict" = "valid" ]; then
                 echo "TURNSTILE_SECRET passed Cloudflare siteverify validation for component=$TARGET_COMPONENT, environment=$TARGET_ENVIRONMENT (probe token was correctly rejected as invalid-input-response, meaning the secret itself was accepted)."
+              elif [ "$shape_seen" = "true" ]; then
+                echo "::warning::Could not reach Cloudflare's siteverify endpoint to validate TURNSTILE_SECRET after 3 attempts for component=$TARGET_COMPONENT, environment=$TARGET_ENVIRONMENT (network failure or endpoint outage) — treating this as validator unavailability, not evidence the secret itself is wrong. Proceeding without this specific check; the empty-value check above still applies."
               else
-                echo "::warning::Could not reach Cloudflare's siteverify endpoint to validate TURNSTILE_SECRET after 3 attempts for component=$TARGET_COMPONENT, environment=$TARGET_ENVIRONMENT (network failure, endpoint outage, or an unexpected response shape) — treating this as validator unavailability, not evidence the secret itself is wrong. Proceeding without this specific check; the empty-value check above still applies."
+                echo "::warning::Cloudflare's siteverify response never contained a \"success\" field across 3 attempts for component=$TARGET_COMPONENT, environment=$TARGET_ENVIRONMENT — this looks like the response SHAPE changed, not just a network blip (this check's grep patterns assume today's known shape). Treating as validator-unavailable for now, but this needs someone to look at whether the probe logic itself is stale, not just retry later. Proceeding without this specific check; the empty-value check above still applies."
               fi
             fi
           done <<< "$POST_DEPLOY_SECRETS"
@@ -307,6 +317,16 @@ jobs:
           # backend URL.
           if [ -z "$PULUMI_BACKEND_URL" ]; then
             echo "PULUMI_BACKEND_URL not set — skipping stack export"; exit 0
+          fi
+          # Same reasoning, different failure shape: CLOUDFLARE_ACCOUNT_ID is
+          # also `required: false` at the workflow_call boundary. Without
+          # this guard, an empty value would silently build the endpoint URL
+          # as "https://.r2.cloudflarestorage.com" and let `aws s3 cp` fail
+          # with an opaque DNS/connection error instead of a clear message —
+          # fail-closed either way, but this is the readable version.
+          if [ -z "$CLOUDFLARE_ACCOUNT_ID" ]; then
+            echo "::error::CLOUDFLARE_ACCOUNT_ID not set — cannot derive the R2 endpoint for the rollback backup upload. Refusing to proceed to 'pulumi up' without a rollback snapshot."
+            exit 1
           fi
           pulumi login "$PULUMI_BACKEND_URL"
           pulumi stack select "$PULUMI_STACK"
--- a/.github/workflows/_deploy-component.yml
+++ b/.github/workflows/_deploy-component.yml
@@ -143,6 +143,36 @@ jobs:
               else
                 echo "::warning::$name is empty/unset for component=$TARGET_COMPONENT, environment=$TARGET_ENVIRONMENT, but ANON_ACCESS_ENABLED is not \"true\" there, so this environment's code path never reads it (see worker/auth.ts:anonymousEnabled). Soft warning only; the post-deploy push step will skip it."
               fi
+              continue
+            fi
+            # TURNSTILE_SECRET-specific: a GitHub secret is write-only — once
+            # set, nobody (not even this workflow, on a later run) can read it
+            # back to eyeball whether it's actually the Turnstile Secret Key
+            # (35 chars) rather than the public Site Key (24 chars, meant to
+            # ship to the client — see apps/web/src/components/
+            # TurnstileGate.tsx). This step is the only place that ever holds
+            # the raw value, so it's the only place that can ever answer that
+            # question. Cloudflare's siteverify endpoint doubles as a
+            # validity check: probing it with a deliberately-wrong response
+            # token distinguishes "bad secret" (`invalid-input-secret`) from
+            # "secret is fine, token just wasn't real" (`invalid-input-response`)
+            # without ever needing a real Turnstile widget solve.
+            if [ "$name" = "TURNSTILE_SECRET" ] && [ "$anon_flag" = "true" ]; then
+              turnstile_probe=$(curl -sS --max-time 10 \
+                --data-urlencode "secret=$value" \
+                --data-urlencode "response=preflight-probe-not-a-real-token" \
+                https://challenges.cloudflare.com/turnstile/v0/siteverify 2>/dev/null || true)
+              # Never echo $value or $turnstile_probe: the response carries no
+              # secret field, but print only the parsed verdict on principle.
+              if printf '%s' "$turnstile_probe" | grep -q '"invalid-input-secret"'; then
+                echo "::error::TURNSTILE_SECRET failed Cloudflare siteverify validation for component=$TARGET_COMPONENT, environment=$TARGET_ENVIRONMENT (error-code: invalid-input-secret). This is what happens when the stored value is a Site Key (24 chars, public, meant for the client) instead of the Secret Key (35 chars, private) — or any other string siteverify doesn't recognize as a real secret. Refusing to deploy with it. Re-provision with: gh secret set TURNSTILE_SECRET"
+                missing=1
+              elif printf '%s' "$turnstile_probe" | grep -q '"invalid-input-response"'; then
+                echo "TURNSTILE_SECRET passed Cloudflare siteverify validation for component=$TARGET_COMPONENT, environment=$TARGET_ENVIRONMENT (probe token was correctly rejected as invalid-input-response, meaning the secret itself was accepted)."
+              else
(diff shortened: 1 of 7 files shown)


lifeodyssey:
# 独立评审(Fable 席)— PR #492

结论:**REQUEST CHANGES**。方向对(先 deploy 后 put、缺失即红、导出 state 做回滚锚点都值得要),但新加的 secret 推送步骤**缺认证环境变量,当前 100% 必挂**,且"缺 secret 即失败"的断言放在 deploy 之后,和它自己的错误文案互相矛盾。逐条如下。

---

## P1

### P1-1 「Push post-deploy secrets」步骤没有 `CLOUDFLARE_API_TOKEN` / `CLOUDFLARE_ACCOUNT_ID` — 每次 root 部署都会硬失败

`_deploy-component.yml` 和 `deploy.yml` 两处新步骤的 `env:` 都只有 `TARGET_*`、`POST_DEPLOY_SECRETS` 和两个 secret 值。上一步 `cloudflare/wrangler-action` 的 `apiToken:` 是 action 进程内部的,不会外泄给后续 step。裸跑 `npx wrangler@4.79.0 secret put` 在无 token 的非交互环境下直接认证失败 → 即使 GH secrets 全部配好,root staging/prod 的每一次部署都会在这一步红掉,并连带 block `post-staging → deploy-prod` 整条链。actionlint/zizmor 都不查运行期 env,所以"本地验证 clean"不能覆盖这一点。

修复:两处步骤各加
```yaml
CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
```
(account id 不是严格必需,但 token 若可见多账户 wrangler 会拒绝推断,补上才稳。)

### P1-2 断言放在 deploy 之后 = 错误文案说谎;缺失的 `ANON_ID_SECRET` 会把 staging Worker 无限期留在降级态

错误信息写的是 "Refusing to leave the Worker running without it instead of deploying silently degraded",但检查发生在 **deploy 成功之后**:job 失败并不会撤销 deploy。按 PR 自述,`ANON_ID_SECRET` 目前**不在** GH secrets 里,所以下一次 staging 部署的实际时间线是:首次创建 `animichi-staging`(`[env.staging]` 无 routes、workers_dev 默认开启,`ANON_ACCESS_ENABLED="true"`)→ 步骤失败 → Worker 带着缺失的 TURNSTILE/ANON_ID secret **一直在线**,直到有人补 secret 并重跑。这不是几秒的窗口,是无限期。(行为上不算最糟:`worker/auth.ts` 里 `ANON_ID_SECRET` 缺失时 `anonymousEnabled()` 为 false、匿名面回落 401,不是 500;但仍与文案宣称的语义相反。)

修复(保持 put 在 deploy 之后不变,这是对的):把**空值断言**拆出来,放到 "Deploy Worker" **之前**做一个 preflight 步骤(只查 `[ -n "$X" ]`,不碰 wrangler、不打印值)。这样:secret 缺失 → 根本不部署,文案成立;secret 齐全 → 降级窗口收缩到 deploy 与 put 之间的几秒,可接受。另:wrangler 2026-03 起支持 wrangler 配置里的 `secrets.required`(deploy 时校验目标 Worker 已配齐,缺失即 deploy 失败),可作为稳态守卫叠加,但注意它对"首次部署 + post-deploy put"有鸡生蛋问题,只能在 bootstrap 完成后启用。

### 失败语义链(核验通过,非问题)

`post-staging` `needs: [... deploy-root-staging]` 无 `if: always()`,`deploy-prod` `needs: post-staging`;`deploy.yml` 是单 job 顺序步骤。新步骤失败 → 下游全 skip、run 红。无 `continue-on-error`、无静默跳过。这条链是干净的。

---

## P2

### P2-1 凭证泄漏面 — 结论:当前不构成泄漏,但要把约束钉死

- **注入路径**:`printf '%s' "$value" | wrangler secret put` 走 stdin,无命令行参数、无 `set -x`、`::error::` 只含 secret **名**不含值,GH 对 `secrets.*` 值有日志掩码。这部分做得对。
- **Pulumi artifact(重点核验项)**:`pulumi stack export` 不带 `--show-secrets` 时,secret 以 ciphertext 落盘。当前 state 的敏感面只有 `catalogDatabaseUrl`,它在 `Pulumi.prod.yaml` 里本就是 `secure:` 加密配置;资源只有 R2 bucket + WorkersRoute,均非敏感。仓库为 **PRIVATE**,artifact 仅 repo 读权限可下载。**所以现状不等于把敏感数据放进可下载产物。** 但两个前提是脆弱的:① 未来任何人往 `infra/index.ts` 加一个把敏感值传给**未标记 secret** 的 resource 属性(Pulumi 只对 `getSecret`/provider 标记的字段加密),明文就会进入 30 天可下载的 artifact,无任何门禁会拦;② 持有 `PULUMI_CONFIG_PASSPHRASE` 的人 + artifact = 可解密。要求:在 export 步骤加注释明确"**禁止** `--show-secrets`;state 中新增敏感值必须走 `config.getSecret`/`pulumi.secret()`",并把这条写进 `infra/AGENTS.md`。

### P2-2 wrangler-action 自身的 `secrets:` 在 deploy **之前**推 — PR 的顺序不变量在同一个 job 里被上一步破坏

核验了 pin 的 v4.0.0(`ebbaa15`,与 v4.0.0 tag 一致)dist 源码:`main()` 顺序是 `uploadSecrets(...)` → `wranglerCommands(...)`,即 `wrangler secret bulk --env staging` 跑在 deploy 之前。首次 staging 部署时 `animichi-staging` 尚不存在,既有 `worker_secrets` 的 10 个 secret 会先对不存在的 Worker 做 secret bulk——正是本 PR 反复强调要避免的 fallback-to-yes 路径。多数情况下紧随其后的真 deploy 会覆盖掉 shell worker,但若 deploy 步骤因构建等原因失败,就又留下一个空壳。至少要修正 `_deploy-component.yml` 里"Deliberately NOT part of the wrangler-action secrets list"注释的表述(它暗示顺序问题已被处理);更彻底的做法是把 root 的 `worker_secrets` 也迁进 post-deploy 循环(P1-1 修好后该循环即可承接)。可另开 issue,但注释必须在本 PR 改准。

### P2-3 回滚方案可行性 — 大体成立,两处事实错误见 P3

`wrangler deploy` 产生的每次部署确实可被 `wrangler rollback` 回滚(官方 Rollbacks 文档);containers 场景:镜像由 CF Registry 保留,回滚到引用旧镜像的版本**可用**,唯一坑是 `wrangler containers delete` 删过镜像后旧版本回滚会失效(官方 Limits 脚注)——文档应加一行"可能回滚期内不要清理容器镜像"。另一个未写的硬限制:**rollback 不能跨越 Durable Object migration 边界**(官方明文)。root worker 已有 `[[migrations]]` v1/v2,未来任何 v3 migration 一落地,跨它的回滚会被平台直接拒绝——这比"DB 迁移不回滚"更隐蔽,警告章节应补。

---

## P3

- **docs/ops/deployment.md 表格用词**:`wrangler rollback [deployment-id]` 配 `wrangler deployments list` 不准确——wrangler v4 的 rollback 接受的是 **version id**(来自 `wrangler versions list`);incident 压力下让人拿着 deployment id 喂 rollback 会白白折腾。
- **Pulumi 恢复步骤取错 run**:文档说从"坏 run 的**前一个** run"下载 artifact,但 export 跑在 `pulumi up` **之前**,坏 run 自己的 artifact 才是未被污染的 pre-mutation 快照,而且更新。应改为"从坏 run 本身下载"。
- **"3 处改动结构上不可避免"——多算了一处**:`secrets.<动态名>` 确实不存在,env map 那一行躲不掉;但 caller 侧改用 `secrets: inherit`(同仓库内部 reusable workflow 的常规用法)可消掉 reusable workflow 的 `secrets:` 声明处,降到 2 处。`toJSON(secrets)` 塞单个 env 再 jq 查值能降到 1 处,但会把全部 secrets 的 JSON 团进一个进程 env,泄漏面更大,不采纳是对的。建议要么改 inherit,要么把 PR 描述里"3 处不可避免"的措辞校正为工程取舍。
- 不加 `workflow_dispatch` 回滚壳的决定:同意,理由充分。

---

## 结论

**REQUEST CHANGES**。P1-1(缺 CF 认证 env,步骤必挂且 block 全链)+ P1-2(断言应前置到 deploy 之前,当前文案与实际语义相反)必须修;P2-2 的注释表述需在本 PR 改准。P2-1 结论:Pulumi artifact 当前**不**构成敏感数据外泄(私有仓 + ciphertext + 无敏感资源),但需按上文钉死约束。

lifeodyssey:
Pushed a fix-up commit (4222d4f8) addressing both review passes. Disposition per item:

## P1 — fixed

- **CLOUDFLARE_API_TOKEN / CLOUDFLARE_ACCOUNT_ID missing** (Fable + Opus, same finding): added to the push step's `env:` in both `_deploy-component.yml` and `deploy.yml`.
- **Production hard-fails on a secret it doesn't read** (Opus): split into a preflight step (before deploy) that reads `ANON_ACCESS_ENABLED` for the target environment straight out of the checked-out `wrangler.toml`. Production = `"false"` -> soft warning + skip (`worker/auth.ts` `anonymousEnabled()` requires flag AND secret, confirmed by reading the source). Staging = `"true"` -> still hard-fails. The push step no longer duplicates the fatal/non-fatal decision -- it just skips empties the preflight already cleared.
- **Rollback doc step 4 unexecutable** (Opus): added `workflow_dispatch` to `_post-deploy-test.yml` (environment/suite as `choice` inputs). Doc now says explicitly that the suites are still TODO no-ops (per the file's own header) pending #493, so a green re-run isn't verification yet -- worded as "will become real" rather than claimed as working today.

## P2 — fixed

- `wrangler rollback` takes a **version id** from `wrangler versions list`, not a deployment id from `deployments list` -- verified against installed wrangler 4.110.0 `--help` output directly. Doc rewritten, plus added the required `-y -m "<reason>"` flags (rollback prompts interactively and hangs non-TTY without them).
- Root gets its own rollback warning: two DO migrations (v1/v2) + a container image, neither of which `wrangler rollback` restores. Flagged as needing manual verification, not a one-liner, unlike catalog/users/web.
- Pulumi artifact section now says explicitly it only exists under catalog's job (only component leaving `run_pulumi` at its `true` default) and must be downloaded from the *same* run that did the bad apply (export runs before `up`, same run), not the prior one. Retention 30 to 7 days, with a sensitivity note (don't paste into issues/PRs).
- Added the empty-`PULUMI_BACKEND_URL` early-exit guard to the export step (matches the existing Atlas-migrate pattern), plus `if-no-files-found: warn` on the upload so a skipped export doesn't hard-fail the job.
- Unified wrangler version in the new steps to 4.112.0 (matches `workers/*/package.json` and `apps/web/package.json`'s actual pin) instead of the unrelated 4.79.0 the wrangler-action step happens to use.
- Added explicit `shell: bash` to both new `run:` blocks (indirect expansion `${!name}` is bash-only) and added `TARGET_COMPONENT` to `deploy.yml`'s copy to stop the drift you flagged.
- `infra/AGENTS.md`: documented the export-artifact constraint (no `--show-secrets`, new sensitive config must go through `requireSecret()`/`getSecret()`).

## Corrected in the review itself (no code change)

- **wrangler-action ordering claim was backwards**: I pulled `dist/index.mjs` at the pinned SHA -- `uploadSecrets()` (the `worker_secrets` bulk-put) runs **before** the `deploy` command, not after. Rewrote the comment to say this plainly: our new step's after-deploy placement doesn't retroactively fix that pre-existing mechanism, it only avoids adding a second instance of the same risk.
- **"Turnstile is DORMANT" -- checked and this does not hold on this branch.** `worker/app.ts` on `origin/main` (what this PR is based on) already has it ARMED via #447: `handleAnonymousV1` calls `guardTurnstile`, which fails **closed** (403) when `TURNSTILE_SECRET` is empty (`worker/turnstile.ts` `usableSecret`+`guardTurnstile`). I grepped the actual file content at `origin/main` and there is no "DORMANT" comment there -- whatever both reviews read it from isn't this branch's `worker/app.ts`. No change made to that file since it's correct and out of this PR's scope either way.
- Also verified the asymmetry Opus called out: `ANON_ID_SECRET` missing -> `resolveAnonymous` returns `null` -> falls through to the ordinary 401 path (no crash, no 403). `TURNSTILE_SECRET` missing only matters once `ANON_ID_SECRET` is *also* present and `ANON_ACCESS_ENABLED=true` -- today, with neither secret provisioned on staging, anonymous access is already fully closed via the `ANON_ID_SECRET` gate regardless of Turnstile.

## Not adopted

- **`secrets: inherit`** (Fable): would drop the reusable workflow's `secrets:` map, cutting 3 touchpoints to 2 for a future post-deploy secret. Rejected -- it would also hand catalog/users/web jobs implicit access to every secret the shared workflow declares (MIMO_API_KEY, SUPABASE_SERVICE_ROLE_KEY, etc.), widening blast radius for components that don't need them. Touchpoint reduction doesn't outweigh losing per-job secret scoping.
- **A dedicated `workflow_dispatch` rollback workflow** (component + version-id inputs): still not added, per the original PR's reasoning -- `wrangler rollback`/`versions list` are already the one-command primitive; a bespoke dispatch workflow would add untested complexity (invalid version ids, wrong environment, the DO-migration/container caveat now documented) to an incident-response path that most needs to stay simple.

actionlint + zizmor both clean across the full `.github/workflows/` directory after this commit.

lifeodyssey:
# 复核轮(Fable 席)— PR #492 fix-up `4222d4f8`

先接受一条对我上轮的更正:关于 Turnstile 的运行时影响,真实情况是 `handleAnonymousV1` 已接线 `guardTurnstile`,`worker/turnstile.ts` 在 secret 不可用时 **403 fail-closed** —— 即 `ANON_ID_SECRET` 在而 `TURNSTILE_SECRET` 缺时,staging 匿名面全量 403(我上轮只覆盖了 `ANON_ID_SECRET` 缺失 → 401 那条路径,评估不完整)。这让 preflight 断言比我上轮认定的更加 load-bearing,新错误文案对此的表述是准确的。

## 逐条判定

| 上轮意见 | 判定 |
|---|---|
| P1-1 push 步骤缺 `CLOUDFLARE_API_TOKEN`/`CLOUDFLARE_ACCOUNT_ID` | **已修** — 两处均补齐,核对无误 |
| P1-2 断言应前置 deploy 之前 | **修对了一半,见下方唯一遗留项** |
| P2-1 Pulumi artifact 约束钉死 | **已修** — `infra/AGENTS.md` 写入 `--show-secrets` 禁令 + `getSecret()` 强制;retention 30→7;docs 补"artifact 视为敏感 / 只在 catalog job 下 / 从坏 run 自己下载"。全部正确 |
| P2-2 wrangler-action 顺序注释失实 | **已修** — 注释按 dist 源码事实重写,不再宣称不变量已被处理 |
| P2-3/P3 回滚文档(version id、`-y -m`、DO migration、容器镜像、坏 run artifact 取向) | **已修** — 全部落实且比要求的更完整(root 单列警告、post-deploy-test 尚是 no-op 的诚实披露都加了) |
| P3 `secrets: inherit` 降触点 | **不采纳,理由成立** — inherit 会把全部 secrets 暴露给 catalog/web/users 的调用,blast radius 论证我接受,撤回该建议 |

## 唯一遗留项(P1):preflight 的 TOML flag 解析是 fail-open —— 恰好是"解析不出来就当 false"的实现

我在 PR head 的 `wrangler.toml` 上实测了这段 awk:

- **当前文件:正确。** staging → `true`(硬门生效)、production → `false`(warn-skip,Opus 席那条"生产别被卡死"成立)。
- **漂移变体:全部静默归零。** 以下每一种都让 `anon_flag` 解析为空 → 走 `::warning::` + skip,硬门无声消失:
  - `ANON_ACCESS_ENABLED = 'true'`(单引号是合法 TOML literal string;`-F'"'` 切不出 $2)→ 空
  - `ANON_ACCESS_ENABLED = true`(无引号布尔)→ 空
  - `[env.staging.vars]` 节头带尾随空格(`$0 == want` 精确匹配失败)→ 空
  - root 将来迁到 `wrangler.jsonc`(`apps/web` 已经是 jsonc,这不是假想)→ `2>/dev/null || true` → 空
  - awk 本身任何失败 → `|| true` → 空

空值与 `"false"` 在分支里不可区分,于是**每一种解析失败都把硬断言降级成一条没人看的 warning** —— 这正是本仓库的既有病症("跳过/静默失败被渲染成没问题"),也正是协调人点名的红线实现。

**修法(3 行,保持 Opus 席的性质不变)**:解析后要求 `anon_flag` 必须是 `"true"` 或 `"false"` 之一,否则 `::error::"cannot determine ANON_ACCESS_ENABLED for env=$TARGET_ENVIRONMENT from wrangler.toml — refusing to guess (fail-closed)"` 并 `exit 1`。生产今天解析出 `"false"`,照旧不被卡;任何格式漂移则当场爆红,而不是把门变成摆设。两处(`_deploy-component.yml` + `deploy.yml`)同改;顺带这两份 awk 是复制粘贴,漂移风险请在注释里互相指认(两份尚可接受,第三份就该提取)。

## 回修引入的新问题(P3,不阻塞)

- **"wrangler 版本统一到 4.112.0"未完成**:两个 put 步骤和 docs 用了 `wrangler@4.112.0`,但 4 处 wrangler-action `wranglerVersion: "4.79.0"` 原封未动 —— 现在是同一个 job 里 deploy 用 4.79.0、secret put 用 4.112.0 的**版本劈叉**,比统一前更不一致。要么把 `wranglerVersion` 一并提到 4.112.0,要么在 PR 描述里撤回"统一"的说法。
- `if-no-files-found: warn`:对"无 PULUMI_BACKEND_URL 跳过 export"的场景是对的,但它同时掩盖了"export 跑了却没产出文件"的异常(回滚备份静默丢失)。可用 step output 标记 export 是否真跑、upload 的 `if:` 挂在它上,后续 PR 收紧即可。

## 结论

**仍 REQUEST CHANGES**,单点:preflight flag 解析改为 ambiguity 即 fail-closed(上述 3 行改法)。此点落地后,本席其余意见全部关闭,即转 APPROVE;P3 两条可随手带上或另开 issue。

lifeodyssey:
Pushed dfe8c793 addressing the three remaining blockers from this round (Fable + Opus, same three items):

## Fixed

1. **awk flag parser was fail-open on ambiguity.** The preflight now hard-errors (`::error::` + `exit 1`) unless the parsed `ANON_ACCESS_ENABLED` value is exactly `"true"` or `"false"` — single-quoted TOML, an unquoted bool, a trailing-space section header, or a config that moved to `wrangler.jsonc` now stop the deploy instead of silently reading as `""` and falling into the soft-warn branch. Also removed the `component == 'root'` gating around computing/checking the flag — per Opus's point, that gating meant every *other* component was permanently stuck on soft-warn with no way to ever hard-fail. It's currently dead code (only root populates `post_deploy_secrets` today) but the check itself shouldn't depend on the component name to be correct. Verified both the current-file case (staging→`true`, production→`false`, no error) and drift cases (single-quote, unquoted, missing file) all correctly hard-fail now.

2. **Follow-up issue number.** `docs/ops/deployment.md` § "Automating the rollback trigger" now references **#496** for rollback automation (including the phone-only-incident case the manual-first tradeoff doesn't cover), replacing the vague "revisit only after..." line. Left the existing **#493** reference (post-deploy suite wiring) untouched and distinct — confirmed it's not the same follow-up.

3. **Local Cloudflare token prerequisites.** Added a new "Prerequisite: a local Cloudflare API token, provisioned BEFORE an incident" section: scope (`Workers Scripts:Edit`, same as CI's `CLOUDFLARE_API_TOKEN`), where to create it (dashboard → API tokens) and store it (password manager/keychain, exported to the shell for the duration, never committed), and an explicit "verify it works now, not during the incident" step (`wrangler whoami`, then a read-only `wrangler deployments list` against a real component).

## Also fixed while in there

- **Preflight step moved** to run immediately after `./.github/actions/setup`, before both "Build component" and "Atlas migrate" (it was previously placed after Atlas migrate). Its own comment claimed "validates BEFORE anything deploys," which was true for the Worker but not for the DB migration — now it actually is, and the comment says so accurately.
- **Pulumi export/upload fail-open gap tightened**: the export step now emits a `skipped` output and hard-fails itself if it ran (`PULUMI_BACKEND_URL` was set) but produced no/an empty backup file, rather than leaving that ambiguity to the upload step's `if-no-files-found`. The upload step's `if-no-files-found` is now `${{ steps.pulumi_export.outputs.skipped == 'true' && 'warn' || 'error' }}` — `warn` is only reachable for the legitimate, expected skip case now.
- **wrangler version consistency** (flagged as inconsistent in the review): re-checked — the prior commit had already unified every `wranglerVersion`/`npx wrangler` reference to `4.112.0` across both files before this round's message arrived; confirmed zero remaining `4.79.0` references with a fresh grep.

`actionlint` and `zizmor` both clean across the full `.github/workflows/` directory after this commit.

lifeodyssey:
Pushed 5491b45f — added the TURNSTILE_SECRET content validation the coordinator requested (Site Key vs Secret Key can't be told apart once a GitHub secret is set, and both a wrong-type secret and a missing secret produce the identical symptom: every anonymous request 403s).

## What changed

Extended the existing TURNSTILE_SECRET handling inside the preflight step's per-name loop (still gated on `ANON_ACCESS_ENABLED == "true"` for the target environment — same condition as the existing empty-value check, so production isn't touched by this either) with a probe against Cloudflare's siteverify endpoint using a deliberately-invalid response token:

- `invalid-input-secret` → the stored value isn't a real Turnstile secret key at all (wrong key type, e.g. a 24-char Site Key instead of the 35-char Secret Key, or garbage) → hard fail, `::error::` + non-zero exit
- `invalid-input-response` → the secret itself was accepted, only the fake probe token was rejected (expected) → pass
- anything else (network failure, Cloudflare outage, unexpected response shape) → fail-closed, same principle as the `ANON_ACCESS_ENABLED` parser fix from the prior round: ambiguity is not treated as a pass

Constraints honored:
- Never echoes the secret value or the full response body — only the parsed error-code verdict makes it into logs (the siteverify response has no secret field regardless, but kept the discipline).
- `curl --max-time 10` bounds the probe so a Cloudflare outage can't hang the deploy job.
- Mirrored into `deploy.yml`'s copy of the preflight step for parity (currently dead code there too, same as the rest of the TURNSTILE_SECRET/ANON_ID_SECRET checks, since production's `ANON_ACCESS_ENABLED` is `"false"`).

## How I verified it (asked for in the request)

Ran the actual siteverify endpoint directly, then reproduced the exact bash logic transplanted from the workflow file (not just the curl call in isolation) as two local dry runs:

1. **Hard-fail branch**: a garbage string (`clearly-not-a-real-turnstile-secret-key`) as `TURNSTILE_SECRET` → siteverify returns `{"error-codes":["invalid-input-secret"],...}` → the transplanted loop logic correctly sets `missing=1` and would `exit 1`.
2. **Pass branch**: Cloudflare's documented "always-fails-response" test key (`2x0000000000000000000000000000000AA`, a public, intentionally-published testing secret — not sensitive) → siteverify returns `{"error-codes":["invalid-input-response"],...}` for any response token → the transplanted loop logic correctly logs the pass message and leaves `missing=0`.

Both runs are in this session's transcript; happy to paste the exact commands if useful for the record, but didn't want to put probe output in the PR body given the "don't print full response body" discipline extends to not needlessly pasting Cloudflare API output either.

`actionlint` and `zizmor` clean.

Ready for the final joint re-review whenever both seats are available.

lifeodyssey:
# 最终复核(Fable 席)— PR #492,批次 `189eecd7` + `a9737d3f`

## 判定:**APPROVE**

## 逐项核验

### 1. fail-closed 解析 — 已封死,四类漂移逐一验证

新增的严格检查(`anon_flag` 必须精确等于 `"true"` 或 `"false"`,否则 `::error::` + exit 1)对我上轮实测的四类漂移全部有效:单引号 TOML、无引号布尔、节头尾随空格、配置迁 jsonc(文件不存在)——每一类都解析为空串 → 落进严格检查 → 硬失败,不再有任何通往 soft-warn 的静默路径。附带收益:awk 若因重复 key 打出多行,同样不等于 `"true"` → 硬失败。两份拷贝(`_deploy-component.yml` + `deploy.yml`)一致。注释把"只认一种精确 TOML 形状、漂移必须爆红不许猜"写得很清楚。**关闭。**

### 2. siteverify 探测 — 自身不是 fail-open,核验通过

逐分支追了:

- **网络故障/超时**:`curl --max-time 10 ... || true` → `turnstile_probe=""` → 不匹配任何一个 error-code → 落 else 分支 → `::error::` + `missing=1`。fail-closed ✓(`--max-time` 有,10s)。
- **Cloudflare 限流/5xx/返回体形状变化**:同上,匹配不到两个已知码 → else → 硬失败 ✓。
- **两码同现**:`invalid-input-secret` 先查,优先判坏 secret,优先级正确 ✓。
- **日志泄漏**:secret 经 `--data-urlencode` 进 POST body,响应体 `$turnstile_probe` 从不 echo,判定消息只含组件/环境/error-code 名,无值。✓
- **探测只在 `anon_flag == "true"` 且值非空时跑**:prod(false)不探测,不给生产部署引入对 siteverify 的可用性依赖。✓
- 空值的 `continue` 在探测之前,不会拿空串去探。✓

一条 **P3(不阻塞)**:`--data-urlencode "secret=$value"` 会让 secret 进入 curl 的 **argv**(runner 上 /proc 可见)。GH runner 是一次性单租户 VM,实际风险接近零,但与本 PR 其余部分的 stdin 纪律(`printf | wrangler secret put`)不一致——后续可改为 `--data-urlencode secret@-` 从 stdin 喂。随手或不改都行。

### 3. 移除 `component == 'root'` 门 — 正确,无副作用

实现者的自查是对的:留着那层门,非 root 组件的 `anon_flag` 永远为空 → 在旧实现里永远 soft-warn(静默解除武装),在新实现里则会误报解析失败。移除后:今天只有 root 传 `post_deploy_secrets`(step 级 `if:` 挡住其他组件),无实际行为变化;将来任何组件加了 `post_deploy_secrets`,会在它自己的 `working_directory` 里解析——`apps/web` 是 wrangler.jsonc → 直接硬失败,逼维护者显式扩展解析器而不是静默放行。这正是想要的性质。**关闭。**

### 4. 上轮两条 P3 — 均已处理,且第二条比我要求的更彻底

- **wranglerVersion 统一**:两文件共 4 处 `4.79.0` → `4.112.0`,与两个 put 步骤及 docs 一致,劈叉消除。✓
- **`if-no-files-found`**:改为 `steps.pulumi_export.outputs.skipped == 'true' && 'warn' || 'error'`,并且 export 步骤在 backend URL 存在但产出空文件时直接 `exit 1`(在 upload 之前就拦住)。边界也对:`run_pulumi: false` 时两步都 skip,表达式退化无影响。"warn 只可达于合法跳过分支"的注释准确。✓

### 5. 文档批次

- 本地 CF token 前置条件章节(`Workers Scripts:Edit` scope、密码管理器存放、事故前 `wrangler whoami` 演练)务实、正确。
- #496(回滚自动化,含手机场景)与 #493(post-deploy 真断言)均已开且 OPEN,引用属实。
- 一条**微 P3**:前置条件第 3 步的验证命令写的是 `wrangler deployments list --env staging`——命令本身存在且只读、能用,但和上文表格刚统一成的 `versions list` 不一致,顺手改掉更好。

### 6. 真实 CI

PR head 检查 **15/15 全绿**,零失败。

## 结论

三轮下来所有 P1/P2 全部关闭:认证 env、preflight 前置(现在更是提到 Atlas migrate 之前,DB 副作用也挡住了)、fail-closed 解析、探测自身 fail-closed、artifact 语义、文档事实。剩余两条 P3(curl argv、`deployments list` 措辞)不阻塞,可随手带或并入 #496。**APPROVE。**

lifeodyssey:
Pushed two commits addressing #516, both to `fix/485-rollback-and-secret-chain` as requested.

## Failure 2/2 (wrangler-action + pnpm workspace) — FIXED, pushed (4a39ec0c)

Confirmed root cause the same way you asked — read `cloudflare/wrangler-action`'s `dist/index.mjs` at the pinned SHA, not the docs:

- `detectPackageManager()` only looks for a lockfile **inside** the step's `workingDirectory`. `apps/web`, `workers/catalog`, `workers/users` are all pnpm workspace members whose lockfile lives at the repo root, not in their own directory — so detection always silently falls back to `"npm"` for all three.
- That's inert until the fallback install path (`npm i wrangler@...`) runs inside a directory whose `package.json` has a `workspace:*` dependency. All three do, via `@seichijunrei/contract` — confirmed by grepping each `package.json`. npm has no concept of that pnpm-only protocol → `EUNSUPPORTEDPROTOCOL`.
- Confirmed the fix input exists exactly as you hypothesized: `action.yml` declares `packageManager` (valid values `npm`/`pnpm`/`yarn`/`bun`), and `dist/index.mjs` reads it via `core.getInput("packageManager")`, using it directly instead of falling through to detection when set.

Added `packageManager: pnpm` to the shared "Deploy Worker" step in `_deploy-component.yml` (covers root/catalog/users/web in one place) and all three inline `wrangler-action` invocations in `deploy.yml`'s manual production path (catalog, users, root) — root's own detection already worked correctly (its `workingDirectory` is repo root, which does have the lockfile), but set it explicitly there too rather than leave it relying on directory-based detection.

Confirmed this is unrelated to the earlier 4.79.0→4.112.0 version unification — the failure is about which package manager runs the fallback install, not which wrangler version is requested.

## Failure 1/2 (Atlas migrate) — NOT changed, blocked on DB read access

I did not touch the Atlas flag. Here's what I found from the repo alone, and where I'm stuck without live DB access:

**What the error means**: Atlas found a schema literally named `atlas_schema_revisions` (a dedicated schema, which is what Atlas creates by default when `--revisions-schema` is *not* passed) while this step passes `--revisions-schema public` (which tells Atlas to look for/use a *table* called `atlas_schema_revisions` *inside* `public`). That mismatch means **some earlier invocation applied migrations to this staging database without the `--revisions-schema public` flag**, creating the ledger in Atlas's default location instead.

**What the docs say the intended design is**: `docs/testing-strategy.md:413` and `docs/ops/neon-test-infra.md:104` both explicitly describe the ledger as living in `public.atlas_schema_revisions` — i.e., the CI step's flag matches the documented, agreed design. That's evidence *against* just dropping `--revisions-schema public` to match whatever's already there — doing that would let a possibly-unrelated legacy ledger (created by manual testing, a spike, or an earlier tool run) silently become authoritative.

**Why I stopped here instead of picking a fix**: I can't tell, from the repo alone, whether:
(a) the `public` schema already physically has the catalog/users tables (meaning migrations *did* run for real, just got tracked in the wrong ledger location — safe fix: baseline `public.atlas_schema_revisions` at the current version, no data touched), or
(b) `public` is empty and the stray `atlas_schema_revisions` schema is orphaned noise from some earlier experiment with no real tables behind it (safe fix: drop the stray schema, let this step do a genuinely fresh apply into `public`).

Picking wrong reproduces exactly the risk you flagged: (a)-fix applied to a (b)-world skips migrations that were never really run; (b)-fix applied to an (a)-world either collides on existing objects or double-applies DML.

**I have no way to distinguish these without a live read against the staging Neon database** — I don't have `NEON_DATABASE_URL`, `NEON_API_KEY`, or any local Neon session, and Neon's MCP server needs an OAuth browser flow I can't complete without you or the owner. I started it to see the URL rather than guess blind:

`https://mcp.neon.tech/api/authorize?response_type=code&client_id=hn0cFwOR&code_challenge=jlLfg_6GZ0AcPL_gw6RnGyfIz5Vbo6t465a6e0L1_rE&code_challenge_method=S256&redirect_uri=http%3A%2F%2Flocalhost%3A3118%2Fcallback&state=qfq-SuiNEqTOHuK56ooRjPTmyEc9hVRF91VoruKkOgI&resource=https%3A%2F%2Fmcp.neon.tech%2Fmcp`

If someone completes that in a browser and pastes back the resulting `localhost:3118/callback?code=...&state=...` URL (even though the page itself will fail to load), I can call `mcp__Neon__complete_authentication` with it and query staging directly. Otherwise, these three read-only queries against the staging branch would answer it without me needing access at all:

```sql
-- 1. Where does the real ledger live, and is there also a spurious one in public?
SELECT schema_name FROM information_schema.schemata
WHERE schema_name IN ('public', 'atlas_schema_revisions');

-- 2. What does the existing (wrongly-located) ledger say was applied?
SELECT version, description, applied_at
FROM atlas_schema_revisions.atlas_schema_revisions
ORDER BY version;

-- 3. Do the tables db/migrations/*.sql would create already exist in public?
SELECT table_name FROM information_schema.tables
WHERE table_schema = 'public' ORDER BY table_name;
```

If (3) shows the catalog/users tables already present and roughly matching what query (2)'s version implies, that's case (a) — baseline `public.atlas_schema_revisions`. If (3) comes back empty (or missing most tables), that's case (b) — drop the stray `atlas_schema_revisions` schema and let the step apply fresh.

Your call on how to unblock this — OAuth, or running those three queries and pasting the output back to me. I'll hold off touching the Atlas step until one of those happens.

lifeodyssey:
Pushed 39d9d0d7 — Atlas fix for #516 failure 1/2, done with the real data now in hand.

## Fix: `search_path=public` on the connection URL

Neither of the two directions either of us proposed earlier was right, once the data showed there's nothing in `public` to baseline against (0 business tables on both staging and production) and no real ledger history to fall back to (0 rows). The actual fix is scoping Atlas's own "is this database clean" check to the one schema it manages:

```
case "$NEON_DATABASE_URL" in
  *\?*) scoped_url="${NEON_DATABASE_URL}&search_path=public" ;;
  *) scoped_url="${NEON_DATABASE_URL}?search_path=public" ;;
esac
atlas migrate apply --dir "file://db/migrations" --url "$scoped_url" --revisions-schema public
```

Applied to both real Atlas invocations: `_deploy-component.yml` and `preview.yml` (preview branches clone from the same project, so they can carry `neon_auth` too).

**Verified empirically, not guessed** — downloaded the exact pinned `atlas-community-v0.30.0` binary (not my local canary build, which has different flags) and reproduced the precise failure shape locally:

1. Local Postgres + an extra `neon_auth`-shaped schema (with a table in it) → identical `"connected database is not clean: found schema"` error, no `search_path`.
2. Same setup + `?search_path=public` → apply succeeds; re-running immediately after returns `"No migration files to execute"` (confirms revisions tracking now works forward, not just this one run).
3. `\dt` after the apply → the other schema's table is completely untouched, not dropped or altered.
4. Repeated with **both** `neon_auth` *and* an empty orphaned `atlas_schema_revisions` schema present simultaneously (staging's exact combination) → `search_path=public` tolerates both; no separate `DROP SCHEMA` cleanup step needed at all.

Deliberately did not use `--allow-dirty` (checked `atlas migrate apply --help` on the pinned binary — that's the only other lever available): it suppresses the cleanliness check for the *whole* database, including genuine future dirtiness in `public` itself. Scoping the URL only removes schemas Atlas was never supposed to see from consideration.

## The "provisioning event" note — added to docs/ops/deployment.md

New subsection right after "Schema change policy": states plainly that staging/production currently have zero business tables, that the first post-fix run applies all 11 `db/migrations/*.sql` files from scratch in one shot (not the routine single-file apply every later deploy will actually be), and requires manually confirming all 23 tables exist on staging (via Neon or a direct connection — not inferring success from a green CI job) before letting `deploy-prod` follow. Called out that production is currently *more* empty than staging (no stray ledger schema either), so it's not a smaller catch-up.

Reviewed all 11 migration files for anything assuming a manual step outside the directory (backfills, hand-run grants, seed data) — found none; every `ALTER`/`DROP` is `IF EXISTS`/`IF NOT EXISTS`-guarded, every `DO $$` block is self-contained and idempotent. Noted in the doc that this is a static review, not a substitute for watching the real run.

`actionlint` and `zizmor` clean across the full `.github/workflows/` directory.

Thanks for running those queries — that reshaped the fix entirely (would have gotten it wrong on the "baseline public" path if I'd guessed).

lifeodyssey:
# 最终确认(Fable 席)— PR #492,批次 `8d559f02` + `4a39ec0c` + `39d9d0d7`

## 判定:**APPROVE**(维持)

> 复核插曲,先自曝以免误导后人:我一度比出"三个新提交把整个 Rollback 文档删回旧版"的假 diff——根因是 FETCH_HEAD 被并发 fetch 覆盖,拿错了基。对真 head `39d9d0d7` 重新核验:rollback 表、R2 流程、token 前置全部在位,且我上轮那条 `deployments list` 措辞微瑕也顺手修掉了。无文档回退。

## 重点核验项

### 1. `aws s3 cp` 失败语义 —— 安全,`pulumi up` 不会在无备份下跑

三重保证:① `shell: bash` 下 GH 注入 `-e -o pipefail`;② `cp` 是该 run 块的**末命令**,退出码直接成为 step 结果;③ "Pulumi up" 是**独立的后续 step**,前一步失败即整个 job 中止。所以 R2 写失败 → step 红 → up 不执行 → 不存在"无备份变更 state"的路径。桶名派生(`#s3://` + 剥 `?...`)、`--endpoint-url` 从 `CLOUDFLARE_ACCOUNT_ID` 拼、`AWS_DEFAULT_REGION: auto`、ubuntu-latest 自带 aws CLI——都对。

### 2. 我夸过的双层 `if-no-files-found` 守卫在新形状下——机制没了,但语义完整保留

upload-artifact 步骤整个删除,`skipped` output 随之移除(无消费者,删得对)。实质性守卫都在:`[ ! -s "$backup_file" ]` 对"export 跑了却空产出"仍硬失败(在 cp 之前拦住),cp 失败硬失败;唯一软路径仍只有 `PULUMI_BACKEND_URL` 为空的合法合同分支。无新增 fail-open 面。**成立。**

### 3. 转 R2 本身:对公开仓的正确响应

理由链准确(公开仓 artifact 任何登录用户可下载;`cloudflareAccountId`/`webDomain` 明文 + `catalogDatabaseUrl` 可离线爆破的 ciphertext);同桶同信任边界、复用既有凭证、零新暴露面。"无 lifecycle 过期规则"作为 Known gap 诚实声明并留 follow-up,而不是假装存在——好。文档取件命令(带 run-id 的对象键、从坏 run 自己取)与实现一致。

### 4. siteverify 改 retry-then-warn + stdin

- 我的 P3 已修:`secret@-` 从 stdin 读,不再进 curl argv。✓
- 重试 3 次退避(4s/6s,单次 `--max-time 10`,最坏约 50s)。只有 `invalid-input-secret` 硬失败;重试耗尽 → `::warning::` 放行。另一席的论证我接受:**验证器不可达不构成 secret 有问题的证据**,且该检查同时跑在手动生产热修路径上——为第三方外链的可用性 block 无关热修,比放行更糟。注释里"本地 TOML 解析歧义永不自愈 vs 网络故障压倒性是瞬态"的区分是对的。
- **一条 P3 残留**(不阻塞,建议记进 follow-up):若 Cloudflare 将来改了 error-code 命名/响应形状,每次运行都会静默走 warning 分支——探测永久退化成软通过,只有翻 warning 才会发现。可加缓解:同时匹配响应里的 `"success"` 字段做形状哨兵,或对连续 N 次 warning 出 alert。

### 5. Atlas `search_path=public` —— 不会误伤 `neon_auth`

机制上 `search_path` 把 Atlas 的可见域收进 `public`,`neon_auth` 从此在 Atlas 视野**之外**——不是"跳过检查"而是"从未看见",所以谈不上误伤;拒绝 `--allow-dirty`(会连 `public` 自身的真脏也压掉)的取舍正确。我抽验了迁移文件:`db/migrations/*.sql` 的 DDL **全部无 schema 限定**(grep 无 `public.`/`neon_auth.` 前缀),schema-bound 模式下落进绑定 schema,与修法自洽。URL 拼接对 `?` 有无两种情况都处理了。实现者对 pinned v0.30.0 二进制的本地复现验证 + docs 里"首跑是 provisioning 事件、staging 先行、人工数表后再放 prod"的操作规程,是这批里最扎实的部分。

### 6. `packageManager: pnpm` —— 根因声明属实(独立核验)

我自己在 pinned SHA 的 `dist/index.mjs` 里核对:`PACKAGE_MANAGER: core.getInput("packageManager")`(42125 行)、`getPackageManager(name || detectPackageManager(workingDirectory) || "npm")`(37212 行),而 `detectPackageManager` 只在 `workingDirectory` 内找锁文件——workspace 成员无本地锁文件 → 回落 npm → `npm i` 撞 `workspace:*` 协议死。修法(所有 wrangler-action 步骤显式 `packageManager: pnpm`,含 root 的防御性声明)正确且注释如实。

### 7. 门禁

真 head 检查零失败(Snyk/Sonar/CodeRabbit 绿,余下 lane 为 path-filter 或尚在排队);我本地对 head 三份 workflow 跑了 `actionlint`,clean。

## 结论

**APPROVE。** 三轮 P1/P2 全闭合;coordinator 点名的两处(s3 cp 语义、search_path 与 neon_auth)核验均安全。遗留仅两条已声明/不阻塞的 P3:R2 lifecycle 规则(已记 follow-up)、siteverify 响应形状漂移的永久软通过风险(建议并入同一 follow-up)。

lifeodyssey:
Rebased onto latest `origin/main` (post-#493 merge). `mergeable` is now `MERGEABLE` and CI is queued/running again for the new head (`f61094aa`).

## The rebase

Two real conflicts, both hit the list-type trap this repo has on record — resolved by keeping **both** sides' entries in both cases, not picking one:

1. **`_deploy-component.yml`**: #493's real per-component smoke-check step and this branch's "Push post-deploy secrets to Worker" step both landed right before the old placeholder "Smoke" step. Kept both, in sequence: secrets push, then smoke checks.
2. **`_post-deploy-test.yml`**: #493 added a `secrets:` block (`CLOUDFLARE_API_TOKEN`/`CLOUDFLARE_ACCOUNT_ID`, sourced from secrets not vars) to the `workflow_call` trigger; this branch added a `workflow_dispatch` trigger alongside it. Kept both — and while merging, caught that my own `workflow_dispatch` comment had gone stale (it still said "every suite below is still a TODO no-op," which #493 made false) and fixed the wording to reflect that the suites are real assertions now.

## Self-check (not just "CI is green")

1. All 6 real assertions from #493 present in `_post-deploy-test.yml` (web-landing, healthz, auth-probe, users-probe, catalog-probe, data-plane-probe), plus the 2 environment-gated ones — `resolve-worker-url.sh` and `post-deploy-assert.sh` both still invoked.
2. All 10 `CLOUDFLARE_ACCOUNT_ID` references in `ci.yml` still `secrets.`, not `vars.` — checked with a fresh grep, not memory.
3. This branch's own additions all present: Preflight step, `packageManager: pnpm` (all 4 wrangler-action call sites), `search_path=public` (both Atlas invocations), the R2 `aws s3 cp` backup path, siteverify's stdin+retry probe.
4. `actionlint` + `zizmor` clean across the full `.github/workflows/` directory.

## Both P3s picked up in the same push

- **`CLOUDFLARE_ACCOUNT_ID` empty-guard** in the R2 backup step, matching the existing `PULUMI_BACKEND_URL` guard's style. This was already fail-closed without it (empty value → broken endpoint URL → `aws s3 cp` fails), just with an opaque DNS/connection error instead of a clear one — now it fails with a readable message before ever calling `aws`.
- **siteverify shape sentinel**: tracks whether any retry attempt's response ever contained a `"success"` field (today's known response shape). If none did across all 3 attempts, warns with a message that explicitly calls out "this looks like a response-SHAPE change, not just a network blip" rather than reusing the generic network-failure wording — so a future Cloudflare API change doesn't quietly degrade this whole check into a permanently silent no-op with no signal that it needs attention. Verified both branches locally: the real "always-fails-response" test key hits `shape_seen=true` + `valid`; a deliberately-unreachable host hits `shape_seen=false` (the new branch).

R2 lifecycle "Known gap" note in both `docs/ops/deployment.md` and `infra/AGENTS.md` now points at **#521** instead of a vague "tracked as a follow-up".

Ready for merge whenever CI comes back green.

PR #500, opened 2026-07-29
docs(ops): inventory every repository secret

lifeodyssey:
`ANON_ID_SECRET` was set on 2026-07-29 with nothing anywhere in the repo saying what it does,
what reads it, or what rotating it would break. `deployment.md` documents non-secret
environment variables; the secrets themselves were undocumented.

**Reworked after dual review (Fable + Opus) found 4 P1s** in the first pass — this revision
addresses all four, plus the P2s. Summary of what changed and why, per finding:

## P1-A — it was a snapshot with no way to detect drift

The first version listed 28 secrets by counting `gh secret list` once; within the same
afternoon `LOGFIRE_TOKEN_PROD`/`LOGFIRE_TOKEN_STAGING` were deleted and replaced by two
environment-scoped `LOGFIRE_TOKEN`s, and the doc was already wrong. `gh secret list` can't run
in CI as the enforcement mechanism (`GITHUB_TOKEN` can't list repo secrets; minting a
repo-admin PAT just for this is a net-negative trade).

Added `apps/agent/agent/tests/unit/test_secrets_docs_consistency.py` (same shape as the
existing `test_anonymous_docs_consistency.py`), zero credentials required:
- **A** = every `${{ secrets.X }}` name across `.github/workflows/**`, ∪ every credential-shaped
  name in `worker/containerEnv.ts`'s `CONTAINER_ENV_KEYS` (`_API_KEY`/`_TOKEN`/`_SECRET` suffix
  + `SUPABASE_DB_URL`).
- **B** = every name in `secrets.md`'s two tables.
- `A ⊆ B` (undocumented-but-used secret → red) and every `Live` entry stays in A (a secret that
  goes dead must move to `Referenced by nothing`, not sit in `Live` claiming a consumer that no
  longer exists).

Mutation-verified: adding an undocumented secret to a workflow, renaming a Live entry's only
reference, and wiring up a "dead" entry all flip the relevant test red (verified locally, not
just asserted).

## P1-B — the doc didn't say which layer actually wins

`_deploy-component.yml`'s deploy job always runs `environment: ${{ inputs.environment }}`, so
for every name that exists at both repo and environment scope, the environment-level value
wins for staging/production deploys — the old doc said "rotate `CLOUDFLARE_API_TOKEN` → breaks
all deploys," which is wrong for the repo-level copy of that secret. Added a `Scope` column per
row, and one further finding surfaced by chasing this down: `preview.yml` never sets
`environment:`, so it's the one caller where the repo-level `CLOUDFLARE_API_TOKEN`/
`CLOUDFLARE_ACCOUNT_ID` *isn't* dead weight; every other repo+env duplicate
(`LOGFIRE_TOKEN`, `NEON_DATABASE_URL`, `PULUMI_*`, `R2_*`) has no non-environment-scoped caller
at all, so the repo-level copy is unreachable in any current workflow. Also added the two
consumed-but-previously-undocumented names: `NEON_AUTH_JWKS_URL` and `CORS_ALLOWED_ORIGIN`
(production-only — staging has neither a secret nor a var for it today, a pre-existing gap).

## P1-C — "Consumed by" conflated two different questions, and was also stale

Split into "value lives in" vs "read by." Along the way found the original claim
("`ANON_ID_SECRET`/`TURNSTILE_SECRET` are zero-hit in workflows, set by hand") had *already*
gone stale the same day: `_deploy-component.yml`'s post-deploy-secrets step (from the #492
follow-through) now pushes both via a CI-driven `wrangler secret put`, not a human running it
locally. Documented as Live with the real mechanism, not as a manual/dead entry.

## P1-D — the reference implementation for "adding a secret" wasn't a secret

`ANON_DAILY_COST_BUDGET_USD` is a plain `wrangler.toml` var, never a GitHub secret — following
it as a template would skip the forwarding step that's the whole point of that section. Split
into the three chains this repo actually has (container-secret / Cloudflare-Worker-secret /
plain-var), each with its own reference implementation and numbered steps. `CORS_ALLOWED_ORIGIN`
is called out as the current chain-1→chain-3 migration in progress (#528) — a one-line diff
under this structure instead of a section rewrite.

## P2s folded in

- `NEON_DATABASE_URL`'s row named the wrong consumer (the anon-quota purge job) — that job
  reads `SUPABASE_DB_URL`, and its own inline comment already warns against this exact mix-up
  (#508 review). Corrected to the real consumers: Atlas migrate + web deploy's `DATABASE_URL`.
- `ANON_ID_SECRET` rotation now also names the #514 session-ownership-migration consequence
  (permanently orphans unmigrated anonymous session history, not just rate-limit resets).
- Dropped the hardcoded secret count (DOCS_POLICY rule 4).
- Every row now has a "how do I know it's broken" signal, not just `TURNSTILE_SECRET`.
- `Referenced by nothing` rows now carry a per-row owner action — `GCP_SA_KEY` (real blast
  radius, needs an external GCP check before deletion) is no longer bucketed with
  `CLAUDE_CODE_OAUTH_TOKEN` (safe to delete now) or the two broken chains
  (`ZETA_API_KEY`/`OPENAI_COMPAT_API_KEY` — a keep-or-retire decision, not a delete).
- Title now states its own scope; `deployment.md` links back to this file.

The "Handling" section — rated the strongest part by both reviewers — is unchanged.

No secret values appear in this file.

## Summary by CodeRabbit

* **Documentation**
  * Added operational documentation covering the repository's secrets inventory, usage, rotation impact, and handling requirements.
  * Documented the process for adding new secrets across deployment and workflow configuration.
  * Added the secrets documentation to the canonical documentation index.


lifeodyssey:
Replaced and delivered by #618: https://github.com/lifeodyssey/animichi/pull/618. The replacement was rebased onto `eded5bb9`, corrected the secret scope matrix and retired Mapbox classification, added consistency/path tests, fixed CI-only secret-write guidance, and passed CI, CodeQL, Snyk, Agent Python CI, Neon integration, gitleaks, and focused gates. It merged with head `b7893c011ccee74a78979d01b4ccd5cb28b887d` as main commit `ab255985ea15ae23447c48f7b012ab379aaa3fde`. `make check` remains an environment blocker because uv downloads are denied by the local proxy; no secret values were included.

PR #501, opened 2026-07-29
fix: APP_ENV per-environment + LOGFIRE_TOKEN split by environment (#498)

### LINKED ISSUE #498: APP_ENV is hardcoded to production, so staging is indistinguishable from prod in Logfire
## What I expected

`setup_logfire` already passes the environment through:

```python
logfire.configure(
    service_name=settings.observability_service_name,
    environment=settings.app_env,      # apps/agent/agent/interfaces/routes/_deps.py
    send_to_logfire="if-token-present",
    ...
)
```

So traces should arrive tagged `staging` or `production` and be filterable in the Logfire UI.

## What actually happens

`worker/containerEnv.ts:115` seeds the container environment with a hardcoded value:

```ts
const envVars: Record<string, string> = { APP_ENV: "production", SERVICE_HOST: "0.0.0.0", SERVICE_PORT: "8080" };
```

`APP_ENV` *is* in `CONTAINER_ENV_KEYS`, so the loop below would override it — but **`wrangler.toml` sets `APP_ENV` in no environment block at all** (`[env.production.vars]`, `[env.staging.vars]`, and the default `[vars]` all omit it). There is nothing to override with.

Result: every container reports `APP_ENV=production`. **Staging traces are tagged `production` in Logfire and cannot be told apart from real production traces.**

This is the third touchpoint in the `env`-var chain going unset — the same shape as the pattern recorded after #472/#451, except here the value isn't missing, it's silently *wrong*, which is harder to notice.

## Blast radius

Currently limited to observability. `Settings.is_production` (`settings.py:312`) has no non-test callers today, so nothing else branches on it — but it is a loaded gun: the next feature that gates on `is_production` will silently take the production branch on staging.

One related consequence already in the tree: `settings.py:256` rejects a `*` CORS origin when `app_env == "production"`, so staging is held to production CORS strictness. That is arguably desirable, but it is happening by accident rather than by choice.

## Fix

Add `APP_ENV` to each environment block in `wrangler.toml` (`"production"` / `"staging"`, and whatever the default block should be), and change the `containerEnv.ts` seed from `"production"` to something that fails loudly rather than defaulting to the most privileged environment. Per the env-var chain rule, the change needs all three touchpoints plus the docs table, and a test asserting `APP_ENV` is forwarded with the per-environment value.

## Also worth deciding

There is a single `LOGFIRE_TOKEN` GitHub secret used for every environment (`_deploy-component.yml:121` passes `secrets.LOGFIRE_TOKEN` regardless of the `environment` input), so all environments write into one Logfire project. Once `APP_ENV` is correct, the `environment` tag is enough to separate them — but if you would rather have separate Logfire projects per environment, that is a second decision and needs a second token.

Found while auditing what becomes visible if the repository is made public.

lifeodyssey:
## Summary

Fixes #498. Reviewed twice (dual review round 1, then a corrected round after a wrong diagnosis was retracted) — see "Review history" at the bottom for exactly what changed and why, and what evidence backs each decision.

**A. APP_ENV hardcoded to production**
- `worker/containerEnv.ts` seeded `APP_ENV: "production"` unconditionally, and no `wrangler.toml`
  environment block set `APP_ENV` at all — every deployed container (staging included) reported
  `APP_ENV=production` to Logfire, making staging traces indistinguishable from production.
- **This bug has a second, more severe consequence than the trace-tagging issue in the original
  report: it can crash the staging container at boot.** `settings.py`'s `cors_allowed_origin`
  validator rejects a `"*"` origin when `app_env == "production"`; the `staging` GitHub Environment
  has no `CORS_ALLOWED_ORIGIN` secret (only `production` does), so `cors_allowed_origin` falls back
  to its `"*"` default. Because staging's container ran with `APP_ENV=production` (this bug),
  `Settings()` — which is constructed eagerly at process/module-import time in
  `agent/interfaces/fastapi_service.py` (`app = create_fastapi_app()`), not lazily on first request —
  raises a `ValidationError` and the process never finishes starting. I reproduced this independently
  (not just trusting the reviewer who first found it): instantiating `Settings(app_env="production",
  fallback_agent_model=None)` against both the pre-PR `settings.py` and current code raises
  `ValidationError: cors_allowed_origin must not be '*' in production` in the exact way this bug would
  manifest on staging. I could not confirm from this repo's CI history whether a staging container has
  ever actually completed a real deploy under the current `wrangler.toml`/`containerEnv.ts` (every
  historical `deploy-root-staging` run either predates that job existing or was blocked upstream by
  billing/gate failures) — so I can't say for certain this is an active live incident right now, but
  the crash is real and reproducible, and would hit on the next real staging deploy attempt regardless.
  **Operational follow-up needed independent of this PR merging: a `CORS_ALLOWED_ORIGIN` secret must
  be added to the `staging` GitHub Environment, or staging will fail to boot after (and, per the above,
  arguably already before) this fix — see P1-3 below for why this PR does not paper over that with a
  looser check.**
- `wrangler.toml`: added `APP_ENV` to all three blocks — `development` in the default `[vars]`
  (matches `Settings.app_env`'s own default), `staging` in `[env.staging.vars]`, `production` in
  `[env.production.vars]`.
- `containerEnv.ts`: moved `APP_ENV` into `CONTAINER_REQUIRED_KEYS` (**option ①** from the issue —
  fail-closed, throws if missing) instead of keeping a seed and just changing its value
  (**option ②**). Chosen because this repo was already sent back for the opposite fail-open shape
  once ("can't parse it, so treat it as false" — #441/#443); a missing wrangler.toml var is a
  deploy-config bug and should break the deploy loudly, not fall back to a plausible-looking value.
  `APP_ENV` is also still listed in `CONTAINER_ENV_KEYS` for the standard forwarding-allowlist shape.
- Tests in `worker/containerEnv.test.ts`: `APP_ENV` membership in both key lists, per-environment
  forwarding, fail-closed throw when missing/empty, and a wrangler.toml text-parse check that all
  three blocks set distinct values. `worker/entry.test.ts`'s `requiredEnv()` helper updated.
- `docs/ops/deployment.md`: added `APP_ENV` to the container-runtime "Required" list.
- **A cross-language guard already existed** (`apps/agent/agent/tests/unit/test_deploy_model_env_consistency.py`)
  asserting every `CONTAINER_REQUIRED_KEYS` entry is also a Wrangler-pushed secret in `deploy.yml`.
  `APP_ENV` is a plain `[vars]` value, not a secret `wrangler secret put` provisions, so it's carved
  out as a documented, deliberate exception in that test rather than silently broken.

**B. LOGFIRE_TOKEN split per environment**
- Two dedicated Logfire projects already exist (`animichi-prod`, `animichi-staging`). The mechanism
  separating them is **GitHub Environment-scoped secrets**: a `LOGFIRE_TOKEN` secret defined directly
  on the `production` and `staging` GitHub Environments (same name as the legacy repo-level secret).
  **This required zero workflow YAML changes** — both `_deploy-component.yml`'s `deploy` job
  (`environment: ${{ inputs.environment }}`) and `deploy.yml`'s `deploy` job (`environment:
  production`) already had a job-level `environment:`, and GitHub's documented precedence is that an
  environment secret of the same name is used instead of whatever the caller workflow explicitly
  passes — see "Review history" for the (twice-independently-verified) evidence.
- `CORS_ALLOWED_ORIGIN` has been a `production`-environment-only secret (no repo-level copy) since
  before this issue, so production deploys were already picking it up correctly via the same
  mechanism — it was never silently empty.
- I grepped every workflow for `LOGFIRE_TOKEN`/`logfire` (case-insensitive): only three files
  reference it at all (`_deploy-component.yml`, `ci.yml`, `deploy.yml`), and both jobs that actually
  resolve `secrets.LOGFIRE_TOKEN` in an `env:` block have `environment:` set at the job level — no
  gap where a job would silently fall through to the legacy repo-level secret.
  `agent-eval-nightly.yml` does not reference Logfire/observability config at all, so it isn't an
  instance of this concern.
- **Recommendation on the legacy repo-level `LOGFIRE_TOKEN` (pointing at the deprecated `seichijunrei`
  project):** given the grep above shows it is now provably unreachable from any current real deploy
  path (every job that resolves it has a job-level `environment:` whose own same-named secret wins),
  I'd lean toward deleting it — unlike the other secrets that are legitimately duplicated at both
  repo and environment scope (same live value at both levels), this one points somewhere genuinely
  different (a dead project), so keeping it around only creates a silent-misdirection risk if a
  future job is ever added without an `environment:`. Deferring to the owner's call on this one.
- `docs/ops/deployment.md` documents the mechanism and how it was verified.

**P1-3 (found in review): CORS check needed to cover staging, not just production — corrected framing**
- `cors_allowed_origin`'s validator originally only rejected `"*"` when `app_env == "production"`.
  Before this PR, APP_ENV was hardcoded to `"production"` everywhere, so this check accidentally also
  applied to staging — and, per Part A above, that's not a benign accident: it's what makes staging
  **fail to boot** today (no `CORS_ALLOWED_ORIGIN` secret there). My first pass through this
  framed the risk as "staging silently goes fail-open (permissive wildcard CORS) once APP_ENV is
  fixed" — that's the correct description of what happens if the CORS check is left untouched at
  `== "production"` after Part A's fix. But the actual state of this PR is not "left untouched": I
  widened the check to `!= "development"` specifically so staging keeps failing loudly instead of
  going quiet, which means the **net effect on staging is unchanged from today (still crashes at
  boot) — for the structurally correct reason now (staging genuinely has no CORS origin configured)
  instead of an accidental one (mislabeled APP_ENV)**. The fail-open scenario is a hypothetical I
  ruled out by choosing the stricter check, not a regression this PR introduces.
- Two test fixtures (`conftest.py`'s `mock_settings`, `test_routes_health.py`) relied on the old
  looser default via non-`"production"`/non-`"development"` sentinel `app_env` values (`"test"` /
  `"testing"`) combined with the implicit `"*"` default; both now pass an explicit origin.
- Added `apps/agent/agent/tests/unit/test_cors_origin_validation.py` (no prior test file covered
  this validator at all). Mutation-verified.

**P2 (found in review): Dockerfile hardcoded APP_ENV=production (4th touchpoint)**
- `Dockerfile:29` set `ENV APP_ENV=production` unconditionally. Doesn't affect the real deploy path
  (the Worker always injects it), but anyone who `docker run`s the image directly — bypassing the
  Worker's fail-closed check entirely, e.g. the "Container Path" local-testing flow in
  `docs/ops/deployment.md`, whose example doesn't pass `-e APP_ENV` — would silently get a
  production-tagged container. Removed the hardcoded default; `Settings.app_env`'s own Field default
  (`"development"`) now applies when unset. Added a guard test, mutation-verified.

**P2 (found in review, follow-up issue filed, not fixed here): wrangler.toml naming collision**
- The default (no-`--env`) block and `[env.production]` share the same Worker name (`animichi`) and
  route (`animichi.com/*`) — a bare `wrangler deploy` would silently deploy the production route
  with the default block's config (including `APP_ENV=development`). Every real deploy path in this
  repo already passes `--env production`/`environment: production` explicitly, so this is a latent
  footgun, not an active bug. Filed as #512, not blocking this PR.

**CodeQL (found by CI, not review): incomplete regex escaping in a test helper**
- A high-severity `js/incomplete-sanitization` finding on `worker/containerEnv.test.ts`'s
  `appEnvInBlock` helper: its escaper only handled `.`/`[`/`]` and never escaped a literal backslash.
  Not exploitable in practice (every caller passes a hardcoded literal with no backslash), but a real
  correctness bug on its own terms — demonstrated with the literal `x\bx`, where the un-escaped
  backslash let `\b` get reinterpreted as the word-boundary metacharacter, so the "escaped" regex
  didn't even match its own source literal. Fixed with a complete escape (backslash first, then every
  other metacharacter), extracted into `escapeRegExp` with its own direct test, mutation-verified.

**Flaky-test check: `Python integration (Neon)` failed once, then passed on rerun — corrected**
- Given this repo's own lesson (#479 misdiagnosed as flaky twice), I didn't just accept "passed on
  retry" — I checked whether the failure could plausibly be a real regression from this diff.
  `test_the_probe_never_exceeds_its_timeout_ceiling` failed with `probe took 1.03s` against an
  `assert elapsed < 1.0` bound; the assertion's own docstring explains the `1.0s` is a generous ~20x
  safety margin over the `0.05s` timeout ceiling actually being tested (specifically to absorb
  scheduling jitter while still proving the timeout enforces, rather than the transport's full 2.0s
  sleep completing). Neither this test, its fixture, nor `agent/interfaces/routes/byok.py` (the
  module under test) appear anywhere in this PR's diff — confirmed with `git diff` against this
  branch's base commit. It passed cleanly on an identical-commit rerun.
  **Correction**: I initially wrote "no prior history of this test flaking" — that was inaccurate.
  This test's sibling in the same file (the oversized-body control case) already flaked in this
  exact CI lane during #479 round-3, and the fix at the time was pulling that test off the
  `asyncio.timeout` pipeline. **This is the same pattern recurring a second time**, and the test's
  own "Mock-clock style" docstring is itself misleading — it asserts real `time.monotonic()` wall
  time (`elapsed < 1.0`), which is exactly the "no timing-dependent asserts" pattern this repo's own
  guardrails warn against; a 20x margin doesn't neutralize runner scheduling stalls, it just makes
  them rarer. Still not in this PR's scope (a real fix belongs in `test_byok_probe_containment.py`,
  not here) and not blocking — but the earlier "no history" framing would have been misleading months
  from now, so correcting it explicitly rather than leaving it. A follow-up issue is being filed
  separately to fix the test itself.

## Review history

**Round 1** (Fable + Opus): the first version of Part B added a hand-rolled ternary
(`inputs.environment == 'production' && secrets.LOGFIRE_TOKEN_PROD || ... || secrets.LOGFIRE_TOKEN`)
plus two new repo-level secrets. Sent back:
1. **Fable** ran the exact ternary under `act 0.2.89` and showed a missing dedicated secret silently
   falls through to the legacy shared token — fail-open, and specifically reachable only via that
   missing-secret path (every real caller's `inputs.environment` is only ever `staging`/`production`).
2. **Opus** proposed environment-scoped `LOGFIRE_TOKEN` secrets instead, since the reusable job
   already declares `environment: ${{ inputs.environment }}`. This raised the open question: does a
   caller job with no `environment:` of its own actually get overridden by the reusable job's
   environment secret, or does the caller's explicit pass-through win?
3. I resolved this with evidence: GitHub's docs state *"If you include environment in the reusable
   workflow at the job level, the environment secret will be used, and not the secret passed from the
   caller workflow"* ([Reuse workflows](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows)),
   and I verified it empirically against this repo's real GitHub Actions runners (not `act`, which
   can't reproduce server-side environment-secret precedence) with a throwaway diagnostic
   workflow/branch (deleted after): three same-named `PROBE_SECRET` values of different lengths (repo
   level, `production` env, `staging` env), a caller job with no `environment:` passing the
   repo-level value, a reusable job with `environment: ${{ inputs.environment }}`. Both the staging
   and production jobs resolved their own environment's marker, not the repo-level one the caller
   passed.
4. I reverted `_deploy-component.yml`/`ci.yml`/`deploy.yml` to the original plain
   `${{ secrets.LOGFIRE_TOKEN }}` form and updated docs accordingly.

**Round 2** (a retracted diagnosis, then reconfirmed): a separately-filed issue (#527) initially
asserted the opposite of the round-1 conclusion — that environment secrets in a caller job's
downstream reusable job *don't* resolve, casting doubt on the revert above. That diagnosis was
retracted after a live GitHub Actions run ([30451736886](https://github.com/lifeodyssey/animichi/actions/runs/30451736886))
showed a caller passing an empty placeholder still resolved the real `staging`-environment secret
value in the called job — independently re-confirming round 1's conclusion (both review seats
cross-checked this evidence chain separately). Net effect on this PR: **no further code change** —
the round-1 revert already matches the now-doubly-confirmed correct mechanism. I additionally grepped
every workflow to confirm no Logfire-using job is missing an `environment:` (see Part B above), and
independently reproduced the CORS/APP_ENV staging-boot-crash coupling raised in this round (see Part
A above) rather than taking it on faith.

**Round 3** (#528 landed on `main` mid-review; rebase + a stale invariant): #527/#528 merged to
`main` while this PR was in review, provisioning staging's `CORS_ALLOWED_ORIGIN` via
`wrangler.toml`'s `[env.staging.vars]` (a plain domain value, not a secret) and removing it from
`deploy-root-staging`'s `worker_secrets`/mapped-secrets lists — independently closing the same
staging-boot-crash coupling described in this PR's Part A, at the `wrangler.toml` layer instead of
the `settings.py` layer this PR touches. Rebasing surfaced two things:
1. A real, currently-red invariant on `main` itself (not introduced by this PR): the pre-existing
   `test_ci_root_deploys_match_manual_root_secrets` still asserted staging's Wrangler secret set
   equals the manual `deploy.yml` path's set exactly, which #528 broke without updating this test.
   Confirmed by running the test's own regex logic directly against `origin/main`'s workflow files
   (not just assuming) before fixing it — carved out `CORS_ALLOWED_ORIGIN` as a documented staging
   exception, mutation-verified.
2. `docs/ops/deployment.md`'s CORS note (added by this PR before #528 landed) had gone stale —
   it still described staging as falling through to the container's `"*"` default. Corrected to
   describe the actual current mechanism and explicitly warn against adding a `CORS_ALLOWED_ORIGIN`
   secret to the `staging` GitHub Environment (per review: it would be a dead, unread secret today,
   and a silent var-vs-secret drift risk if a future `worker_secrets` edit ever re-added the name).

The `.github/workflows/` diff against `origin/main` is now empty (`git diff origin/main..HEAD --
.github/workflows/` — verified byte-identical) — Part B's LOGFIRE_TOKEN mechanism needed no code at
all, exactly as round 1/2 concluded.

## Mutation verification (evidence)

- Deleted `APP_ENV = "staging"` from `wrangler.toml` → the "three APP_ENV values are pairwise
  distinct" test failed as expected (restored after).
- Removed `APP_ENV` from `CONTAINER_REQUIRED_KEYS` → **4** tests failed as expected (membership,
  throws-when-missing, throws-when-empty-string, and the "never seeds APP_ENV on its own" guard) —
  restored after.
- Reverted the CORS check to `== "production"` → the new staging test case failed as expected
  (restored after).
- Reintroduced `APP_ENV=production` in the Dockerfile → the new guard test failed as expected
  (restored after).
- Reverted `escapeRegExp` to the old incomplete escaper → the new backslash test failed as expected
  (restored after).
- Live GitHub Actions probe (see Review history) — real empirical confirmation of the secrets
  precedence, not just docs interpretation; independently re-confirmed in round 2 by a separate run.
- Independently reproduced the CORS/APP_ENV staging-boot-crash coupling against both pre-PR and
  current `settings.py` (see Part A) rather than relying on another reviewer's claim.
- `actionlint` and `zizmor` both clean across the full `.github/workflows/` directory.
- `pnpm run test:worker`: 216/216 passing (post-rebase onto main).
- `cd apps/agent && uv run python -m pytest agent/tests/unit/`: 1679/1679 passing, coverage 88.39%
  (floor 87%). `ruff check`/`ruff format --check` clean.
- Reverted the `test_ci_root_deploys_match_manual_root_secrets` staging exception back to strict
  equality → failed against the actual current workflow files as expected (restored after).

## Test plan

- [x] `pnpm run test:worker` (216/216 pass, post-rebase)
- [x] `uv run pytest agent/tests/unit/` (1679/1679 pass, post-rebase)
- [x] `actionlint` on all workflows (clean)
- [x] `zizmor` on all workflows (clean, 0 findings)
- [x] CI green: CodeQL, SonarCloud, Semgrep, TruffleHog, gitleaks, osv-scanner, Agent CI, Agent Eval,
      Python integration (Neon, after confirming the one failure was unrelated flakiness)
- [x] Mutation-tested every new logic branch (see above) — no deploy performed, no real secrets
      written or exposed; the diagnostic probe used throwaway marker strings and was fully cleaned up
- [ ] **Follow-up needed outside this PR**: provision a `CORS_ALLOWED_ORIGIN` secret on the `staging`
      GitHub Environment before/alongside merging, or staging will fail to boot (see Part A)


lifeodyssey:
**独立评审（Opus 席 / 系统影响与治理）** — 核验基准 `origin/fix/498-app-env-and-logfire-env-split` @ `086c5ed3`，非本地 checkout。

核心修复（`wrangler.toml` 三块 × `CONTAINER_REQUIRED_KEYS` fail-closed × 去掉硬编码 seed）方向正确、测试有实质断言（含 wrangler.toml 三值互异的守卫，这个写法我认可）。但**选用的 Logfire 环境隔离机制比仓库自己已有的机制更差**，且引入了两个静默失效面。以下为 REQUEST CHANGES 的依据。

---

## P1

### P1-1 · fallback 是一把「观测系统静默失效」的枪，必须拆掉

```yaml
LOGFIRE_TOKEN: ${{ inputs.environment == 'production' && secrets.LOGFIRE_TOKEN_PROD || (inputs.environment == 'staging' && secrets.LOGFIRE_TOKEN_STAGING) || secrets.LOGFIRE_TOKEN }}
```

GHA 的 `&&`/`||` 返回操作数值、按**真值**短路。空字符串是 falsy。所以真正的触发条件不是注释里写的「其他/未列出的 environment」，而是：

> **`LOGFIRE_TOKEN_PROD` 被删除、被轮换成空值、或某次 secret 写入失败 → 表达式静默降级到 `secrets.LOGFIRE_TOKEN`（旧 project `lifeodyssey/seichijunrei`）。**

后果链条完整闭合成一个无人可见的故障：部署绿 → 容器拿到一个**有效**的 token → Logfire 正常收数 → 数据全部流进已废弃的旧 project → 而人盯着 `animichi-prod` 面板，看到的是「零 trace」。观测系统失效时，恰恰没有任何观测能告诉你它失效了。这比 token 直接为空（`send_to_logfire="if-token-present"`，同样静默，但至少没有错误目的地）更难排查。

而且 GHA 语义上**无法区分「未设置」和「故意置空」**，所以这个降级永远不会报错。

要求：删掉 `|| secrets.LOGFIRE_TOKEN` 这条尾巴，改成显式的空值前置检查（`if [ -z "$LOGFIRE_TOKEN" ]; then echo "::error::..."; exit 1; fi`），让配置错误在部署时炸出来 —— 这正是本 PR 自己在 `CONTAINER_REQUIRED_KEYS` 上采用的 fail-closed 逻辑。同一个 PR 里，容器 env 选择 fail-closed、workflow secret 选择 fail-open 且降级到错误目的地，这个不一致本身就说明 fallback 是想当然加上去的。

配套：旧 `LOGFIRE_TOKEN`（repo secret，2026-04-01 创建）应当**删除或轮换**。它现在是一个指向废弃 project 的活凭据，issue #498 本身就来自「仓库要公开」的审计，留一个无人看的活 token 不符合那次审计的初衷。

### P1-2 · 生产 Logfire token 放在 repo scope，绕过了 `production` 环境闸门

`gh secret list` 实测：

- repo scope 新增：`LOGFIRE_TOKEN_PROD`、`LOGFIRE_TOKEN_STAGING`（今天 09:29 创建）
- **`production` environment 已有 env-scoped secrets**：`CORS_ALLOWED_ORIGIN`、`NEON_DATABASE_URL`、`NEON_AUTH_JWKS_URL`、`PULUMI_*`、`R2_*`、`CLOUDFLARE_*`
- `staging` environment 同样有一套同名 env-scoped secrets

也就是说，**这个仓库早就在用「同名 secret + environment 作用域」来做环境隔离**，`_deploy-component.yml` 也已经写了 `environment: ${{ inputs.environment }}`。正确解法是：在 `production` / `staging` 两个 environment 下各建一个**同名** `LOGFIRE_TOKEN`，然后把 `_deploy-component.yml` 里那行三元表达式改回 `${{ secrets.LOGFIRE_TOKEN }}`，并从 `ci.yml` 的 `secrets:` 映射中**移除** `LOGFIRE_TOKEN` 透传（否则 caller 传入的 repo 级值会与 environment 值产生优先级歧义）。净结果：workflow 改动为零、新 secret 名为零、fallback 分支不存在、P1-1 自动消失。

本 PR 的做法额外付出了两个代价：

1. `LOGFIRE_TOKEN_PROD` 在 repo scope 意味着**任何分支上的任何 workflow 都能读到生产 token**，而 `production` environment 的人工 approve 闸门本来就是为了防这个。相比 `NEON_DATABASE_URL` 用 env scope 保护，生产 Logfire token 反而降级到 repo scope，与仓库既有约定相反。
2. 环境数量与 secret 名数量绑死：将来加 preview/canary 就要再加一个名字和一段三元。

### P1-3 · CORS 通配符守卫在 staging 上被静默摘除 —— issue 明确点过名，PR 没有处理

`settings.py:256` 只在 `app_env == "production"` 时拒绝 `cors_allowed_origin == "*"`。本 PR 把 staging 从 `production` 翻成 `staging` 后，这条守卫对 staging **永久失效**。

这不是理论问题，因为三个事实叠在一起：

- `settings.py:241` `cors_allowed_origin` 的**默认值就是 `"*"`**；
- `gh secret list --env staging` 实测：**staging 环境没有 `CORS_ALLOWED_ORIGIN`**（只有 `production` 有）；
- `ci.yml` 从 repo scope 透传 `CORS_ALLOWED_ORIGIN`，而 repo scope 也没有这个 secret。

所以 staging 上唯一阻止「配置成通配符 CORS 且无人察觉」的东西，恰恰就是这个 PR 要修掉的 bug。修完之后是**从 fail-loud（容器启动即 ValueError）变成 fail-open（静默 `*`）**。这是一次收紧变松绑，而且和 P1-1 同型：都是把响亮的失败换成安静的错误。

issue #498 原文已经把这条写进 blast radius（「arguably desirable, but it is happening by accident rather than by choice」），明确要求做一个**选择**。PR 没有改 `apps/agent/` 任何文件，等于把「by accident」原样换成了另一个「by accident」。

要求二选一并写进 PR body：
- (a) 一行改动：`if v == "*" and str(app_env).lower() != "development"` —— staging 由「意外严格」变成「有意严格」，语义与本 PR 的意图一致；或
- (b) 明确决定 staging 允许 `*`，并同时给 staging 环境补上 `CORS_ALLOWED_ORIGIN`，把决定记进 `docs/ops/deployment.md`。

我倾向 (a)：本 PR 正在把 `is_production` 从「碰巧覆盖 staging」缩窄成「只覆盖 prod」，任何依赖旧宽度的守卫都应当在同一个 PR 里显式改成 `is_development` 的否定形式，否则就是把语义翻转的账留给下一个人。

---

## P2

### P2-1 · `CONTAINER_ENV_KEYS` 全表审计 —— 31 个 key 里 20 个没有任何注入链路

注入路径只有两条：`wrangler.toml` 的 `[vars]` / `[env.*.vars]`，或 `ci.yml`/`deploy.yml` 的 `worker_secrets:` 清单。两条都没有 = 容器永远拿不到，只能吃 `Settings` 默认值。

**有链路（11）**：`DEEPSEEK_API_KEY`、`MIMO_API_KEY`、`SUPABASE_DB_URL`、`GEMINI_API_KEY`、`GOOGLE_MAPS_API_KEY`、`LOGFIRE_TOKEN`、`CORS_ALLOWED_ORIGIN`（以上 worker_secrets）；`CATALOG_API_URL`、`ANON_DAILY_COST_BUDGET_USD`、`ANON_DAILY_MESSAGE_QUOTA`、`APP_ENV`（以上 wrangler vars，`APP_ENV` 由本 PR 补上）

**无任何链路（20）**：

| Key | 落到的默认值 | 备注 |
|---|---|---|
| `MODEL_INPUT_COST_PER_MTOK_USD` | `0.0` | ⚠️ 见下 |
| `MODEL_OUTPUT_COST_PER_MTOK_USD` | `0.0` | ⚠️ 见下 |
| `FALLBACK_AGENT_MODEL` | `None` | ⚠️ 见下 |
| `DEFAULT_AGENT_MODEL` | `openai:mimo-v2.5@...` | 硬编码种子默认值，改模型必须改代码 |
| `ZETA_API_KEY` | `""` | 已知（今日第三例） |
| `OPENAI_COMPAT_API_KEY` | `""` | 已知（今日第三例） |
| `OPENAI_COMPAT_BASE_URL` | `https://api.xiaomimimo.com/v1` | 硬编码 |
| `ANITABI_API_URL` | `https://api.anitabi.cn/bangumi` | 硬编码；docs 列在「Common runtime config」 |
| `LOG_LEVEL` | `INFO` | docs 列了，不可调 |
| `MAX_RETRIES` | `3` | docs 列了，不可调 |
| `TIMEOUT_SECONDS` | `120` | docs 列了，不可调 |
| `OBSERVABILITY_SERVICE_NAME` | `animichi-runtime` | docs 列了，不可调 |
| `OBSERVABILITY_SERVICE_VERSION` | `0.1.0` | docs 列了，不可调；Logfire 里版本永远显示 0.1.0 |
| `DEBUG` | `False` | |
| `CACHE_TTL_SECONDS` | `3600` | |
| `USE_CACHE` | `True` | |
| `RATE_LIMIT_CALLS` | `100` | |
| `RATE_LIMIT_PERIOD_SECONDS` | `60` | |
| `GOOGLE_APPLICATION_CREDENTIALS` | — | |
| `GOOGLE_CLOUD_PROJECT` | — | |

其中三条不只是「不可调」，而是功能实际失效：

- **X4/#274 匿名日成本熔断器在所有已部署环境都是空转的。** `ANON_DAILY_COST_BUDGET_USD = "5.0"` 三个 block 都设了，但计价常数 `MODEL_INPUT/OUTPUT_COST_PER_MTOK_USD` 无注入链路、默认 `0.0`，而 `public_api.py:334-335` 正是用它们算成本。每次请求算出 $0，$5 的天花板永远不会触及。这是一个**看起来已启用、实际从未生效的防御线**，与 `ANON_DAILY_MESSAGE_QUOTA`（有链路，真的生效）形成对照 —— 后者 docs 明说「是公平性机制而非防线」，那么真正的防线现在是空的。
- **`docs/ops/deployment.md` 当前正在指导一个做不到的操作**：「After recharging DeepSeek, set `FALLBACK_AGENT_MODEL=deepseek:deepseek-v4-flash`」。`FALLBACK_AGENT_MODEL` 不在任何 wrangler `[vars]`、不在任何 `worker_secrets`。照文档做不会有任何效果，而且同样不报错。本 PR 恰好在这一段的正上方加了 `APP_ENV` 说明，等于站在这个坑边上没看见。
- `OBSERVABILITY_SERVICE_VERSION` 恒为 `0.1.0`：本 PR 的目标是让 Logfire 数据可分辨，但即便环境分开了，版本维度仍然是死的。

不阻塞本 PR，但请开一个 follow-up issue 收口。建议同时加一个结构性守卫（一个测试即可）：**`CONTAINER_ENV_KEYS` 中的每个 key，必须要么出现在 `wrangler.toml` 三个 `[vars]` 块中，要么出现在 `ci.yml` 的 `worker_secrets` 清单中，否则从清单里删掉它** —— 一个只声明不注入的 allowlist 是纯粹的误导，它让人以为配置项可调。本 PR 已经为 `APP_ENV` 写了「读 wrangler.toml 做断言」的测试，把它泛化成全表覆盖的成本很低，而这正是今天一天内第三次踩同型问题的根因治理。

### P2-2 · `Dockerfile:29` 仍硬编码 `APP_ENV=production` —— 第四个触点

PR 的注释和 docs 都宣称「不再 seed 硬编码默认值」，但镜像层的 `ENV APP_ENV=production` 原封不动。Worker 在启动容器时显式传入的 `envVars` 会覆盖它，所以当前不影响线上；但这意味着**任何绕过 Worker 的容器启动路径（本地 `docker run`、将来别的编排）仍然默认自称 production**，且 `CONTAINER_REQUIRED_KEYS` 的 fail-closed 在那条路径上不会触发（因为镜像里有值）。fail-closed 的价值在于「缺失即炸」，而这一层默认值恰好使它炸不出来。建议改成 `APP_ENV=development` 或直接删除该 ENV，并把 Dockerfile 加进 `feedback_env_var_three_touchpoints` 的清单（实际是四触点）。

### P2-3 · 两条生产部署路径对 fallback 的处理不一致

`deploy.yml`（manual `workflow_dispatch` 生产路径）直接用 `secrets.LOGFIRE_TOKEN_PROD`，**没有** fallback；`_deploy-component.yml`（`ci.yml` 的 staging→prod 路径）有 fallback。同一个生产环境，两条路径在 token 缺失时行为不同：一条静默降级到旧 project，一条静默无 token。按 P1-1 删掉 fallback 后两者自然对齐。

### P2-4 · docs 表述有两处不准

1. 「`worker_secrets: LOGFIRE_TOKEN` in `wrangler.toml`/CI stays unchanged」—— `worker_secrets` 是 `ci.yml` 传给 reusable workflow 的 input，`wrangler.toml` 里没有这个概念，容易误导下一个人去 wrangler.toml 找。
2. 「The original `LOGFIRE_TOKEN` secret remains as the fallback for any other/unlisted environment」—— 如 P1-1 所述，这不是它唯一（也不是最可能）的触发条件。文档描述的触发条件比实际窄，等于把风险写没了。

另：`APP_ENV` 被插入在标题为 **「Required:」** 的 GitHub secret 列表里，条目自己又说 "NOT a GitHub secret"。正文说清楚了，但位置有歧义，建议单独起一个「Per-environment vars（wrangler.toml）」小节，把 `CATALOG_API_URL` / `ANON_*` 一并归位 —— 这样 P2-1 那张表的「哪些走 vars、哪些走 secrets」在文档上也就自然可查了。

---

## P3

- **`is_production` 语义翻转的连带后果（逐个核实结论）**：`is_production` 确认**仍无任何非测试调用方**（全仓库只有 `settings.py:312` 定义处），issue 的判断成立。但 issue 说「只影响 trace 标签」是**不准确的**，`app_env` 另有三个真实消费者：
  - `agents/base.py:62` `_app_client_name()` → 发给模型 provider 的 `X-App-Client` header，staging 从 `animichi Prod` 变成 `animichi Staging`。**这是本 PR 期望的改进**（provider 侧账单/用量也能分环境了），且 `test_model_failover.py:151` 已参数化覆盖三值。无需处理，但值得写进 PR body —— 它是本 PR 的一个未申报的正向收益。
  - `handlers/_helpers.py:23` `rewrite_image_urls()`：只判 `== "development"`，staging 从 production 变 staging 后仍走代理，**线上无变化**。
  - `settings.py:256` CORS —— 见 P1-3。
- **`wrangler dev` 的图片代理路径不再被本地覆盖**：默认 `[vars]` 块现在设 `APP_ENV="development"`，而 `rewrite_image_urls` 的 docstring 假设是「development = 没有 CF Worker」。但 `wrangler dev` **是有** Worker（`/img` 代理在跑）的，所以本地从此拿到 anitabi 原始 CDN URL，`/img` 代理路径在本地永远不被走到，其回归只能到 staging 才暴露。选 `development` 与 `Settings` 默认对齐这个理由本身没错，但请确认这个本地行为变化是有意的；若不是，`[vars]` 用 `"development"` 之外的值（或把 `_helpers` 的判据从 `APP_ENV` 换成一个真正表达「有没有 Worker」的开关）更贴合语义。
- 测试里 `appEnvInBlock` 用 `\n[` 找下一个 section 边界：如果将来某个 `[vars]` 块的**最后一行**是行首以 `[` 开头的注释或数组续行，切片会提前截断。当前文件形态下正确，锚定行首的正则处理得也确实比 `indexOf` 谨慎（注释里也说明了原因，这点写得好），只是边界启发式仍偏脆。
- `_deploy-component.yml` 的三元表达式对 `catalog`/`users`/`web` 组件同样生效，而这些组件根本不消费 `LOGFIRE_TOKEN`。无害，但如果改成 env-scoped 方案（P1-2），这层无关传递也一并消失。

---

## 范围诚实性

对照 issue #498 全文：

- ✅ 「Add `APP_ENV` to each environment block in `wrangler.toml`」—— 三块齐全，值互异，有测试守卫。
- ✅ 「change the seed to something that fails loudly」—— fail-closed 实现正确，且空串也判失败（比只判 `undefined` 更严），mutation guard 也真的守到了「函数不得自行发明值」。
- ✅ 「all three touchpoints plus the docs table, and a test asserting `APP_ENV` is forwarded with the per-environment value」—— 都做了（Dockerfile 第四触点漏了，见 P2-2）。
- ✅ 「Also worth deciding」的 Logfire 双 project —— 做了决定并实施，方向对（owner 已建好两个 project），但机制选错（P1-2）且引入了新的静默面（P1-1）。
- ❌ **「`settings.py:256` … arguably desirable, but it is happening by accident rather than by choice」—— 未处理，PR 未改 `apps/agent/` 任何文件，PR body 也未记录「决定不管」。** 这是本 PR 唯一一处实质性的范围缩小，且恰好落在 issue 明确要求「做选择」的那一条上。

其余部分范围诚实，没有偷工。

**preview 环境（确认无风险）**：`preview.yml` 只部署 `workers/catalog` 和 `apps/web` 两个 Worker（`pnpm --filter catalog/web exec wrangler ... --env preview`），**完全不部署根 Worker**（`worker/entry.ts` + 根 `wrangler.toml`），因此不经过 `buildContainerEnvVars`，`CONTAINER_REQUIRED_KEYS` 的 throw 不会影响 preview 路径。根 `wrangler.toml` 也没有 `[env.preview]` 块，所以也不存在「preview 拿不到 `APP_ENV`」的场景。此外根 Worker 的两条部署路径（`ci.yml` 的 `deploy-root-staging/prod`、`deploy.yml`）都显式传 `environment: staging|production`，不会落到默认 `[vars]` 块 —— 默认块的 `APP_ENV="development"` 只在 `wrangler dev` 生效（其影响见 P3）。

---

## 结论：REQUEST CHANGES

阻塞项三条，都不涉及大改：

1. **P1-1** 删掉 `|| secrets.LOGFIRE_TOKEN` 尾巴，改显式空值前置检查；旧 `LOGFIRE_TOKEN` secret 轮换或删除。
2. **P1-2** 改用 environment-scoped 同名 `LOGFIRE_TOKEN`（`production` / `staging` 两个 environment 各建一个），删掉三元表达式、两个 repo 级新 secret，以及 `ci.yml` 里的 `LOGFIRE_TOKEN*` 透传。这一条同时消化掉 P1-1 和 P2-3，净删代码。
3. **P1-3** 对 CORS 守卫做出显式选择（推荐 `!= "development"` 一行改动），并写进 PR body 与 docs。

P2-1 那张审计表建议单开 issue，但请在本 PR 的 PR body 里引用 —— 熔断器空转和「文档教你设一个设不了的变量」这两条，性质上和 #498 是同一个 bug 的不同实例，值得在同一次修复的记忆里留痕。

技术实现层面（测试质量、fail-closed 的取舍论证、注释里对历史决策的引用）我认为高于本仓库平均水准，尤其 mutation guard 和「为什么不用更安全的非 production 种子值」那段论证写得站得住。问题全部集中在 workflow/secret 治理这一层。

lifeodyssey:
## 独立评审(Fable 席)— REQUEST CHANGES

所有关键声明均已实证复核(实跑测试 + 变异重放 + act 实跑 GHA 表达式 + 逐点 grep),A 部分质量很高,但 B 部分存在一处与本仓库上次被打回同型的 fail-open,须改。

---

### P1(阻断)

**1. `_deploy-component.yml:130` 的 legacy fallback 是 fail-open,已实证。**

用 act 0.2.89 实跑该表达式(secrets 值按长度区分:LEGACY=6 / STAGINGTOKEN=12 / PRODTOKEN=9):

```
staging + LOGFIRE_TOKEN_STAGING 缺失 → len=6  (LEGACY)   ← 静默回退旧共享 project
staging + LOGFIRE_TOKEN_STAGING 存在 → len=12 (STAGING)  ← 正常路径正确
production + LOGFIRE_TOKEN_PROD 缺失 → len=6  (LEGACY)   ← 同样 fail-open
```

GHA 表达式里缺失 secret 求值为空串、空串 falsy、`&&`/`||` 返回操作数——PR 里 Node 模拟的语义假设是对的,但结论走反了:**这个"忠实模拟"恰好证明了 fallback 分支就是静默降级路径**。

两条加重因素:

- **fallback 分支唯一可达的触发条件就是专用 secret 缺失。** 逐个 grep 了全部 caller:`ci.yml` 8 处调用只传 `environment: staging|production`,`deploy.yml` 只有 production——不存在"其他/未列出的 environment"。130 行上方注释"fallback for any other/unlisted `inputs.environment` value"描述的是一条不存在的路径(本仓库刚记过"陈旧注释骗过评审席"的教训)。
- **删掉 fallback 即天然 fail-closed,零成本。** 已核验 pinned SHA `ebbaa15` 的 wrangler-action `dist/index.mjs`:`getSecret()` 对空串/未定义一律 `throw new Error("Value for secret ... not found in environment.")`——表达式落空串时部署会硬失败,而 root 组件的 `worker_secrets` 列表里正有 `LOGFIRE_TOKEN`。`deploy.yml` 改后直接用 `secrets.LOGFIRE_TOKEN_PROD` 无 fallback,已经是这个(正确的)形状;reusable 里却留了软垫,两处不一致。

**要求:** 130 行改为
`${{ inputs.environment == 'production' && secrets.LOGFIRE_TOKEN_PROD || secrets.LOGFIRE_TOKEN_STAGING }}`
(或等价的显式两分支写法),删除 `|| secrets.LOGFIRE_TOKEN`;同时删除 caller 里已成死代码的 legacy 传递(`ci.yml:451`、`ci.yml:556`)与 reusable 的 `LOGFIRE_TOKEN` secret 声明,或至少改掉那条不实注释。GitHub 侧 `LOGFIRE_TOKEN_PROD`/`_STAGING` 两个 secret 已确认存在(gh secret list,2026-07-29T09:29Z 创建),删 fallback 不会破当前部署。

---

### P2(应改/应立案,不单独阻断)

**2. 默认块 `name = "animichi"` + `routes = ["animichi.com/*"]` 与 `[env.production]` 完全同名同路由。** 裸 `wrangler deploy`(不带 `--env`)部署的就是生产 worker 和生产路由,此时带上的是默认块的 `APP_ENV = "development"`——生产流量被标成 development。这正是"野生 animichi worker"的来路。改前的硬编码 "production" 在这条(且仅这条)路径上碰巧是对的;本 PR 把它反转了。现有缓解:`block-local-deploy` hook + CI 恒带 `--env`,所以不阻断本 PR;但请开 follow-up issue 给默认块去别名(去掉 routes、或改名),让裸 deploy 物理上够不到生产。PR 描述称默认块"matches Settings.app_env default"是对的,但没提这个同名同路由的暗雷。

**3. SonarCloud 质量门是真实失败(非 billing):** Security Rating C / Maintainability C on new code(要求 ≥A),是 required check。回修 P1 时顺带看一下 Sonar 的具体 findings。

---

### P3(备忘)

- **CI 12 个失败 check 与本 PR 无关**:全部在 ~5 秒内无 step 直接失败,annotation 明示 "The job was not started because recent account payments have failed or your spending limit needs to be increased"——GitHub Actions billing 问题,需 owner 处理后 re-run 才能拿到真实绿灯。我已在本地独立补跑:`pnpm run test:worker` 214/214 绿、`actionlint` 干净、`zizmor` 0 findings。
- **变异证据属实且保守**:我独立重放了两条——删 `APP_ENV = "staging"` → 2 个测试失败(staging 块断言 + 三值互异);从 `CONTAINER_REQUIRED_KEYS` 移除 `APP_ENV` → **4** 个测试失败(PR 声称 3 个,实际连 mutation-guard 也会红,比声称的更强)。
- **A 部分三触点齐**:wrangler.toml 三块 ✓、`CONTAINER_ENV_KEYS`+`CONTAINER_REQUIRED_KEYS` ✓、`docs/ops/deployment.md` ✓;`entry.ts` 在 main 基线上确认是从 `containerEnv.ts` import(无第二份内联拷贝);`entry.test.ts` 的 `requiredEnv()` 已补 `APP_ENV`;preview.yml 只部署 `workers/catalog`/`apps/web` 自己的 wrangler 配置,不走根 wrangler.toml,无 preview 炸点;`wrangler dev` 走默认 `[vars]`,已含 `APP_ENV=development`。
- wrangler.toml 文本解析测试里的"锚定行首防注释误匹配"处理得很细,fail-closed 的选择及其论证(#441/#443 同型教训)方向完全正确——问题只在 B 部分没把同一原则贯彻到底。

**结论:REQUEST CHANGES**(仅 P1;改法一行为主,回修后可快速复核)。

Issue #537, opened 2026-07-29
P0: root Worker CI deploy is structurally impossible — the build never produces .open-next, and wrangler.toml points one directory too high


lifeodyssey:
`Deploy root staging` failed on `main` (run 30457317465) with:

```
✘ [ERROR] The directory specified by the "assets.directory" field in your
  configuration file does not exist:
  /home/runner/work/animichi/animichi/.open-next/assets
```

**Two independent defects, either of which alone makes root undeployable.** Both were reached for the first time only after the `changes`-job fix (#490) let deploy jobs actually run.

## 1. The build command never produces `.open-next`

`ci.yml` passes `build_filter: frontend`, and `_deploy-component.yml` runs `pnpm --filter frontend build`. In `frontend/package.json`:

```json
"build":  "next build",
"deploy": "opennextjs-cloudflare build && opennextjs-cloudflare deploy"
```

`next build` emits `frontend/.next`. `.open-next` is produced only by `opennextjs-cloudflare build`, which CI never invokes. The artifact wrangler is asked to upload is not built by anything in the pipeline.

## 2. Even if it were built, the path is one level off

`frontend/open-next.config.ts` is `defineCloudflareConfig()` with no options — the default output is `.open-next` **relative to the Next.js app**, i.e. `frontend/.open-next`. The root `wrangler.toml` deploys with `working_directory: "."` and declares, in all three places:

```toml
[assets]                    directory = ".open-next/assets"
[env.production.assets]     directory = ".open-next/assets"
[env.staging.assets]        directory = ".open-next/assets"
```

## Where the wrong path came from

There is a **second wrangler config**, `frontend/wrangler.jsonc`:

```jsonc
{ "name": "animichi-frontend",
  "main": ".open-next/worker.js",
  "assets": { "directory": ".open-next/assets", "binding": "ASSETS" } }
```

Relative to `frontend/`, those paths are correct. This is the config `pnpm --filter frontend deploy` uses, and it deploys a **different Worker** (`animichi-frontend`) from the one `wrangler.toml` describes (`animichi`, `main = worker/entry.ts`, routes on `animichi.com/*`).

So the repo carries two overlapping deploy descriptions for the same frontend, and the root one has the sub-config's relative paths pasted in at the wrong level.

## Consequence

[...]

Issue #484, opened 2026-07-29
P0 ci: the staging→production promotion gate is vacuous (all smoke suites are echo)


lifeodyssey:
From the 2026-07-29 infra/CI audit, before the S1 staging deploy.

`deploy-prod` is gated on `post-staging`, which calls `_post-deploy-test.yml` — and every suite in that file is an `echo`: api (`:41-48`), e2e (`:53-60`), smoke (`:65-72`). The `Smoke` step in `_deploy-component.yml:123-127` is likewise `echo "smoke for $COMPONENT@$ENVIRONMENT — verify in follow-up"`.

**`post-staging` cannot fail.** The comment at `ci.yml:459-460` ("staging validated → production") describes something that does not exist; the real prod gate today is "four `wrangler deploy` commands exited 0" plus a human clicking approve.

The original blocker recorded in the TODOs ("catalog has no public route") is stale: `workers/catalog/wrangler.toml:73-75` has a preview env and `apps/web` now deploys a real user-facing surface.

Scope: replace the echoes with real HTTP assertions against the deployed staging surface — at minimum (a) apps/web landing 200 + expected marker, (b) an authenticated `/v1` probe or a service-binding reachability check for catalog, (c) an anonymous-path assertion once Turnstile secrets are wired (#484). Estimated ~1 day; highest-ROI CI fix in the repo.

Issue #305, opened 2026-07-07
infra: per-PR preview deployments (CF preview URL + Neon branch-per-PR)


lifeodyssey:
## Summary
CI has zero preview deployments today (verified: no `preview` in any `.github/workflows/*`). The 78 rebuild stories all flow through the Executor→Reviewer→Tester loop, and Tester can only validate against a locally assembled stack (Supabase + agent + frontend). A per-PR preview URL would speed up the review/test loop for every story of every iteration.

## Why deferred
2026-07-07 direction triage picked the cross-stack error contract + the eval quality push as the two foundational PRs. Additionally, DB branch-per-PR requires a `NEON_API_KEY` repo secret which does not exist yet (secrets audit 2026-07-07: `CLOUDFLARE_API_TOKEN`/`CLOUDFLARE_ACCOUNT_ID`/`NEON_DATABASE_URL` present, no Neon API key).

## Scope
- **v1 (unblocked today):** on PR open/sync, `wrangler versions upload` a preview of `workers/catalog` (CF native preview URLs, existing CF token), post the URL as a PR comment; fixed shared preview database.
- **v2 (needs `NEON_API_KEY`):** Neon branch per PR, wired into the preview worker's env; auto-delete branch on PR close.
- `apps/web` (TanStack) joins the same mechanism once it exists (S0.2).
- Validate workflow changes on a throwaway test branch first (GHA reusable-workflow errors are compile-time-remote only; lesson recorded 2026-06).

## Spec source
Session 2026-07-07 foundational-direction triage; repo secrets audit same day.

lifeodyssey:
Unblocked 2026-07-07: `NEON_API_KEY` repo secret is now set (API key `gha-pr-preview`, minted via locally-authenticated neonctl OAuth session → Neon API; key value never displayed). v2 scope (Neon branch-per-PR) no longer has a missing-secret dependency.

lifeodyssey:
**v2 implementation up: PR #330** (base `feat/frontend-rebuild`)

What landed:
- Standalone `.github/workflows/preview.yml` — opt-in via the `preview` label (`labeled`/`synchronize`/`reopened`), `closed` always tears down. `ci.yml`/`deploy.yml` untouched.
- Neon branch-per-PR `preview/pr-<N>` via official `create-branch-action` (idempotent reuse), atlas migrations applied to the branch, pooled connection string injected **per Worker version** (`versions secret put` → `versions upload`), never echoed/persisted.
[...]

PR #539, opened 2026-07-29
fix(catalog): stop publishing the Worker on workers.dev

lifeodyssey:
## The exposure

`catalog-staging.<account>.workers.dev` was publicly reachable and answering unauthenticated requests with real data. Verified directly:

| request | response |
|---|---|
| `GET /healthz` | `{"status":"ok","service":"catalog","env":"staging"}` |
| `POST /catalog/resolve` | 200, real bangumi rows |
| `POST /catalog/geocode` | 200, real gazetteer rows |
| `POST /catalog/nearby` | 200, PostGIS query executed |

The 404 shapes distinguish it cleanly: `catalog-staging` returns the *application's* plain-text `404 Not Found` (Worker present, path unrouted), while `catalog` and `catalog-preview` return Cloudflare's `error code: 1042` (no Worker on that subdomain — production has never deployed).

## Root cause

catalog declares no `routes`, and wrangler resolves `defaultWorkersDev = routes.length === 0`. So **every environment that does not say `workers_dev = false` is published publicly by default.** Only `[env.preview]` said it.

`workers/users/wrangler.toml` declares it in all three blocks and is confirmed unreachable (edge `1042` on all three subdomains). Same repo, same pattern, one written out fully and one not.

## Severity — stated honestly

The read surface is **catalog-domain data that is public by design** (anime titles, covers, pilgrimage points, place-name coordinates). catalog's Drizzle schema only declares `bangumi` / `points` / `cluster_version` / `aliases` / `series_edges`; no handler touches user-domain tables. **This is not a data leak, and I am not going to call it one.**

What does matter is **`POST /catalog/ingest`** — an unauthenticated *write* path that triggers outbound Bangumi/Anitabi calls and writes Neon. Per-work singleflight bounds a single work; nothing bounds a scan across `bangumi_id`s. (Existence confirmed by reading the router; not invoked.)

## The deeper issue this PR does *not* fix

catalog's entire security model is the comment at `worker/app.ts:330`:

> `// NOTE: no /catalog/* route — catalog is private (reached via service binding by the container)`

A service binding forwards an ordinary `Request`. Nothing in it distinguishes a binding caller from the open internet — `worker/app.ts:66-68` adds no shared secret, and catalog's handlers make no authorization check. A default value flipped the whole model off while that comment kept confidently stating the old assumption.

Application-layer defence for `ingest` is tracked separately. The structural fix is a **named `WorkerEntrypoint`**, which removes `ingest`'s HTTP surface entirely rather than guarding it with a shared secret — no secret to provision in four places, no new route that can forget the check, and the binding is validated at deploy time.

## Test

`test/wrangler-private.worker.test.ts` asserts no `routes` are declared and that **every** environment sets `workers_dev = false` explicitly, plus a parser self-check so a newly added `[env.X]` cannot be silently skipped.

**Mutation-verified**: removing the line from `[env.staging]`, `[env.production]`, or the top-level block each turns it red. All three checked.

### On test placement

`AGENTS.md` routes filesystem checks to the `*.spike.test.ts` Node pool. This one deliberately stays in the worker pool, and the guide is updated to say why:

- the rule exists because the worker pool's filesystem is sandboxed — this test reads nothing at runtime, since Vite's `?raw` inlines the config at transform time
- following the rule here would be **actively harmful**: the spike pool's `globalSetup` skips the entire suite without `NEON_API_KEY`/`NEON_PROJECT_ID`, so this guard would silently vanish in CI. A skipped guard is not a guard.

## Gates

`typecheck` clean · `lint:oxlint --type-aware --deny-warnings` clean · `test:worker` 262/262 · coverage unchanged (92.69/78.37/95.63/94.69) · zero suppressions.

## Summary by CodeRabbit

* **Security & Privacy**
  * Disabled public Workers.dev and preview URLs for the catalog service across all environments.
  * Retired per-pull-request preview hosting; staging remains the only preview-like exposure.

* **Tests**
  * Added automated checks to verify deployment privacy settings remain disabled and consistently configured across environments.

* **Documentation**
  * Expanded guidance for test pools and configuration safeguards.

### DIFF
--- a/workers/catalog/AGENTS.md
+++ b/workers/catalog/AGENTS.md
@@ -55,9 +55,12 @@ Root guide: `../../AGENTS.md`.
 - `*.worker.test.ts` runs inside workerd via `vitest.config.ts`; its filesystem is sandboxed.
 - `*.spike.test.ts` runs in the Node pool via `vitest.spike.config.ts` for filesystem, TCP, Docker,
   or child-process work. Filesystem parity checks belong here, not in Worker tests — **unless the
-  check must never be skippable**. The spike pool's `globalSetup` skips the entire suite without
-  `NEON_API_KEY`/`NEON_PROJECT_ID`, so a guard placed there silently disappears in a CI job that
-  has no Neon credentials. Config-as-data guards that must always run belong in the worker pool,
-  reading their file via Vite's `?raw` suffix (inlined at transform time, so the sandboxed
-  filesystem never comes into it). See `test/wrangler-private.worker.test.ts`.
+  check must never be skippable**. Two mechanisms make the spike pool skippable: `globalSetup`
+  provides `{ enabled: false }` without `NEON_API_KEY`/`NEON_PROJECT_ID` and `databaseDescribe`
+  turns that into `describe.skip`; and, decisively, the `catalog-spikes` CI job's `if:`
+  (`ci.yml`) restricts it to `workflow_dispatch` and same-repo `pull_request` — **on a push to
+  `main`, or on any fork PR, it does not run at all**. Config-as-data guards that must always run
+  therefore belong in the worker pool, reading their file via Vite's `?raw` suffix (inlined at
+  transform time, so the sandboxed filesystem never comes into it).
+  See `test/wrangler-private.worker.test.ts`.
 - TDD via `vitest-pool-workers`; keep `test/contract-parity.worker.test.ts` green.
--- a/workers/catalog/AGENTS.md
+++ b/workers/catalog/AGENTS.md
@@ -54,5 +54,10 @@ Root guide: `../../AGENTS.md`.
 
 - `*.worker.test.ts` runs inside workerd via `vitest.config.ts`; its filesystem is sandboxed.
 - `*.spike.test.ts` runs in the Node pool via `vitest.spike.config.ts` for filesystem, TCP, Docker,
-  or child-process work. Filesystem parity checks belong here, not in Worker tests.
+  or child-process work. Filesystem parity checks belong here, not in Worker tests — **unless the
+  check must never be skippable**. The spike pool's `globalSetup` skips the entire suite without
+  `NEON_API_KEY`/`NEON_PROJECT_ID`, so a guard placed there silently disappears in a CI job that
+  has no Neon credentials. Config-as-data guards that must always run belong in the worker pool,
+  reading their file via Vite's `?raw` suffix (inlined at transform time, so the sandboxed
+  filesystem never comes into it). See `test/wrangler-private.worker.test.ts`.
 - TDD via `vitest-pool-workers`; keep `test/contract-parity.worker.test.ts` green.
--- a/workers/catalog/test/raw-modules.d.ts
+++ b/workers/catalog/test/raw-modules.d.ts
@@ -1,0 +1,9 @@
+// Vite's `?raw` import suffix inlines a file's contents as a string at
+// transform time. `vite/client` declares this, but adding it to tsconfig's
+// `types` pulls in the whole DOM-flavoured client surface alongside
+// `@cloudflare/workers-types`, which is the wrong ambient environment for a
+// Worker package. Declaring just the one form keeps the type surface honest.
+declare module "*?raw" {
+  const content: string;
+  export default content;
+}
--- a/workers/catalog/test/wrangler-private.worker.test.ts
+++ b/workers/catalog/test/wrangler-private.worker.test.ts
@@ -21,10 +21,13 @@ import toml from "../wrangler.toml?raw";
 //                  public URL per deployed version, with real bindings.
 //
 // Not hypothetical. `catalog-staging.<account>.workers.dev` was answering
-// unauthenticated requests with real data. The Worker's entire security model
-// is the comment at worker/app.ts:330 ("catalog is private, reached via
-// service binding") — nothing in the request tells catalog whether a caller
-// came through the binding or off the open internet.
+// unauthenticated requests with real data, and preview.yml published the
+// per-version URLs as comments on public PRs.
+//
+// Do NOT cite worker/app.ts's "no /catalog/* route — catalog is private"
+// comment as the authority here: seventeen lines below it the same file
+// registers `/catalog/public/anime-overview/:bangumiId`. The reachability
+// story is what this file asserts, not what any comment claims.
 //
 // Parsing here is deliberately line-based rather than via a TOML library: the
 // point is to assert on the file a human edits, and adding a parser dependency
@@ -65,15 +68,15 @@ const environments = (toml: string): Section[] =>
 // policy: a reader deciding whether an environment is private should not have
 // to know wrangler's inheritance rules to answer it.
 //
-// `[env.preview]` is the one deliberate exception — per-PR previews (#305) are
-// reached through version preview URLs, so `preview_urls` is true there on
-// purpose. It is listed explicitly rather than excluded by a wildcard, so a
-// second environment cannot quietly join the exception.
+// There is no exception. `[env.preview]` briefly carried preview_urls = true
+// for per-PR previews (#305); the owner retired those on 2026-07-29 because
+// staging already serves that role, and because preview.yml published the
+// resulting public URLs on public PRs.
 const PRIVACY: Record<string, string[]> = {
   "top-level": ["workers_dev = false", "preview_urls = false"],
   "env.staging": ["workers_dev = false", "preview_urls = false"],
   "env.production": ["workers_dev = false", "preview_urls = false"],
-  "env.preview": ["workers_dev = false", "preview_urls = true"],
+  "env.preview": ["workers_dev = false", "preview_urls = false"],
 };
 
 describe("catalog has no public host", () => {
--- a/workers/catalog/test/wrangler-private.worker.test.ts
+++ b/workers/catalog/test/wrangler-private.worker.test.ts
@@ -5,21 +5,26 @@ import { describe, expect, it } from "vitest";
 // AGENTS.md routes "filesystem parity checks" to the *.spike.test.ts Node pool.
 // That rule exists because the worker pool's filesystem is sandboxed, which
 // does not apply to a transform-time inline; and following it here would be
-// actively harmful — the spike pool's globalSetup skips the whole suite without
-// NEON_API_KEY/NEON_PROJECT_ID, so this guard would silently vanish in CI. A
-// skipped guard is not a guard.
+// actively harmful — the `catalog-spikes` CI job's `if:` restricts it to
+// workflow_dispatch and same-repo pull_request, so on a push to main (and on
+// any fork PR) it does not run at all. A guard that does not run is not a guard.
 import toml from "../wrangler.toml?raw";
 
-// catalog has no `routes` in any environment, and wrangler resolves
-// `defaultWorkersDev = routes.length === 0` — so every environment that does
-// not say `workers_dev = false` is published on `*.workers.dev` BY DEFAULT.
+// catalog must have no public host. Two independent defaults would give it
+// one, and both were live:
 //
-// This is not hypothetical. `[env.preview]` declared it, `[env.staging]` and
-// `[env.production]` did not, and `catalog-staging.<account>.workers.dev` was
-// answering unauthenticated requests with real data. The Worker's entire
-// security model is the comment at worker/app.ts:330 ("catalog is private,
-// reached via service binding") — nothing in the request tells catalog whether
-// a caller came through the binding or off the open internet.
+//   workers_dev  — catalog declares no `routes`, and wrangler resolves
+//                  `defaultWorkersDev = routes.length === 0`, so leaving the
+//                  key out publishes the Worker.
+//   preview_urls — leaving the key out sends no `previews_enabled` at all, so
+//                  the server keeps Cloudflare's enabled-by-default state: a
+//                  public URL per deployed version, with real bindings.
+//
+// Not hypothetical. `catalog-staging.<account>.workers.dev` was answering
+// unauthenticated requests with real data. The Worker's entire security model
+// is the comment at worker/app.ts:330 ("catalog is private, reached via
+// service binding") — nothing in the request tells catalog whether a caller
+// came through the binding or off the open internet.
 //
 // Parsing here is deliberately line-based rather than via a TOML library: the
 // point is to assert on the file a human edits, and adding a parser dependency
@@ -31,22 +36,23 @@ interface Section {
   lines: string[];
 }
 
+// A sub-table ([env.staging.vars], [[env.staging.r2_buckets]]) ends the
+// environment's own key list; keys after it belong to the sub-table.
+const nameOf = (line: string, current: string): string | null => {
+  const header = /^\[(env\.[A-Za-z0-9_-]+)\]$/.exec(line);
+  if (header?.[1]) return header[1];
+  return line.startsWith("[") ? `${current}:subtable` : null;
+};
+
 const sections = (toml: string): Section[] => {
   let current: Section = { name: "top-level", lines: [] };
   const out: Section[] = [current];
-  const open = (name: string): void => {
-    current = { name, lines: [] };
-    out.push(current);
-  };
   for (const raw of toml.split("\n")) {
     const line = raw.trim();
     if (line.startsWith("#")) continue;
-    const header = /^\[(env\.[A-Za-z0-9_-]+)\]$/.exec(line);
-    // A sub-table ([env.staging.vars], [[env.staging.r2_buckets]]) ends the
-    // environment's own key list; keys after it belong to the sub-table.
-    if (header?.[1]) open(header[1]);
-    else if (line.startsWith("[")) open(`${current.name}:subtable`);
-    else current.lines.push(line);
+    const opened = nameOf(line, current.name);
+    if (opened === null) current.lines.push(line);
+    else out.push((current = { name: opened, lines: [] }));
   }
   return out;
 };
@@ -54,28 +60,35 @@ const sections = (toml: string): Section[] => {
 const environments = (toml: string): Section[] =>
   sections(toml).filter((s) => s.name === "top-level" || /^env\.[A-Za-z0-9_-]+$/.test(s.name));
 
-describe("catalog is not published on workers.dev", () => {
+// Both keys are `inheritable` in wrangler (4.112 cli.js:35080 / :35088), so the
+// top-level value alone would suffice. Asserting per environment anyway is the
+// policy: a reader deciding whether an environment is private should not have
+// to know wrangler's inheritance rules to answer it.
+//
+// `[env.preview]` is the one deliberate exception — per-PR previews (#305) are
+// reached through version preview URLs, so `preview_urls` is true there on
+// purpose. It is listed explicitly rather than excluded by a wildcard, so a
+// second environment cannot quietly join the exception.
+const PRIVACY: Record<string, string[]> = {
+  "top-level": ["workers_dev = false", "preview_urls = false"],
+  "env.staging": ["workers_dev = false", "preview_urls = false"],
+  "env.production": ["workers_dev = false", "preview_urls = false"],
+  "env.preview": ["workers_dev = false", "preview_urls = true"],
+};
+
+describe("catalog has no public host", () => {
   it("declares no routes, so workers_dev cannot be left to its default", () => {
     expect(toml).not.toMatch(/^\s*routes\s*=/m);
   });
 
-  it.each(environments(toml).map((s) => s.name))(
-    "%s sets workers_dev = false explicitly",
-    (name) => {
-      const section = environments(toml).find((s) => s.name === name);
-      expect(section?.lines).toContain("workers_dev = false");
-    },
-  );
+  it.each(Object.entries(PRIVACY))("%s declares %s", (name, expected) => {
+    const section = environments(toml).find((s) => s.name === name);
+    expect(section?.lines).toEqual(expect.arrayContaining(expected));
+  });
 
-  it("finds every environment the file actually declares", () => {
-    // Guards the parser itself: if a new [env.X] is added and this list is not
-    // updated, the count assertion fails and the per-environment case above
-    // cannot silently skip it.
-    expect(environments(toml).map((s) => s.name)).toEqual([
-      "top-level",
-      "env.staging",
-      "env.production",
-      "env.preview",
-    ]);
+  it("finds exactly the environments the file declares", () => {
+    // Guards the parser and the table together: a new [env.X] that nobody adds
+    // to PRIVACY fails here rather than being silently skipped by it.each.
+    expect(environments(toml).map((s) => s.name)).toEqual(Object.keys(PRIVACY));
   });
 });
--- a/workers/catalog/test/wrangler-private.worker.test.ts
+++ b/workers/catalog/test/wrangler-private.worker.test.ts
@@ -1,0 +1,81 @@
+import { describe, expect, it } from "vitest";
+// `?raw` inlines the file at transform time, so by the time this runs inside
+// workerd the config is just a string constant — no filesystem is touched.
+//
+// AGENTS.md routes "filesystem parity checks" to the *.spike.test.ts Node pool.
+// That rule exists because the worker pool's filesystem is sandboxed, which
+// does not apply to a transform-time inline; and following it here would be
+// actively harmful — the spike pool's globalSetup skips the whole suite without
+// NEON_API_KEY/NEON_PROJECT_ID, so this guard would silently vanish in CI. A
+// skipped guard is not a guard.
+import toml from "../wrangler.toml?raw";
+
+// catalog has no `routes` in any environment, and wrangler resolves
+// `defaultWorkersDev = routes.length === 0` — so every environment that does
+// not say `workers_dev = false` is published on `*.workers.dev` BY DEFAULT.
+//
+// This is not hypothetical. `[env.preview]` declared it, `[env.staging]` and
+// `[env.production]` did not, and `catalog-staging.<account>.workers.dev` was
+// answering unauthenticated requests with real data. The Worker's entire
+// security model is the comment at worker/app.ts:330 ("catalog is private,
+// reached via service binding") — nothing in the request tells catalog whether
+// a caller came through the binding or off the open internet.
+//
+// Parsing here is deliberately line-based rather than via a TOML library: the
+// point is to assert on the file a human edits, and adding a parser dependency
+// to guard three lines is a worse trade than a strict reader that fails loudly
+// on anything it does not understand.
+
+interface Section {
+  name: string;
+  lines: string[];
+}
+
+const sections = (toml: string): Section[] => {
+  let current: Section = { name: "top-level", lines: [] };
+  const out: Section[] = [current];
+  const open = (name: string): void => {
+    current = { name, lines: [] };
+    out.push(current);
+  };
+  for (const raw of toml.split("\n")) {
+    const line = raw.trim();
+    if (line.startsWith("#")) continue;
+    const header = /^\[(env\.[A-Za-z0-9_-]+)\]$/.exec(line);
+    // A sub-table ([env.staging.vars], [[env.staging.r2_buckets]]) ends the
+    // environment's own key list; keys after it belong to the sub-table.
+    if (header?.[1]) open(header[1]);
+    else if (line.startsWith("[")) open(`${current.name}:subtable`);
+    else current.lines.push(line);
+  }
+  return out;
+};
+
+const environments = (toml: string): Section[] =>
+  sections(toml).filter((s) => s.name === "top-level" || /^env\.[A-Za-z0-9_-]+$/.test(s.name));
+
+describe("catalog is not published on workers.dev", () => {
+  it("declares no routes, so workers_dev cannot be left to its default", () => {
+    expect(toml).not.toMatch(/^\s*routes\s*=/m);
+  });
+
+  it.each(environments(toml).map((s) => s.name))(
+    "%s sets workers_dev = false explicitly",
+    (name) => {
+      const section = environments(toml).find((s) => s.name === name);
+      expect(section?.lines).toContain("workers_dev = false");
+    },
+  );
+
+  it("finds every environment the file actually declares", () => {
+    // Guards the parser itself: if a new [env.X] is added and this list is not
+    // updated, the count assertion fails and the per-environment case above
+    // cannot silently skip it.
+    expect(environments(toml).map((s) => s.name)).toEqual([
+      "top-level",
+      "env.staging",
+      "env.production",
+      "env.preview",
+    ]);
+  });
+});
--- a/workers/catalog/wrangler.toml
+++ b/workers/catalog/wrangler.toml
@@ -20,7 +20,7 @@ compatibility_flags = ["nodejs_compat"]
 # below are a deliberate policy choice enforced by
 # test/wrangler-private.worker.test.ts — a reader should not have to know the
 # inheritance rules to see that an environment is private — NOT a wrangler
-# requirement. `[env.preview]` overrides preview_urls back to true on purpose.
+# requirement. Every environment is private; there is no exception.
 workers_dev = false
 preview_urls = false
 
@@ -91,14 +91,17 @@ ENVIRONMENT = "production"
 binding = "MEDIA_BUCKET"
 bucket_name = "catalog-media"
 
-# Per-PR preview environment (issue #305). Routeless: CI deploys only the
-# one-time bootstrap that creates the Worker; previews are undeployed versions
-# reached through preview URLs. DATABASE_URL is injected per version from the
-# Neon branch preview/pr-<N> by .github/workflows/preview.yml.
+# Per-PR preview environment (issue #305). Version preview URLs are
+# `<alias>-<name>.<subdomain>.workers.dev` and are PUBLIC when enabled — and
+# `.github/workflows/preview.yml` posts them as a comment on a public PR, so
+# enabling them published the address of an unauthenticated catalog carrying a
+# DDL-privileged DATABASE_URL. Owner retired per-PR previews on 2026-07-29:
+# staging already serves that role. This block stays private like every other
+# until the machinery itself is removed (tracked separately).
 [env.preview]
 name = "catalog-preview"
 workers_dev = false
-preview_urls = true
+preview_urls = false
 
 [env.preview.vars]
 ENVIRONMENT = "preview"
--- a/workers/catalog/wrangler.toml
+++ b/workers/catalog/wrangler.toml
@@ -2,6 +2,10 @@ name = "catalog"
 main = "src/index.ts"
 compatibility_date = "2025-09-23"
 compatibility_flags = ["nodejs_compat"]
+# No public workers.dev host — reached only via the root Worker CATALOG service
+# binding. wrangler defaults workers_dev to true when no route is declared, so
+# every environment must say this explicitly (see workers/users/wrangler.toml).
+workers_dev = false
 
 # Hyperdrive binds the Worker to a pooled Postgres connection in production.
 # src/index.ts reads env.HYPERDRIVE.connectionString first, falling back to
@@ -48,6 +52,7 @@ ENVIRONMENT = "development"
 
 [env.staging]
 name = "catalog-staging"
+workers_dev = false
 
 [env.staging.vars]
 ENVIRONMENT = "staging"
@@ -58,6 +63,7 @@ bucket_name = "catalog-media-staging"
 
 [env.production]
 name = "catalog"
+workers_dev = false
 
 [env.production.vars]
 ENVIRONMENT = "production"
--- a/workers/catalog/wrangler.toml
+++ b/workers/catalog/wrangler.toml
@@ -2,10 +2,27 @@ name = "catalog"
 main = "src/index.ts"
 compatibility_date = "2025-09-23"
 compatibility_flags = ["nodejs_compat"]
-# No public workers.dev host — reached only via the root Worker CATALOG service
-# binding. wrangler defaults workers_dev to true when no route is declared, so
-# every environment must say this explicitly (see workers/users/wrangler.toml).
+# No public host of any kind — catalog is reached only through the root
+# Worker's CATALOG service binding. Two separate doors have to be shut:
+#
+#   workers_dev  — the <name>.<subdomain>.workers.dev host. wrangler resolves
+#                  `defaultWorkersDev = routes.length === 0`, and catalog
+#                  declares no routes, so omitting this publishes the Worker.
+#   preview_urls — the per-version <prefix>-<name>.<subdomain>.workers.dev
+#                  host. When the key is absent wrangler sends no
+#                  `previews_enabled` at all, so the server keeps Cloudflare's
+#                  enabled-by-default state — a public URL per version, with
+#                  real bindings. wrangler's "unintentionally public" warning
+#                  does not save us: it early-returns under isNonInteractiveOrCI.
+#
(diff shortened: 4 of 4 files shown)


lifeodyssey:
## Fable 席 — independent review of `339a99a`

**Verdict: REQUEST_CHANGES** — one blocking gap; everything else verified sound.

All claims below were verified independently: gates re-run in a clean worktree at the head SHA, wrangler 4.112.0 source read in `node_modules/wrangler/wrangler-dist/cli.js`, 11 mutations executed against the guard test.

---

### Blocking

**B1 (P1, security) — Preview URLs are still a `*.workers.dev` surface on staging and production.** `workers/catalog/wrangler.toml` now sets `workers_dev = false` everywhere but sets `preview_urls` only in `[env.preview]` (line 82, `true`). In wrangler 4.112's deploy path (`getSubdomainValues`, cli.js:139106-139115) the default for `preview_urls` is `undefined`, and `subdomainDeploy` then POSTs `previews_enabled: undefined` — `JSON.stringify` drops the key, so the server **keeps its existing state**, and Cloudflare's default for Preview URLs is enabled. The preview hostname wrangler itself constructs is `https://<VERSION_PREFIX>-<script>.<subdomain>.workers.dev` — a workers.dev host serving the version with its real bindings (staging DB, R2). wrangler even warns about exactly this state — "To prevent this Worker from being unintentionally public, you may want to … `preview_urls = false`" — but `validateSubdomainMixedState` returns early under `isNonInteractiveOrCI()` (cli.js:139136), so **CI deploys will never show that warning**. Given the owner's stated position is zero workers.dev exposure, the PR title's claim is not yet fully achieved.

Fix is small and lives in the file already being edited:
1. Add `preview_urls = false` at top level. It is `inheritable` (cli.js:35088), so staging/production inherit it, and `[env.preview]`'s explicit `true` still wins for the per-PR preview flow.
2. Extend `test/wrangler-private.worker.test.ts` to guard it the same way.
3. Because `undefined` config never turns the server-side toggle off, the current state of `catalog-staging` must be checked/flipped in the dashboard (or via one explicit deploy after the config change).

**B1-adjacent, owner decision (not a code change here):** `[env.preview]`'s whole design (issue #305) reaches previews *through* workers.dev preview URLs. If the position is truly "no workers.dev at all", that design conflicts and needs an explicit decision; the PR should not silently settle it either way.

### Suggestions (non-blocking)

**S1 (P2, stale-comment risk) — `wrangler.toml:6-8` teaches a wrong wrangler model.** The comment says "every environment must say this explicitly". Verified false: `workers_dev` is an **inheritable** key (cli.js:35080, `inheritable()` at 34229 — env value ?? top-level ?? undefined), so the top-level line alone already covers `[env.staging]`/`[env.production]`. Explicit per-env declaration as a *test-enforced policy* is fine (and I'd keep it), but reword the comment so the next reader doesn't internalize wrong inheritance semantics — this repo has been bitten before by confident wrong comments steering reviewers. Same nuance applies to the PR body's "Only `[env.preview]` said it" framing: a single top-level line would have fixed all envs.

**S2 (P2, docs accuracy) — `AGENTS.md` mechanism claim is imprecise.** "The spike pool's `globalSetup` skips the entire suite without `NEON_API_KEY`" — verified against `test/spike-db-global.ts` (last function): globalSetup does not skip anything; it provides `{ enabled: false, skipMessage }`, and skipping happens per-suite via `databaseDescribe` → `describe.skip` (`test/spike-db.ts:172-177`). A plain `describe` in a spike file would still run without creds. The *stronger and accurate* reason for worker-pool placement is one level up: the `catalog-spikes` CI job (`.github/workflows/ci.yml:325-343`) only runs on same-repo PRs with matching path filters or `workflow_dispatch` — it does not run at all on push-to-main or fork PRs, so a guard there vanishes from most CI runs regardless of globalSetup behavior. The placement decision is correct; fix the stated mechanism in both `AGENTS.md` and the test-file comment.

**S3 (P2, 1-10-50) — `sections()` in `test/wrangler-private.worker.test.ts:33-51` is ~14 lines**, over the repo's ≤10-line function guardrail. Extractable: the per-line classification (comment / env header / other header / key line) into a helper. Indentation depth (2) is within limits.

**S4 (P2, minor accuracy) — PR body says the Drizzle schema "only declares" 5 tables.** True of `src/db/schema.ts`, but `test/spike-db.ts`'s truncate list shows the catalog DB also holds `leg_cache`, `raw_anitabi`, `raw_bangumi`, `media_assets`, `ingest_jobs`, `route_anime` reached via raw SQL. All still catalog-domain — the severity conclusion is unchanged — but "schema declares" understates the surface.

### Verified sound (independent evidence)

- **Root cause**: confirmed in wrangler source — `defaultWorkersDev = routes.length === 0` (cli.js:139107), `workers_dev = config ?? default` (139108). catalog declares no routes; the diagnosis is exact.
- **Mutation testing, extended set — 11/11 killed**: delete `workers_dev = false` from staging / production / top-level; **flip to `true`** at top-level and staging (the flip the PR did not claim — caught, because `toContain("workers_dev = false")` no longer matches); trailing comment `# keep private`; no-space `workers_dev=false`; line commented out; new `[env.qa]` without the key (caught by the parser self-check); spaced header `[ env.staging ]` (env drops out of the list → self-check red); `routes =` added. The hand-rolled parser fails **red** on every form it does not understand — the safe direction; I found no false-green vector. The `[[env.X.r2_buckets]]` sub-table boundary is handled correctly.
- **Severity narrative honest**: `grep` over `src/index.ts`/`src/router.ts` finds zero auth/Authorization/JWT handling; `src/db/schema.ts` declares only catalog-domain tables; `POST /catalog/ingest` (router.ts:75-88 → `ingestWork`) is a real unauthenticated write path. Neither exaggerated nor minimized. The `worker/app.ts:330` comment quote checks out (line ~334 at head).
- **`test/raw-modules.d.ts`**: semantically identical to `vite/client`'s `*?raw` declaration; the DOM-pollution rationale for not pulling `vite/client` into a `@cloudflare/workers-types` package is legitimate; the wildcard only affects `?raw` imports, so it loosens nothing else. Mutations prove the `?raw` inline tracks the real file, not a stale copy.
- **Gates re-run at head SHA**: `pnpm run typecheck` clean · `pnpm run lint:oxlint` (type-aware, deny-warnings) clean · `pnpm run test:worker` green, coverage 92.69/78.37/95.63/94.69 matching the PR's claim · new test file 81 lines (≤200) · zero suppressions in the diff.

B1 is a one-line config change plus a test extension in the exact file under edit — worth landing in this PR rather than as a follow-up, since the PR's stated goal is "not published on workers.dev".

— Fable 席

lifeodyssey:
## Round 2 — Fable seat's blocker was right, and the fix was genuinely incomplete

`e43cecc`. All four findings addressed.

### P1 · `preview_urls` — the side door

This is the one that mattered. Closing `workers_dev` left every *deployed version* reachable at `<prefix>-catalog-staging.<subdomain>.workers.dev`, with real bindings, because an absent key means wrangler sends no `previews_enabled` and the server keeps Cloudflare's enabled-by-default state. And wrangler's own "unintentionally public" warning early-returns under `isNonInteractiveOrCI` — **so CI would never have told us.**

The seat found this by reading `node_modules/wrangler/wrangler-dist/cli.js` rather than by trusting the PR. That is the difference between a review and a rubber stamp.

`preview_urls = false` added at top level plus staging and production; the test now asserts both keys per environment.

### P2 · my comment taught a wrong model

I wrote "every environment must say this explicitly". **False** — `workers_dev` and `preview_urls` are both `inheritable` (`cli.js:35080` / `:35088`), so the top-level value alone covers every environment. Reworded: the per-environment repeats are a *test-enforced policy choice* (a reader should not need wrangler's inheritance rules to see that an environment is private), **not** a wrangler requirement.

Worth naming what this was: a PR whose whole point is "a default value silently did the wrong thing" shipped a comment asserting the wrong rule about that same default.

### P2 · the AGENTS.md mechanism was imprecise, and the true reason is stronger

I verified the seat's correction myself. `globalSetup` provides `{ enabled: false }` (`spike-db-global.ts:302`) and `databaseDescribe` turns that into `describe.skip` — so "globalSetup skips the suite" was wrong in mechanism. The decisive fact is bigger:

```yaml
# catalog-spikes if:
(github.event_name == 'workflow_dispatch' ||
  (github.event_name == 'pull_request' &&
    github.event.pull_request.head.repo.full_name == github.repository))
```

**On a push to `main` — and on any fork PR — that job does not run at all.** Both AGENTS.md and the test comment now say this instead.

### P2 · `sections()` over the 10-line guardrail

Split into `nameOf()` + `sections()`.

### Mutations — five, all killed

| mutation | verdict |
|---|---|
| drop top-level `preview_urls` | RED |
| drop staging `preview_urls` | RED |
| flip staging `preview_urls` to `true` | RED |
| flip production `workers_dev` to `true` | RED |
| add an unregistered `[env.qa]` | RED |

The last one **first appeared to survive**. It had not: my script's `replace(..., count=1)` matched a `[env.preview]` reference inside a *comment* instead of the section header, so nothing was really mutated. Re-run with an assertion that the edit landed on a real header — killed by the environment-list assertion, with `+ "env.qa"` in the diff.

Two false survivors in one day (the seat hit the same thing via BSD `sed` silently ignoring GNU `0,/re/`). The lesson generalises: **assert that the mutation landed where you meant, before reading the test result.** "Test didn't catch it" and "mutation never happened" look identical and mean opposite things.

### Left deliberately

`[env.preview]` keeps `preview_urls = true` — per-PR previews (#305) are *reached through* version preview URLs. It is listed explicitly in the `PRIVACY` table rather than excluded by a wildcard, so a second environment cannot quietly join the exception.

**This does conflict with the owner's decision to use no `workers.dev` host anywhere.** The per-PR preview mechanism needs replacing with `pr-<N>.animichi.com`; that is scoped in #541, not silently changed here.

Gates: `typecheck` clean · `lint:oxlint --type-aware --deny-warnings` clean · `test:worker` 262/262 · coverage unchanged · test file 94 lines · zero suppressions.

lifeodyssey:
# Opus 席 — 独立评审 PR #539 @ `339a99a`

**Verdict: REQUEST_CHANGES** — one blocker, in the same file, four lines below the fix.

The diff itself is correct and I verified every claim it makes. What I am blocking on is that the PR
declares an invariant ("no workers.dev host for catalog") and then writes a guard that permits the one
environment where that invariant is still violated — by a *different* key on a *different* default.

Everything below labelled **[verified]** I ran or resolved myself. **[PR claim]** = repeated from the
body, independently re-checked. **[assessment]** = my judgement, not a fact.

---

## 0. Gates, re-run by me — all reproduce

**[verified]** Fresh worktree at `339a99a`, `pnpm install --frozen-lockfile`, run from `workers/catalog`:

| gate | result |
|---|---|
| `pnpm run typecheck` | exit 0 |
| `pnpm run lint:oxlint` (`--type-aware --deny-warnings`) | exit 0 |
| `pnpm run test:worker` | `PASS (262) FAIL (0)` |
| coverage | 92.69 / 78.37 / 95.63 / 94.69 — matches the body exactly |

Zero suppressions in the diff. `gh pr checks 539` → 19 passed, 0 failed.

### Mutation verification — reproduced, and it is stronger than claimed

**[verified]** I deleted `workers_dev = false` from each of the **four** places it now appears
(`wrangler.toml:8` top-level, `:55` staging, `:66` production, `:81` preview) and ran the guard alone:

```
del-line-8  exit=1   del-line-55 exit=1   del-line-66 exit=1   del-line-81 exit=1   baseline exit=0
```

The body claims three; all four are killed. I also ran two mutations the body does not mention:

- **M5** — append `[env.canary]` with no `workers_dev`: **caught**, and caught twice (the per-env case
  *and* the parser self-check at `:70`). The self-check earns its place.
- **M6** — move staging's `workers_dev = false` down into `[env.staging.vars]`: **caught**. The
  sub-table handling at `:44` is real, not decorative.

The guard is genuinely green for the right reason. Test placement argument (worker pool over the spike
pool, because `globalSetup` would skip it without Neon credentials) is correct and the `?raw` reasoning
holds — `import ... ?raw` is transform-time, the sandboxed FS never participates.

---

## 1. BLOCKER — `preview_urls = true` is the same bug, unguarded, in the same block

`workers/catalog/wrangler.toml:76`

**[verified]** Cloudflare's own docs (Preview URLs, last updated 2026-07-03):

> Both preview URL types follow the format `<VERSION_PREFIX OR ALIAS>-<WORKER_NAME>.<SUBDOMAIN>.workers.dev`
> … If Preview URLs have been enabled, they are **public** and available immediately after version creation.
> … You **cannot** currently configure Preview URLs to run on a subdomain other than workers.dev.
> … Preview URLs are disabled by default when `workers_dev` is disabled.

So `[env.preview]` says `workers_dev = false` — which by itself would have disabled preview URLs — and
then explicitly turns them back on. **Setting `workers_dev = false` on this block buys nothing.**
`catalog-preview` is still published on a workers.dev hostname.

**[verified]** I resolved the configs through wrangler's own parser (`unstable_readConfig`, wrangler
4.112.0), not by reading TOML by eye:

```
catalog:preview     workers_dev=false  preview_urls=true
web:preview         workers_dev=false  preview_urls=true
```

**[verified]** And this is not a dormant setting. `.github/workflows/preview.yml`:

- `:112` `wrangler versions upload --env preview --preview-alias "pr-${PR_NUMBER}"`
- `:111` the version carries the DB connection string for the Neon `preview/pr-<N>` branch
- `:139-140` derives `https://pr-<N>-catalog-preview.<subdomain>.workers.dev`
- `:165-192` **posts that URL as a comment on the pull request** — on a public repo

That is an unauthenticated catalog — including `POST /catalog/ingest` — reachable from the open
internet, with its address *published*, for the life of every `preview`-labelled PR. It is the exact
exposure this PR exists to close, and the new guard asserts `env.preview` as compliant.

**[verified] M7** — I added `preview_urls = true` to `[env.staging]` and ran the guard: **exit 0, not
caught.** The next person to copy the preview block into staging reintroduces a public workers.dev host
for staging and every gate stays green.

**[verified]** Teardown does not clean this up either. `preview.yml:266` deletes only the Neon branch;
`:290` tells the reader *"CF preview versions are routeless and expire from relevance."* They are not
routeless — an aliased preview URL is a route, on workers.dev, and it persists after the PR closes.
That sentence is the same species of confident-but-false in-repo self-description as
`worker/app.ts:330`, which is the one this PR was written to disprove.

**What I need, and it is small.** I am not asking you to break previews (#305 depends on them).

1. `workers/catalog/test/wrangler-private.worker.test.ts:62-68` — extend the per-environment case to
   `preview_urls`: every environment except an explicitly named exception must not enable it. M7 must
   go red.
2. `wrangler.toml:76` — the exception becomes explicit and reasoned in the config comment, not silent.
3. Owner decision, recorded: if workers.dev is genuinely banned, previews need **Cloudflare Access**
   in front of the preview URLs (the documented mitigation — "You can use Cloudflare Access to require
   visitors to authenticate before accessing preview URLs"). That part can be a follow-up issue; the
   guard cannot, because without it the analogue silently recurs.

**[assessment]** If you would rather stop the bleeding on staging/production first, I will not fight an
approve-and-follow-up — but the follow-up that unblocks this is item 1, not #540.

---

## 2. Full-repo scan for the same shape — this is the part worth keeping

**[verified]** Every config resolved through wrangler's parser, not read by eye. `undefined` +
no routes -> `defaultWorkersDev = routes.length === 0` -> **published**
(`node_modules/wrangler/wrangler-dist/cli.js:139107`).

| config | env | name | `workers_dev` | `preview_urls` | routes | verdict |
|---|---|---|---|---|---|---|
| `wrangler.toml` | *(top)* | `animichi` | undefined | undefined | routed | ok |
| `wrangler.toml` | production | `animichi` | undefined | undefined | routed | ok |
| **`wrangler.toml`** | **staging** | **`animichi-staging`** | **`true`** | false | **empty array** | **PUBLISHED** |
| `workers/catalog` | staging | `catalog-staging` | false | undefined | none | fixed by this PR |
| `workers/catalog` | production | `catalog` | false | undefined | none | fixed by this PR |
| `workers/catalog` | preview | `catalog-preview` | false | **`true`** | none | **PUBLISHED** (see 1) |
| `workers/users` | all three | `users*` | false | undefined | none | ok |
| **`apps/web`** | **staging** | **`animichi-web-staging`** | **undefined** | undefined | none | **PUBLISHED** (#541) |
| **`apps/web`** | **production** | **`animichi-web`** | **undefined** | undefined | none | **PUBLISHED** (#541) |
| `apps/web` | preview | `animichi-web-preview` | false | **`true`** | none | **PUBLISHED** (see 1) |
| **`frontend`** | *(only)* | **`animichi-frontend`** | **undefined** | undefined | none | **PUBLISHED if it exists** |

Root `wrangler.toml` **staging** is the new one — not covered by #540 or #541. wrangler resolves it to a
literal `workers_dev = true`, because `[env.staging]` declares an *explicit* empty routes array
(`wrangler.toml:220`). Deployed by `ci.yml:421-424` (`component: root`, `environment: staging`) on every
push to `main`. That Worker is the full SSR app plus `/v1/*`, and `[env.staging.vars]` sets
`ANON_ACCESS_ENABLED = "true"` (`wrangler.toml:224`) — the LLM-spend path. **[assessment]** A
workers.dev hostname is not in your zone, so zone-scoped WAF / rate-limiting rules / bot management do
not apply to it; the only guards left are the in-Worker ones. Related: `worker/turnstile.ts` never
checks the `hostname` field siteverify returns, so a token minted on the real origin verifies fine
against the workers.dev host. Pre-existing, not this PR's — but it is what makes the root-staging
exposure worth an issue rather than a shrug.

`frontend/wrangler.jsonc` has no `workers_dev` in the only block it has, and is in no deploy job I can
find — **[assessment]** so the risk is an *orphan*: if `animichi-frontend` was ever deployed, it is live
on workers.dev serving whatever code it had, and nothing in CI will ever correct it. Worth one
`wrangler deployments list` check.

### The generative cause, which nobody has written down

`infra/index.ts:6-8, 22-23`: *"routes belong to Pulumi, worker code belongs to wrangler."* That decision
is fine — and it **guarantees** zero routes in every wrangler config in this repo, forever. Which means
`workers_dev` defaults to `true` in every config in this repo, forever. The split-brain between Pulumi
routes and wrangler deploys is not a neutral organisational choice; it silently converts wrangler's
default into "public unless you remember." Neither `infra/index.ts` nor `AGENTS.md` says so. Whatever
else happens, that sentence belongs in `infra/index.ts` next to the split-brain rule.

Second-order instance of the same shape, same repo: the root `package.json` `deploy` script runs
`wrangler deploy` with **no `--env`**, so it deploys the *top-level* block to the Worker named
`animichi` — which is production's name (`wrangler.toml:45` and `:148` are the same string) — carrying
`ANON_ACCESS_ENABLED = "true"` (`:62`) where production says `"false"` (`:155`). One default (unnamed
env) plus one incomplete declaration (a top-level block that is a shadow production with different
vars). `block-local-deploy` covers you today; the config does not.

---

## 3. Is the network/application split (this PR vs #540) reasonable?

**[assessment]** The split is right in principle and wrong in its current boundary.

Right in principle: the network layer is the only fix available *today* — no secret provisioning, no
contract change, no coordination with the container, and it is defence-in-depth that survives #540
anyway. Landing it first is correct sequencing, and I would not accept "wait for the RPC entrypoint" as
a reason to leave the hole open.

Wrong at the boundary, for one reason: **#540 is only deferrable if the network layer is complete.** It
is not (see 1). While `preview_urls = true` stands, `POST /catalog/ingest` is reachable *and advertised*
from the internet on every labelled PR, so "application-layer defence is a follow-up" is not a statement
about defence-in-depth — it is the *sole* remaining control, and it does not exist. Either the blocker
closes in this PR and #540 stays a follow-up, or #540 is urgent. Pick one; the PR currently implies the
first while shipping the conditions of the second.

On #540's chosen shape — named `WorkerEntrypoint` RPC instead of a shared secret — I agree, and for the
reason stated: it deletes the HTTP surface rather than guarding it, and the binding is validated at
deploy time so there is nothing to forget. One thing the issue should capture that the PR body does not:
the migration is not "move `ingest` to RPC." Catalog today has **three** distinct callers at three trust
levels — the container via `outboundByHost` (`worker/app.ts:298-303`), an anonymous public GET the root
Worker proxies (`worker/app.ts:347-351`), and nothing else legitimate. Splitting `ingest` off to RPC
while the rest stays `fetch` leaves the read surface on the same undifferentiated `Request`. The
entrypoint should carve the whole boundary, not one method.

### Other places relying on "the network can't reach me"

1. **`worker/app.ts:330`** — the comment the PR quotes as catalog's security model is *already false
   today*, and not only in the way the PR describes. It reads `no /catalog/* route — catalog is
   private`, and **seventeen lines below it** `worker/app.ts:347` registers
   `app.get("/catalog/public/anime-overview/:bangumiId{[0-9]+}")`, proxying straight into the CATALOG
   binding. The PR body and the new test file both cite that comment as authoritative without noticing
   it contradicts the code beneath it. Given the standing lesson about in-repo self-descriptions fooling
   reviewers, quoting a stale comment as the security model *inside the fix for a stale comment* is
   worth a second pass.
2. **`.github/workflows/_deploy-component.yml:127`** — the entire post-deploy smoke step is an `echo` of
   the sentence "catalog has no public route; verify via service binding / wrangler tail in follow-up".
   Every catalog deploy since that line was written has asserted the very proposition that turned out to
   be false, by printing it. This is the single highest-value control missing from the whole picture
   (see 4).
3. `infra/index.ts:44` — "catalog has NO public route". Narrowly true of catalog's own routes; reads as
   a reachability claim, and was relied on as one.
4. `preview.yml:290` — "CF preview versions are routeless" (see 1).

---

## 4. Should the guard be repo-wide? — no new `ci/` package, but do not stop at catalog

**[assessment]** Splitting the answer, because "where does the invariant live" and "where is it
enforced" have different answers here.

**Where it lives:** repo-wide, and it is not "each toml sets `workers_dev = false`." It is *routes are
Pulumi's, therefore wrangler always sees zero, therefore every Worker x every environment must opt out
of both `workers_dev` and `preview_urls` explicitly.* Package-local tests cannot state that; the premise
is architectural.

**Where it is enforced, now:** not in a new cross-package checker. You have a CI/CD rebuild in flight
(`docs/superpowers/specs/2026-07-29-cicd-rebuild-spec.md`) that already proposes a `ci/` package.
Standing one up this week means building against a layout that is about to be replaced, and it would be
the third place the same knowledge lives. Add it to that spec as a stated requirement instead — and note
while you are there that the same spec asserted catalog has no route in any environment, which was
false; the requirement should be phrased as a check, not a claim.

**What to do this week instead — and this is the better control anyway:** replace
`_deploy-component.yml:127` with a real negative assertion. After deploy, `curl` the component's
workers.dev hostname and require Cloudflare's no-Worker error rather than a response from the app — the
`1042` / application-404 distinction the PR body already uses to diagnose this bug is exactly the
oracle. It is an edit to one existing file, it is uniform across all four components for free, and
unlike any static config check it observes **runtime truth**: Cloudflare's docs warn explicitly that
toggling these in the dashboard diverges from the config file until the next wrangler deployment. A
config parser cannot see that; a probe can. Static guard = cheap regression net (keep the catalog one,
extend it per the blocker). Deploy probe = the actual control.

---

## 5. Severity — where I agree, and where the PR undersells itself

**[verified] Where the PR is right, and admirably disciplined:** I read `workers/catalog/src/db/schema.ts`
end to end. It declares `bangumi`, `points`, `cluster_version`, `aliases`, `series_edges` — anime
metadata, pilgrimage coordinates, alias tables. No user-domain table. I read every handler in `src/api/`
and `src/router.ts`; none reaches a user table. I checked every raw-SQL site: all interpolations are
parameterised, and the one `sql.raw` (`src/ingest/raw-store.ts:45`) takes a
`"raw_anitabi" | "raw_bangumi"` union-typed literal, not input. **No injection path, no user data on the
read surface.** Refusing to call it a leak was the right call and I am not going to reverse it.

**[verified] Where the framing is nonetheless too narrow — the credential, exactly as suspected.**
`.github/workflows/ci.yml:368` and `:475`: catalog deploys with `worker_secrets: DATABASE_URL`, sourced
from the `NEON_DATABASE_URL` repository/environment secret. `_deploy-component.yml:76-83` feeds **that
same secret** to `atlas migrate apply --dir file://db/migrations`. A role that applies migrations is a
schema owner. So the connection string sitting inside a publicly-reachable Worker is **DDL-privileged on
the whole data plane** — the same data plane that holds `workers/users`' `routes` table (`user_id`,
`session_id`, saved itineraries; both `.dev.vars.example` files mint their DSN from the same Neon
branch). The handlers do not read it. The **role can**.

So the correction to the framing: "no handler touches user tables" is a statement about *today's code*.
What was actually exposed to the internet is not a set of endpoints, it is an **owner-privileged
database credential with an HTTP front end**. Every future handler, every future raw-SQL edit, every
dependency in the oRPC/Drizzle path inherits that blast radius. That is the difference between "we got
lucky" and "we were fine," and only the second one is worth writing in a PR body.

Mitigating, and I checked before asserting the worst: `NEON_DATABASE_URL` exists as an
**environment-scoped** secret under both the `staging` and `production` GitHub environments (confirmed
via `gh api repos/.../environments/staging/secrets` — names only), so staging plausibly points at its own
Neon branch rather than production's. I cannot read secret values, so I am **not** claiming the exposed
`catalog-staging` was pointed at production data. **[owner action]** Confirm the two values differ, and
confirm the role is not the migration owner — if catalog can run `CREATE`/`DROP`, that is the finding,
independent of everything else in this PR. `db/` should grow a read-scoped role for the serving path.

**[assessment] Second correction, smaller:** the PR judges severity purely in confidentiality terms.
`nearby` / `route` / `geocode` are PostGIS work — `ST_DWithin` scans, and `route` accepts up to 500
`point_ids` (`src/router.ts:60`). Unauthenticated, from the open internet, against metered Neon compute.
That is a **cost and availability** surface even where it is not a confidentiality one, and it is the
same category of harm as `ingest`'s outbound amplification, which the body does take seriously. Worth
one sentence, not a rewrite.

**[verified] Third, small:** `GET /catalog/img/:pointId` (`src/index.ts:73-84`) is a second
unauthenticated *write* path — origin pull from `image.anitabi.cn`, R2 `put`, `media_assets` insert —
that the body's "only `ingest` matters" does not mention. Bounded (once per existing point, tombstoned
on 404), so genuinely minor. But "the only write path is ingest" is not accurate as written.

---

## Findings

| # | P | file:line | issue |
|---|---|---|---|
| 1 | **P0 (blocker)** | `workers/catalog/wrangler.toml:76` + test `:62-68` | `preview_urls = true` publishes `pr-<N>-catalog-preview.<subdomain>.workers.dev`, publicly, with `POST /catalog/ingest` open and the URL posted to a public PR comment (`preview.yml:165-192`). The guard does not cover the key — mutation M7 passes. Fix: guard `preview_urls`, make the exception explicit, decide on Cloudflare Access. |
| 2 | P1 | `wrangler.toml:220` (+ `ci.yml:421-424`) | Root **staging** declares an empty routes array, so wrangler resolves `workers_dev = true`. `animichi-staging` is on workers.dev with `ANON_ACCESS_ENABLED="true"` and no zone-level WAF. Not covered by #540/#541 — needs a new issue. |
| 3 | P1 | `_deploy-component.yml:127` | Post-deploy smoke asserts nothing and prints the false claim. Replace with a negative probe of the workers.dev hostname. This is the control that would have caught the original bug. |
| 4 | P1 | ci/ops | Catalog's `DATABASE_URL` is the same secret handed to `atlas migrate apply`. Verify staging != production, and scope the serving role to read-only. |
| 5 | P2 | `preview.yml:266-290` | Teardown deletes the Neon branch only; CF preview aliases survive, and the comment claims otherwise. |
| 6 | P2 | `worker/app.ts:330` vs `:347` | The quoted security-model comment is already contradicted 17 lines below it. Both the PR body and the new test cite it as authoritative. |
| 7 | P2 | `frontend/wrangler.jsonc` | No `workers_dev`, no routes, no deploy job — orphan risk. Check whether `animichi-frontend` exists in the account. |
| 8 | P2 | root `package.json` `deploy` script | No `--env`, so it targets the production Worker name with the top-level block's anon-access vars. |
| 9 | P2 | `src/index.ts:73-84` | `GET /catalog/img/:pointId` is a second unauthenticated write path the severity section omits. |

Quality ratchet: AC "catalog is not published on workers.dev" — test present, mutation-verified, 4/4
mutants killed. AC "no environment can silently regress" — **partially met**: holds for `workers_dev`,
fails for `preview_urls` (M7).

— Opus 席

lifeodyssey:
## Round 3 — Opus seat's blocker was correct, and the fix stopped four lines short

`b3edec5`.

### The blocker

`[env.preview]` had `preview_urls = true`. Cloudflare version preview URLs are `<alias>-<worker>.<subdomain>.workers.dev`, **public when enabled**, and cannot be moved off `workers.dev` — so `workers_dev = false` on that block bought nothing, and the next line reopened the door.

And the seat found the part that turns this from a config nit into a real exposure: `.github/workflows/preview.yml:149-150` builds those URLs and `:176-206` **upserts them as a comment on the pull request**. This repo is public. So the address of an unauthenticated catalog — one serving `POST /catalog/ingest` with no authorization, and holding the same data-plane credential that `_deploy-component.yml` hands to `atlas migrate apply` — was being published.

My guard asserted that block as compliant. It blessed the hole.

### Fix

**Owner retired per-PR previews** (2026-07-29): staging already serves that role. `[env.preview]` is now private like everything else, the `PRIVACY` table has **no exception left**, and flipping `preview_urls` back to `true` in any of the four environments turns the guard red — staging, production and preview each verified.

Removing the machinery itself (`preview.yml`, both `[env.preview]` blocks, the Workers and Neon branches that already exist on the account, `docs/ops/preview.md`) is **#543**.

### Also: I was citing a false comment as authority

The seat found `worker/app.ts`'s

> `NOTE: no /catalog/* route — catalog is private`

is contradicted **seventeen lines below it**, where the same file registers `/catalog/public/anime-overview/:bangumiId`. Verified. Both the previous version of this test and my PR body quoted that comment as the authority. Removed, with a note telling the next reader not to.

That is the third stale self-description this PR has run into, which is the actual lesson here more than any single config key.

### On the seat's finding about root `wrangler.toml` staging

Verified, and I want to record where I think the framing is wrong. `[env.staging]` does resolve `workers_dev = true`, and `animichi-staging` is on `workers.dev` on every push to `main`. But it is **not** the same defect: it is written explicitly, with a fifteen-line comment explaining that #484's post-staging gate needs a stable reachable URL, and it **already pins `preview_urls = false`**. That is the opposite of "public unless you remember" — someone thought about it and wrote down why.

It does conflict with the owner's newer "no `workers.dev` anywhere" decision. That belongs in **#541**, as a change of policy, not as a bug.

The seat's *generative* observation stands regardless and is the best thing in this review: `infra/index.ts:6-8,22-23` puts routes under Pulumi's ownership, which guarantees `routes.length === 0` in every wrangler config in this repo, permanently — silently converting wrangler's default into **"public unless you remember."** That sentence is written nowhere in the repo. It is the reason this defect appeared three times today in three different packages, and it belongs in the CI rebuild's static checks rather than in three separate per-package guards.

### Runtime probe

Agreed and not done here: `_deploy-component.yml:127` is an `echo` of the same false "catalog is private" claim, and should become a real negative probe of the `workers.dev` hostname. Config parsing cannot see dashboard toggles that diverge until the next deploy. Logged for #541 / the CI rebuild rather than bolted onto this PR.

Gates: `typecheck` clean · `lint:oxlint --type-aware --deny-warnings` clean · `test:worker` 262/262 · coverage unchanged · zero suppressions.

lifeodyssey:
## Opus 席 · 第三轮复核 — **APPROVE**

范围:`339a99a..b3edec5` 增量(含第二轮 Fable 回修)。我的第二轮阻断项已消除,守卫经变异验证为真。两条被驳回的判断我**全部撤回**——驳回成立。下面是自跑证据。

---

### 1. 变异验证:守卫是真的

在独立 worktree(PR HEAD)跑 `test/wrangler-private.worker.test.ts`,基线 6 passed。逐个投毒:

| 变异 | 结果 | 杀死它的断言 |
|---|---|---|
| `wrangler.toml:25` (top-level) → `preview_urls = true` | RED | `top-level declares …` |
| `:73` (`env.staging`) → `true` | RED | `env.staging declares …` |
| `:85` (`env.production`) → `true` | RED | `env.production declares …` |
| `:104` (`env.preview`) → `true` | RED | `env.preview declares …` |

四个变异体全部被杀,**且每个只杀对应的那一条**——其余用例保持绿色,没有过度耦合。杀手统一是 `it.each(Object.entries(PRIVACY))` 那条,测试文件 87–89 行:

```ts
expect(section?.lines).toEqual(expect.arrayContaining(expected));
```

失败形如 `expected [ 'name = "catalog-staging"', …(3) ] to deeply equal ArrayContaining{…}`。

补跑第五个变异(你没要求,但这才是守卫的真实用途)——追加一个**不在 `PRIVACY` 表里、且不带任何隐私键**的 `[env.demo]`:

```
× finds exactly the environments the file declares
AssertionError: expected [ 'top-level', 'env.staging', …(3) ] to deeply equal [ 'top-level', 'env.staging', …(2) ]
```

RED。新环境无法悄悄绕过 `it.each`。闭合性成立。

### 2. 删除 `[env.preview]` 整块(模拟 #543 执行后)

会红,两条:`env.preview declares …` + `finds exactly the environments …`。

**这是期望行为,不阻碍 #543。** 理由:代价是同一个文件里删两行(toml 块 + `PRIVACY` 一个条目),而 #543 的清单第 6 项已经把它写进去了,并明确说 "the assertion will fail loudly if the two get out of step, which is the intent"。配置即数据的守卫本来就该要求成对编辑;若删块能静默通过,才是守卫失效。

一个 P2 小瑕(不阻塞):section 缺失时走 `section?.lines` → `undefined`,报错是 `expected undefined to deeply equal ArrayContaining{…}`。「环境不存在」和「环境缺键」两种失败**信息上不可区分**。加一句 `expect(section, name).toBeDefined()` 会让 #543 的执行者少猜一轮。

### 3. `preview.yml` 会不会发死链?——**不会,但原因和你我先前设想的都不同**

先纠正一个前提:`preview.yml` 是 **label 触发的 opt-in**(`contains(github.event.pull_request.labels.*.name, 'preview')`),不是每个 PR 都跑。

更关键的,我读了 wrangler 4.112 源码:

- `subdomainDeploy`(`cli.js:139172-139195`)是**唯一**会向 `${workerUrl}/subdomain` POST `{ enabled, previews_enabled }` 的地方,只从 `triggersDeploy` 进入 → 即只有 `wrangler deploy` / `wrangler triggers deploy`。
- `versions upload`(`cli.js:144503-144517`)只**读** `${workerUrl}/subdomain`,`if (previews_available_on_subdomain)` 才打印 URL。**它从不写 subdomain 设置。**

而 `preview.yml` 的引导部署被 `if ! wrangler versions list --env preview` 挡着。#543 明说 `catalog-preview` Worker **已存在于账户上**,所以 `versions list` 会成功 → `wrangler deploy --env preview` **永不执行** → 服务端 `previews_enabled` **永远不会被翻成 false**。

推论,两条都要说:

**(a) 死链问题不存在。** 要么服务端仍是 enabled → URL 真实可达(不是死链);要么某次 deploy 落了地 → `versions upload` 不打印 URL → `Derive preview facts` 撞上 `::error::wrangler versions upload printed no preview URL for catalog` 然后 `exit 1`,**在 "Upsert PR comment" 之前就红掉**。响亮失败,不会发死链评论。这个失败模式是对的,本 PR 无需处理,留给 #543 合适。

**(b) 但这意味着本 PR 的 `[env.preview] preview_urls = false` 对那个已provision的 Worker 是纯声明式的。** 因此 **#543 正文开头那句 "The immediate exposure is already closed in #539" 不准确**——对 `catalog-preview` 这个已存在的 Worker,暴露面要到 #543 删掉它(或有人跑一次 `wrangler deploy --env preview`)才真正关闭。#543 自己的清单第 3 项其实已经写对了("they exist on the account; deleting config does not delete them"),只是开头的总结句和它打架。

**这不阻塞本 PR** —— wrangler.toml 里的注释是诚实的("This block stays private ... until the machinery itself is removed")。但建议改一下 #543 的开头句,否则未来读者会以为已经关了。顺带:`apps/web/wrangler.jsonc:19` 仍是 `"preview_urls": true`,web 那半边本 PR 未触及,与 #543 同时收口是对的。

### 4. 关于根 `animichi-staging` —— **我错了,撤回**

读了 `wrangler.toml:228-245`。你说得对,性质相反:

- catalog 的缺陷是**隐式默认**——文件里根本没写,`defaultWorkersDev = routes.length === 0` 把一个「安全模型就是不可达」的 Worker 悄悄发布了。
- 根 staging 的 `workers_dev = true` 是**显式写的**,带 15 行注释绑定 #484 的部署后门禁需要稳定可达 URL,而且 `preview_urls = false` 已经钉在下一行。

一个是「无人知晓的默认」,一个是「有人决定、写下理由、并已收紧相邻旋钮的政策」。归 #541(其 step 6 和排序约束正好覆盖)是对的分类,不是 bug。我上一轮把两者并列,是把缺陷的**类别**弄错了。

### 5. `animichi-staging` 实测 —— **你对,而且比你说的更强**

```
/           HTTP 404  17 bytes  "error code: 1042"
/healthz    HTTP 404  17 bytes
/robots.txt HTTP 200  1248 bytes
```

但我多做了一步对照:

```
https://nonexistent-xyz-9911.zhenjiazhou0127.workers.dev/robots.txt
→ HTTP 200, 1248 bytes  (字节数完全相同)
```

一个**根本不存在**的 Worker 主机名返回同一份 robots.txt。那是 Cloudflare 在 `*.workers.dev` 区域层面托管的 content-signals robots.txt(正文就是那段 EU Directive 2019/790 声明),仓库里 `find -name robots.txt` **零命中**。所以 `/robots.txt` 的 200 对「是否部署了站点」**零信息量**。

我上一轮「从配置推断它服务着更完整的站点」两条腿都是错的:配置推断错(部署从未成功,#537),robots.txt 佐证也错(那不是这个 Worker 发的)。撤回。

### 6. 门禁复跑(`workers/catalog/`,PR HEAD,干净树)

| 门禁 | 结果 |
|---|---|
| `typecheck` (`tsc --noEmit`) | PASS |
| `lint:oxlint` | `check-no-inline-config` PASS;`oxlint --deny-warnings` PASS (exit 0)。`--type-aware` 那条臂在我的沙箱里起不来原生 `tsgolint` 二进制(spawn EAGAIN),**非 lint 失败**;CI 的 `Catalog CI / TS CI` 绿,该 lane 跑的正是 `--type-aware --deny-warnings`。 |
| `test:worker` + coverage | PASS。Stmts 92.69% (1193/1287) · Branch 78.37% · Funcs 95.63% · Lines 94.69% |
| PR 全量 CI | 19 passed / 0 failed |

patch coverage 不适用:增量是 toml + 测试 + AGENTS.md,无生产代码。

### 7. 增量代码质量

`nameOf` 抽取得当(4 行 / 1 层缩进),`sections` 主体 9 行 / 2 层缩进,合 1-10-50。事实性注释这次我核了源:`catalog-spikes` 的 `if:` 确为 `workflow_dispatch || (pull_request && head.repo == repo)`(`ci.yml:331-343`),push 到 main 确实不跑——AGENTS.md 与测试注释的这条主张**属实**。「不要引用 `worker/app.ts` 那句注释」的警告写法我认可,这正是防止下一个评审席被仓库内自述带偏。

P2(不阻塞):`out.push((current = { name: opened, lines: [] }))` 把赋值塞进表达式,读起来要停一拍。拆成两行更直白。

---

### 结论

**APPROVE**。阻断项已解,守卫经五个变异体验证为真守卫而非绿灯装饰,门禁全绿,我两条判断错误已撤回。

合并前唯一建议(**在 #543 里改,不在本 PR**):修正 #543 开头 "The immediate exposure is already closed in #539" 这句——`versions upload` 从不推送 subdomain 设置,已存在的 `catalog-preview` Worker 的 `previews_enabled` 要到删 Worker 时才真正关闭。#543 自己第 3 项清单已经说对了,让开头别跟它打架。

— Opus 席,第三轮复核

Issue #538, opened 2026-07-29
P0: animichi.com has no DNS record at all — the wrangler route can never receive traffic


lifeodyssey:
`animichi.com` does not resolve. Not a 404, not a 5xx — no connection.

```
$ dig +short animichi.com A       → (empty)
$ dig +short animichi.com AAAA    → (empty)
$ dig +short animichi.com CNAME   → (empty)
$ dig +short www.animichi.com     → (empty)

$ dig +short animichi.com NS
addilyn.ns.cloudflare.com.
coen.ns.cloudflare.com.

$ curl https://animichi.com/
curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL
```

The zone is delegated to Cloudflare, but the zone contains no record for the apex or `www`.

## Why the route doesn't save us

`wrangler.toml` declares, for both the default block and `[env.production]`:

```toml
routes = [ { pattern = "animichi.com/*", zone_name = "animichi.com" } ]
```

A Workers route is not a DNS record. It says "if a request arrives for this hostname, run this Worker" — it does not cause requests to arrive. Without a proxied DNS record at the apex, nothing reaches Cloudflare's edge in the first place, so the route never fires. The usual pattern is a proxied dummy record (`AAAA 100::`) whose only purpose is to pull traffic into the edge.

This is why `wrangler deploy` can report success while the domain serves nothing.

## Scope

Nothing in `infra/` creates DNS records — Pulumi manages R2, routes and secrets. `webRoutesEnabled` exists in `Pulumi.staging.yaml` only as a **commented-out instruction** (`#   pulumi config set webRoutesEnabled true --stack staging`) and is set in neither stack.

## Correcting an earlier assumption

Today's handoff recorded the topology problem as "`apps/web` is on `*.workers.dev` with no `/v1` proxy; the only `/v1` provider is the root Worker on `animichi.com`". That framing assumed `animichi.com` serves something. It has never served anything. The site is not, and has never been, live at its own domain.

## Blocked on the owner

Whether the apex should point at the root Worker (`animichi`, the `/v1` + OpenNext path, currently undeployable — #537) or at `apps/web` (deployed and healthy today at `animichi-web-staging.<account>.workers.dev`) is a topology decision, not a DNS chore. Filing the fact; not choosing.

Related: #537, #512, #325, #252.

Issue #506, opened 2026-07-29
P1: web deploys never inject VITE_* build vars — magic-link login reports not_configured in production


lifeodyssey:
From the unresolved `qodo-code-review` audit (2026-07-29). Flagged **Action required** on #463 and again on #483; verified against `origin/main` and still present. The two reports are the same CI gap seen from two sides, so they are merged here.

## The gap

`grep VITE_TURNSTILE_SITE_KEY` and `grep VITE_NEON_AUTH_BASE_URL` return **zero hits across every file in `.github/workflows/`**. The build step in `_deploy-component.yml` injects only the legacy `NEXT_PUBLIC_*` variables from the frozen Next.js package. Vite inlines `import.meta.env.VITE_*` at build time, so a variable that is absent during the build is absent from the shipped bundle — there is no runtime fallback.

## Two symptoms, both live

**1. Magic-link login is broken in production.** `apps/web/src/lib/auth/neonAuth.ts:baseUrl()` returns `undefined` when `VITE_NEON_AUTH_BASE_URL` is missing, and `sendMagicLink()` then returns `"not_configured"`. Every deployed web client — staging *and* production — cannot send a login link.

**2. Anonymous chat on staging 403s with no way through.** `TurnstileGate.tsx:69` reads `env.VITE_TURNSTILE_SITE_KEY`; empty means the widget never renders. `wrangler.toml` sets `ANON_ACCESS_ENABLED = "true"` at the top-level `[vars]` (line 62) — production overrides it to `"false"` (line 155), staging does not. So on staging every anonymous turn is challenged by an edge gate whose widget cannot be drawn.

Neither is a future risk. Both are the current behaviour of what is deployed.

## Fix

Add both variables to the web build step in `_deploy-component.yml`, thread them through `ci.yml` and `deploy.yml`, and record them in `docs/ops/secrets.md`. `VITE_TURNSTILE_SITE_KEY` is a **site key** — public by design, so a repository variable rather than a secret is the honest home for it, the same call already made for `NEXT_PUBLIC_MAPBOX_TOKEN`.

Related: the four-touchpoint rule in `docs/ops/secrets.md`; this is the build-time analogue of the container-env chain that also bit #472, #451 and #498.

Severity: **P1**.

Issue #541, opened 2026-07-29
infra: provision DNS + routes for animichi.com / www / staging.animichi.com, then turn workers.dev off everywhere


lifeodyssey:
Implements the domain topology decided in #252. This is the infra half; the SEO half lives there.

## Today

`animichi.com` has **no DNS record** (#538), so the `routes = [{ pattern = "animichi.com/*" }]` in `wrangler.toml` can never fire — a Workers route says "run this Worker if a request arrives for this hostname", it does not cause requests to arrive. Everything reachable today is on `*.workers.dev`, which the owner does not want used at all.

## Target

| Hostname | Worker | Protection |
|---|---|---|
| `animichi.com/v1/*`, `/img/*`, `/healthz` | root (`animichi`) | public |
| `animichi.com/*` | `animichi-web` | public |
| `www.animichi.com/*` | — | 301 → apex |
| `staging.animichi.com/*` | `animichi-web-staging` (+ root staging for `/v1`) | Cloudflare Access |
| `pr-<N>.animichi.com/*` | `animichi-web-preview` | Cloudflare Access |

The scaffolding already exists and is switched off: `infra/index.ts:19-73` has a `webRoutesEnabled` Pulumi flag (default `false`) that creates the `WorkersRoute` resources, and `infra/Pulumi.staging.yaml:6-8` carries `webDomain` / `cloudflareZoneId` as **commented-out instructions**. Neither stack sets any of them. Nothing in `infra/` manages DNS records at all — that gap is new work.

## Ordering is load-bearing

**Do not turn `workers_dev` off before the replacement hostname answers.**

`apps/web` staging's *only* current access path is `animichi-web-staging.<account>.workers.dev`. Turning it off first makes staging unreachable and takes CI's post-deploy smoke assertions red with it. Sequence:

### ⚠️ Corrected 2026-07-29 — the original order became harmful once `frontend/` was retired

This issue first said "DNS first, narrow the route later." That was right while the root Worker still fell through to `frontend/`'s Next.js pages. **It is now wrong.** `feat/537-retire-legacy-frontend` removes that fallthrough, and root `wrangler.toml` **still claims `routes = ["animichi.com/*"]`** — so with the fallthrough gone, **the moment a DNS record exists the apex answers a JSON 404**, because the root Worker is still the thing claiming the hostname and it no longer has pages to serve.

Narrowing the route now comes **before** DNS, not after:

[...]

PR #542, opened 2026-07-29
feat(web): X-Robots-Tag noindex on all non-production apps/web responses

lifeodyssey:
## Why

`animichi.com` has no DNS (#538); the only live origin is `animichi-web-staging.…workers.dev`, served with no robots signal at all. If crawlers index it, staging becomes the canonical site before the real domain exists. This is the code layer of the owner's defense-in-depth decision (the other layer is account-level Cloudflare Access).

## Mount point (verified against installed sources, not memory)

- **Nitro runtime plugin on the `beforeResponse` hook.** Verified in the installed `nitropack@2.13.4` (resolved from `@tanstack/nitro-v2-vite-plugin@1.155.0`'s `^2.13.1`): `dist/runtime/internal/app.mjs:81-84` wires h3 `onBeforeResponse` to `hooks.callHook("beforeResponse", event, response)` for **every** response; hook signature is `(event: H3Event, response: { body?: unknown })` (h3 `AppOptions["onBeforeResponse"]`, h3 1.15.11).
- **Registration**: `nitroV2Plugin({ plugins: ["./src/server/noindex-plugin.ts"] })` — the wrapper's only argument is a `NitroConfig` spread verbatim into `createNitro` (`dist/esm/index.js:56-79`), and Nitro resolves `plugins` paths against `srcDir` = `apps/web` (`dist/core/index.mjs:468` → `resolveNitroPath`).
- The plugin default export is typed `NitroAppPlugin` directly instead of calling `defineNitroPlugin` — that helper is the identity function in 2.13.4, and skipping it keeps nitropack's runtime barrel out of the vitest unit pool.
- Note: requests served straight off the `ASSETS` binding (public files) bypass the Worker and don't get the header; all HTML (the thing crawlers canonicalize) is Worker-rendered and covered.

## Where APP_ENV comes from

`wrangler.jsonc` `vars` per env block (`staging`/`production`/`preview`), read at runtime from `event.context.cloudflare.env` (the cloudflare-module preset passes `{ cloudflare: { env } }` via `_platform`, spread into the event context in `app.mjs:50-56`). Not `VITE_*` — those were never actually injected (#506). `vars` is non-inheritable, hence the per-block declarations and the config guard.

**Decision rule is `APP_ENV !== "production"` → noindex** (fail-safe): missing, empty, or unknown values get the header; only an explicit `production` suppresses it — the inverse failure direction of #498.

## AC → test → mutation kill

Tests: `apps/web/tests/unit/noindex-header.test.ts` (real h3 app + `hookable`, wired exactly like Nitro's own `onBeforeResponse`; zero mocks) and `apps/web/tests/unit/wrangler-app-env.test.ts` (`?raw` + jsonc-parser + zod, always-run unit pool — the PR #539 pattern).

| Mutation | Killed by |
|---|---|
| M1 `=== "production"` → `!==` (inverted) | 9 assertions, incl. **AC2** "does NOT add the header when APP_ENV is production" |
| M2 `readAppEnv` hardcoded `"production"` (silent never-send, the #498 analog) | 7 assertions: both AC1, all four AC3, AC4-404 |
| M3 `undefined` treated as production (fail-safe broken) | exactly the 3 AC3 tests (missing APP_ENV / absent cloudflare / cloudflare without env) |
| M4 header only when `event.path === "/"` | exactly AC4 "404 page gets the header too" (survived the first harness — `app.use(base)` strips the path — fixed by `createRouter` in commit 2, then killed) |
| M5 value `"noindex"` instead of `"noindex, nofollow"` | 7 exact-value assertions (AC1) |
| M6 registered on `afterResponse` instead of `beforeResponse` | 7 header-present assertions |
| M7a prod `vars` emptied | zod schema + all 4 AC5 assertions (loud) |
| M7b staging `vars` block deleted | exactly "env.staging declares APP_ENV matching its env name" |
| M7c fourth env block added | exactly "finds every environment the file actually declares" (iterator guard) |

## End-to-end verification (beyond unit tests)

- `pnpm run build` → plugin present in emitted `.output/server/chunks/nitro/nitro.mjs`.
- `wrangler dev --env staging` (local dev, no deploy): `/` → 200 + `x-robots-tag: noindex, nofollow`; `/no-such-page` → 404 + header.
- `wrangler dev --env production`: header absent on both.

## Gates (all green)

- `pnpm run typecheck` — clean
- `pnpm run lint:oxlint` — clean (type-aware, --deny-warnings)
- `pnpm test` — 203 files / 1523 tests, coverage 98.53/95.44/98.6/99.46 vs floors 98/95/98/99

## Out of scope (per card)

- No `workers_dev: false` for apps/web (staging's only access path today).
- No robots.txt / sitemap / OG changes.

Closes the code-layer half of #538's interim protection.


lifeodyssey (commented):
# Review — Opus 席 (architecture & systemic impact)

**Verdict: APPROVE.** No blocking findings. The mount point is correct (I re-derived it from installed sources, not from the PR body), the fail-safe direction is right, and I reproduced all three gates independently with results matching the PR body exactly. Findings below are follow-ups, one of which must be tracked before the #541 apex cutover.

My brief was architecture/systemic impact, so I deliberately did not duplicate the other seat's line-level review.

---

## 5. Evidence grading (doing this first, because everything else depends on it)

**Re-verified by me, line by line, from installed sources:**
- `nitropack@2.13.4` `dist/runtime/internal/app.mjs:81-84` — `onBeforeResponse` → `hooks.callHook("beforeResponse", event, response)`.
- `dist/presets/cloudflare/runtime/_module-handler.mjs:92-104` — puts `cloudflare.env` inside `_platform`; `app.mjs:50-54` spreads `_platform` into `event.context`. Chain confirmed end to end.

**True, and stronger than the PR claims:** h3 1.15.11 `dist/index.mjs:2332` — the *error* path also fires the hook (`if (app.options.onBeforeResponse && !event._onBeforeResponseCalled)`). So thrown-error 500s (reachable today per #536's JSON-only-`Accept` 500) carry the header too. Not claimed in the PR; it holds.

**Gates — re-run by me, not taken on trust.** Fresh worktree at `8f3bfc64`, `origin/main` confirmed an ancestor:
- `pnpm run typecheck` → exit 0
- `pnpm run lint:oxlint` → exit 0
- `pnpm test` → exit 0, **203 files / 1523 tests**, coverage **98.53 / 95.44 / 98.6 / 99.46**

Identical to the PR body. `src/server/noindex-plugin.ts` does not appear in the uncovered-lines table.

**Inferred from code, not executed by me:** the ASSETS-bypass gap (§6) — I read the config and Cloudflare's assets-first default; I did not build and curl a static path.

**Taken from the PR, unverified by me:** the `wrangler dev --env staging|production` runs and the `.output/server/chunks/nitro/nitro.mjs` build check. Worth noting these are *load-bearing, not decorative*: the unit test's 404 is a **handler-returned** 404, whereas Nitro's own unmatched-route 404 is a different path. Only the `wrangler dev` run covers that one.

**Unverifiable before merge, by construction:** real deployment — `animichi.com` has no DNS (#538). `wrangler dev` runs the same workerd, so the delta is small but not zero: a real deploy additionally exercises `--env` resolution of `vars` through the Cloudflare API, and asset-vs-Worker routing at the *edge* rather than in the local emulator. Neither is exercised anywhere in CI today.

**Required post-merge verification:**
1. After the next `main` push deploys staging — curl the staging origin, assert the header is **present**.
2. At #541 step 3 — assert **present** on `staging.animichi.com` and **absent** on `animichi.com`.
3. Make (2) permanent by wiring the absent-assertion into post-deploy smoke (see P1-B).

---

## 1. The `APP_ENV` injection chain — "三触点" audit

**Two of three touchpoints need nothing; I verified rather than assumed:**
- `CONTAINER_ENV_KEYS` — **not needed.** That list is root-container-only; `apps/web` has no container.
- `_deploy-component.yml` — **not needed.** `vars` ship inside the config file. I traced `environment:` → wrangler-action → `--env`, and confirmed both `apps/web` deploy jobs pass it explicitly (`ci.yml:379-394` staging, `ci.yml:486-500` production). `deploy.yml` does not deploy `apps/web` at all.

**P2-A — the doc touchpoint is missed.** `docs/ops/deployment.md:100` (on `main`) documents `APP_ENV` as *the root Worker's* container var, from `wrangler.toml [vars]`, values `development`/`staging`/`production`. This PR adds a **second, unrelated** `APP_ENV` and does not touch that doc.

**Are the two semantically consistent?** Yes on meaning — both answer "which deployment environment am I". But two things diverge and the doc must say so:
- Value sets differ: root `development`/`staging`/`production`; web `staging`/`production`/`preview`.
- **Failure contracts are inverted.** Root is fail-*closed* — `containerEnv.ts` `CONTAINER_REQUIRED_KEYS` throws when it's missing (#498). Web is fail-*open-to-noindex*. Same name, opposite behaviour on absence. That is worse undocumented than two different names would be.

---

## 4. Fail-safe direction — correct, and one concrete misfire path

The direction is **right**: an unlabelled deployment should assume it is not production. The inverse (#498's hardcoded `"production"`) is the failure that actually bites.

**P1-A — `apps/web/wrangler.jsonc:3` `"name": "animichi-web"` is identical to `:25` `env.production.name`, and the top level declares no `vars`.** So `wrangler deploy` from `apps/web` **without `--env`** publishes to the *production* Worker with `APP_ENV` undefined → the fail-safe fires → **production silently serves `noindex, nofollow`**.

Before this PR a bare deploy was merely redundant. This PR makes it harmful. Not reachable today (CI always passes `--env`; local deploy is hook-blocked), so not a blocker — but it is precisely the "one default silently did the wrong thing" shape that has cost this repo three times. Cheap containment, no production code: extend `tests/unit/wrangler-app-env.test.ts` to assert the top-level `name` matches no env block's `name`.

**P1-B — the asymmetry that matters: there is no detector for the expensive direction.** Every mutation in the PR's table is killed at unit level, but all of them concern *failing to send* the header. The failure that costs money is *sending it in production*, and nothing anywhere asserts absence — `_post-deploy-test.yml` is entirely TODO no-ops. Fold `! grep -i x-robots-tag` into the cicd-spec's AC-3.9 `curl https://animichi.com/` assertion.

---

## 2. Right abstraction level? Yes — and now is the wrong time to abstract

Agreed with keeping this in `apps/web`:
- catalog and users have no public hostname and serve JSON only. Crawlers cannot reach them.
- The cicd-rebuild spec already owns the cross-service config invariant via the new `wrangler-config-contract` static check (§1053, AC-1.10, AC-2.2). A second parallel shared-policy layer built now would collide with it.

**But the PR's premise is incomplete, and this is my most substantive finding.** The PR (quoting #252) says the only live origin is `animichi-web-staging…workers.dev`. I checked `main`:

- `wrangler.toml` `[env.staging]` (lines 218-247) sets `workers_dev = true` **explicitly and deliberately** (#484).
- `[env.staging.assets]` (line 306) serves `.open-next/assets`, built by `pnpm --filter frontend build` (`ci.yml:426-437`).
- `git grep -i x-robots-tag origin/main` → **zero hits.**

So `animichi-staging.<account>.workers.dev` is a **second live, fully indexable HTML origin with no robots signal whatsoever** — and it serves the older, more *complete* `frontend/` site, which makes it arguably the more exposed of the two. This PR closes one of two doors. That deserves its own issue; it is not scope creep for this one.

---

## 3. Domain topology (#252 / #541) — consistent

**(a) Still holds after the staging move.** The scheme keys off a wrangler `vars` value, not a hostname, so `staging.animichi.com` changes nothing — `--env staging` still ships `APP_ENV=staging`. Strictly better than a hostname sniff, which would have broken at the cutover. Good call.

**(b) Deferring `workers_dev: false` is correct and required by #541's own ordering** (step 6, "**Only then**"). Touching it now makes staging unreachable and takes CI's smoke red, exactly as #541 warns. It is not lost either: `wrangler-config-contract` (AC-2.2) will force an explicit `workers_dev` on every routeless env. Right trade-off, correctly deferred.

---

## 6. ASSETS bypass — confirmed; low today, rising in Phase 3b

Confirmed: `wrangler.jsonc:12-15` declares `assets.directory`/`binding` with no `run_worker_first`, so Cloudflare serves matching static paths ahead of the Worker.

**Today's severity is genuinely low.** `apps/web/public/` contains fonts, `.webp` images and `sw.js` — **no HTML**. No indexable *document* bypasses the header, and images on a noindexed page are not normally indexed on their own.

**The real issue is a coupling to a known upcoming change.** The cicd-rebuild spec (§1307, §1319, AC-3.9) puts `robots.txt` and `sitemap.xml` into `apps/web/public/` in Phase 3b. Those will bypass the Worker, so a staging deploy would serve the **production** `robots.txt`/`sitemap.xml` verbatim with zero environment awareness — and the noindex header cannot reach them to compensate. Phase 3b should generate them per-environment or serve them from a route rather than `public/`. Please write this into #252 now, while the constraint is visible.

---

## Other

**P2-B — dependency classification.** `h3`, `hookable` and `nitropack` all landed in **devDependencies**, but `src/server/noindex-plugin.ts` is *production runtime source* importing `h3`. It works because the nitro build bundles it and CI does a full `pnpm install --frozen-lockfile`; a production-only install would break the build. `h3` belongs in `dependencies`. (`hookable` and `nitropack` are genuinely test-only — correct where they are.)

**Note, not a finding.** `src/features/anime/head.ts:64` already emits `<meta name="robots" content="noindex">` for non-indexable pages. Meta and header combine most-restrictive-wins, so there is no conflict; on non-prod both fire, harmlessly.

---

### Summary

| # | Item | Severity | Blocking |
|---|---|---|---|
| P1-A | `wrangler.jsonc:3` top-level name == `:25` production name; bare deploy deindexes prod | P1 | No — add config guard |
| P1-B | No assertion that production *lacks* the header | P1 | No — required at #541 step 3 |
| P2-A | `docs/ops/deployment.md:100` not updated; two `APP_ENV`s with inverted failure contracts | P2 | No |
| P2-B | `h3` is a devDependency but imported by runtime source | P2 | No |
| P2-C | ASSETS bypass — low now, matters at Phase 3b `robots.txt`/`sitemap.xml` | P2 | No |
| — | Second indexable origin: root staging Worker has no robots signal at all | P1 | No — separate issue |

Ship it.

— Opus 席

lifeodyssey (commented):
## Independent review — Fable 席

Verdict: **APPROVE**. No blocking findings. All claims below re-verified locally from a clean worktree of `8f3bfc64` (not taken from the PR description).

### 1. The dangerous direction: production accidentally noindexed

Attacked this specifically; the deployed-production path is safe.

- Injection chain verified in installed sources: CF `fetch(request, env, ctx)` → `nitropack/dist/presets/cloudflare/runtime/_module-handler.mjs:92-105` passes `_platform.cloudflare.env` on **every** Worker-handled request (unconditional — no branch that omits it) → `nitropack/dist/runtime/internal/app.mjs:47-56` spreads `_platform` into `event.context`. So whenever the Worker handles the request at all, `event.context.cloudflare.env` exists and equals the wrangler env.
- CI is the only deploy path (`block-local-deploy` hook) and always names the env: `.github/workflows/ci.yml:395` (`environment: staging`) and `:517` (`environment: production`) → wrangler-action `--env production` → `vars.APP_ENV = "production"`. Confirmed empirically: `wrangler dev --env production` → no header on 200 and 404.
- Prerender/static-HTML path: built the app; `.output/public` contains **zero .html files** (only assets/fonts/images/sw.js, empty `_headers`), so there is no prerendered HTML that could be served with a stale/missing header.

**One real footgun found (P2, non-blocking, pre-existing):** the top-level `wrangler.jsonc` block has `"name": "animichi-web"` — the production Worker's name — but no top-level `vars`. A bare `wrangler deploy` (no `--env`) would overwrite the production Worker without `APP_ENV`, and the fail-safe would silently deindex prod. Reproduced: bare `wrangler dev` (no `--env`) serves `x-robots-tag: noindex, nofollow` on `/`. Today this is fenced by CI-only deploys + the local-deploy hook, but consider either adding `"vars": { "APP_ENV": "production" }` at top level (consistent with the name it targets) or renaming the top-level worker so a bare deploy cannot clobber prod. The zod guard in `tests/unit/wrangler-app-env.test.ts` could then assert the top-level block too.

### 2. The #498-shaped trap (staging silently unprotected, tests green)

The tests do **not** repeat #498's mistake. Two independent legs:
- `noindex-header.test.ts` injects env through the same shape the runtime uses (`event.context.cloudflare.env`), wired via `hookable` + h3 `onBeforeResponse` exactly as `app.mjs:81-84` does (verified against the installed nitropack — same `callHook("beforeResponse", event, response)` wiring).
- `wrangler-app-env.test.ts` asserts on the **actual file** (`?raw` import, always-run pool) per env block, so hardcoding or deleting a `vars` block fails loudly. The iterator guard closes the new-env escape hatch.
- Residual gap the unit tests cannot cover — whether CI passes the right `--env` — I closed by reading `ci.yml` directly (see above).

### 3. Strength of the `wrangler dev` evidence

Reproduced all three runs locally (staging: header on `/` 200 and `/no-such-page` 404; production: absent on both). `wrangler dev` resolves `vars` from the same env-block config parsing as `wrangler deploy`, deploy re-uploads config `vars` each run (no dashboard drift survives), and there is no `.dev.vars` file to shadow them. So the dev evidence is strong for "config → runtime env → header" — what it does **not** prove is that CI passes the right `--env`, which is verified separately in `ci.yml:395/517`.

### 4. Response coverage

- ASSETS bypass confirmed as described: `cloudflare-module.mjs` short-circuits `env.ASSETS.fetch` for public asset URLs; reproduced (`/sw.js` → 200 without header). All HTML is Worker-rendered (no `.html` in `.output/public`), so the crawler-canonical surface is covered.
- **One additional bypass the PR did not list (informational):** h3 1.15.11's error path skips `onBeforeResponse` when Nitro's `errorHandler` sends the response itself — the web-handler catch (`h3/dist/index.mjs` ~2415-2430) calls `onError`, the error handler marks the event handled, and `onBeforeResponse` is never reached (same early-return in the node path at ~2327). Net effect: a 5xx error page on staging likely ships **without** the header. SEO impact is nil (search engines do not index 5xx responses), so no change requested — but the "every response" claim excludes handled-error responses. Redirects/404s from TanStack Start return `Response` values through the normal handler path and are covered (404 verified empirically).

### 5. Mutation reproduction (2 of 8, as spot-check)

- **M1** (`===` → `!==` at `src/server/noindex-plugin.ts:24`): 9 failures including AC2. Killed. Reverted.
- **M4** (header only when `event.path === "/"`): exactly one failure — the AC4 test "404 page gets the header too" — confirming the `createRouter` harness (commit 2) preserves `event.path` where the original `app.use(base)` harness would have masked this mutant. Killed. Reverted; tree verified clean afterward.

### 6. Dependency changes

Justified and low-risk. `h3@^1.15.11` / `hookable@^5.5.3` / `nitropack@^2.13.1` were already in the graph transitively (nitropack resolves to the same 2.13.4 the vite plugin uses; h3 stays on the 1.x line — no dual-major risk); making them explicit devDeps is correct since the plugin and tests import them directly. `pnpm-lock.yaml` delta is importer-entries only (+9 lines), matching the claim; `pnpm install --frozen-lockfile` clean.

### 7. Hard rules + gates (all re-run locally)

- Functions: 2/6/3/2 lines — all ≤10, ≤2 indent. Test files 95 and 42 lines (≤200). **Zero mocks**, no wall-clock assertions, no conditional logic in tests, no suppressions anywhere in the diff.
- `pnpm run typecheck` → exit 0. `pnpm run lint:oxlint` → exit 0. `pnpm test` → 203 files / 1523 tests passed; coverage 98.53 / 95.44 / 98.6 / 99.46 vs floors 98/95/98/99 — matches the PR's numbers exactly.
- Quality Ratchet: AC1-AC5 each map to named assertions in the diff (unit). CI: 18 checks green, 0 failed.

### Summary

- Blocking: none.
- Suggested (P2): top-level wrangler block shares production's worker name without `APP_ENV` — a bare deploy would silently deindex prod (reproduced via bare `wrangler dev`); consider top-level `vars` or a distinct top-level name + extending the zod guard.
- Informational: handled-error (5xx) responses bypass `beforeResponse` and ship without the header; harmless for SEO.

— Fable 席 (independent seat; other seat not consulted)

lifeodyssey:
## Round 2 — both seats found the same footgun independently

`13b3e0c`. Two seats, two different review briefs, converging on one finding is the strongest signal this PR produced — and Fable **reproduced** it rather than reasoning about it: a bare `wrangler dev` served the noindex header.

### The footgun

`wrangler.jsonc`'s top-level `name` is `animichi-web` — the **same Worker** as `env.production.name` — and the top level declared no `vars`. A `wrangler deploy` with no `--env` therefore publishes to production with `APP_ENV` undefined, the fail-safe fires, and the live site silently serves `X-Robots-Tag: noindex`.

Fenced today: CI always passes `--env` (`ci.yml:395` / `:517`) and `block-local-deploy` stops manual deploys. But as Opus put it, **this PR is what converts a harmless redundant deploy into a harmful one.** The config now states the truth instead of leaning on the fence.

### The guard asserts a relation, not a literal

```
top-level APP_ENV  ===  APP_ENV of whichever env block shares the top-level Worker name
```

Written this way, a rename cannot leave it green while meaning something else, and a top-level `name` matching no env block fails loudly rather than vacuously passing.

Mutations, all killed: delete the top-level `vars`; set it to `staging`; change the top-level `name` so no env block matches.

### Coverage limits, recorded rather than smoothed over

The PR claimed "every response". Two exceptions, both now in the plugin comment:

- **ASSETS-binding responses** never reach the Worker, so static files get no header. All HTML is Worker-rendered, so canonicalisation is covered; static assets are the account layer's job.
- **A *handled* error skips `beforeResponse`** — nitro's errorHandler marks it handled before the hook runs (h3 1.15.11), so a staging 5xx ships bare. Search engines do not index 5xx, so this is documented, not fixed. Fable also found the flip side and the PR was *understating* it there: unhandled errors **do** reach the hook.

### The doc touchpoint (Opus P2-A)

`docs/ops/deployment.md` now records that this is the **second** `APP_ENV` in the repo and that the two have **opposite absence contracts** — the container's is fail-**closed** (throws, #498), this one is fail-**open-to-noindex**. Both directions are deliberate: a mislabelled trace is cheap; a live site that stops sending `noindex` is not, and neither is one that starts. The doc says so explicitly so nobody "unifies" them without deciding which cost they are choosing.

Opus checked the other two of the "three touchpoints" and found them genuinely inapplicable rather than skipped — `CONTAINER_ENV_KEYS` is container-only, and `vars` ship inside the config file. That is the rule being used correctly: understand what it prevents, then decide which touchpoints apply.

### Not done here

**Opus P1-B — no post-deploy assertion that the header is *absent* in production.** The unit suite does assert it (AC2, and the inverted-check mutation dies there), but nothing observes the deployed site. `_post-deploy-test.yml` is entirely TODO no-ops today. That is CI work on a file the CI rebuild replaces, so it belongs there rather than bolted on here — logged, not forgotten.

**The second origin Opus flagged** — `animichi-staging.<account>.workers.dev` — I checked empirically and the config-based reading does not hold: `/` returns **404** (17 bytes), because the root Worker's staging deploy has never succeeded (#537, `.open-next/assets` does not exist). The Worker shell answers `/robots.txt` with 200 but serves no site. So it is not a second indexable origin today. It does carry no `X-Robots-Tag` and is on `workers.dev` against the owner's decision — that belongs in #541.

### Gates

`typecheck` clean · `lint:oxlint --type-aware --deny-warnings` clean · 203 files / **1524** tests pass · coverage 98.53 / 95.44 / 98.6 / 99.46, unchanged and above floors · zero suppressions.

Issue #545, opened 2026-07-29
Retire the legacy site: 301 the old hostnames to animichi.com, then delete the app


lifeodyssey:
Owner decision, 2026-07-29: retire the legacy site, keep the legacy hostnames, and serve **only a 301** from them.

The distinction matters — delete the *app*, keep the *hostname*. A hostname that 404s throws away whatever equity the old URLs have; a hostname that 301s hands it to the new domain.

## Measured state (2026-07-29, verify before acting — this will drift)

| Host | Status | Notes |
|---|---|---|
| `seichijunrei.zhenjia.dev` | **200, live site** | `<title>アニメ聖地巡礼 スポット検索・ルート計画 \| Seichijunrei</title>`; `robots.txt` allows search crawlers (`Content-Signal: search=yes,ai-train=no`), blocks training bots; `sitemap.xml` 200 |
| `seichijunrei.zhenjia.org` | **301 → `.dev`** | |
| `animichi.com` | **no DNS record at all** | apex and `www`, A/AAAA/CNAME all empty (#538) |
| `aninavi.app` | on Cloudflare (`coen.ns.cloudflare.com`, same NS as `animichi.com`) | held, but no records |
| `seichijunrei.app`, `kitsunavi.com` | no NS | not held |

## A defect in the live site worth knowing about

`.dev` serves `<link rel="canonical" href="https://seichijunrei.zhenjia.org/">` and the matching `og:url` — and `.org` 301s back to `.dev`.

So the canonical points at a URL that redirects to the page declaring it. Google treats that as a conflicting signal and generally **ignores the canonical and picks a URL itself**. Practical consequence for this migration: whichever of the two is currently indexed is not something the site chose, so do not assume `.org` is the indexed URL just because the canonical says so. Check Search Console before deciding which host the 301 rules must cover — the answer is probably "both", which is what this issue specifies anyway.

## Ordering — this cannot go first

**A 301 to a domain with no DNS is worse than the current state**: every old URL would start failing to connect instead of serving a page. The sequence is:

1. #544 — `apps/web` gets the SEO layer *(in review)*
2. #541 — DNS + routes, `animichi.com` actually answers
3. #537 / `feat/537-retire-legacy-frontend` — `frontend/` deleted, root Worker reduced *(in progress; independent of 1 and 2, and it is what makes the root Worker deployable at all)*
4. **This issue** — 301 rules, then delete the old app

Steps 1–3 can overlap. Step 4 must be last.

[...]

PR #544, opened 2026-07-29
feat(web): port the SEO layer onto apps/web at the animichi.com apex (#252)

lifeodyssey:
Closes the porting half of #252 — the precondition for retiring `frontend/` from the request path (#537).

## What was actually missing (independently re-verified against the tree, not taken from the card)

The card's gap list said `apps/web` had "verification utilities but nothing calling them" for hreflang and JSON-LD. That is **wrong**, and I confirmed it before writing code:

- `src/features/seo/json-ld.ts` is already the emission path (router `head().scripts`, `<` escaped) and `/anime/$bangumiId` already uses it.
- `src/features/anime/head.ts` already emits real hreflang alternates per work.

So the actual gap is **site-level and home-level**, not per-page. This PR reuses both existing mechanisms rather than standing up a parallel stack: `homeAlternates()` returns the same `HreflangLink` shape `features/anime/head.ts` defines, and the home JSON-LD goes through `jsonLdScripts()`. **No `dangerouslySetInnerHTML`** (that is what `frontend/app/layout.tsx` uses; it was deliberately not copied).

## Origin choice and why

`CANONICAL_ORIGIN = "https://animichi.com"`, a single constant in `src/features/seo/site.ts`, **not** a `VITE_*` read.

- `VITE_*` values are inlined at build time and none are injected in this app's builds (#506). An env-driven origin would silently resolve to `""` and ship relative canonicals — a failure mode with no loud symptom.
- `public/robots.txt` and `public/sitemap.xml` are copied verbatim by Nitro and cannot be templated at all, so the origin has to exist as a literal *somewhere*. One literal, asserted by a test to be the only one in `src/`, makes a future domain change a one-line edit — which is what the `CANONICAL_DOMAIN` AC is actually asking for.
- Staging/preview serving the production canonical is correct, not a bug: those hosts are behind Cloudflare Access and are getting `X-Robots-Tag: noindex` on a sibling card, and a canonical pointing at the real site is what we want any leaked crawl to consolidate onto.

`animichi.com` still has no DNS record (#538) — this PR ships the markup that becomes correct the moment the record exists, and is inert until then.

## Where the tests live and why

Option (a): vitest inside `apps/web`. The Python `apps/agent/agent/tests/unit/test_seo_static_files.py` is **deleted**, not repointed.

[...]

PR #548, opened 2026-07-29
feat(#537): retire the legacy frontend/ and reduce the root Worker to /v1 + image proxy

### LINKED ISSUE #537: P0: root Worker CI deploy is structurally impossible — the build never produces .open-next, and wrangler.toml points one directory too high

lifeodyssey:
Retires the frozen `frontend/` Next.js package and reduces the root Worker to `/v1` + the image proxy. Implements the topology decision in #537 (owner picked option B, 2026-07-29).

## This is what unblocks #537, not merely cleanup

The root Worker has **never been deployed successfully by CI**. The reason: `ci.yml` passes `build_filter: frontend`, which runs `next build` and emits `frontend/.next`, while `wrangler.toml` declares `[assets] directory = ".open-next/assets"` in three places — a directory **nothing in the repo produces**. `.open-next` only comes from `opennextjs-cloudflare build`, which CI never invokes.

Deleting the OpenNext fallthrough removes both the config and the build step. The root Worker becomes deployable for the first time.

## Nothing live is lost

- `animichi.com` has **no DNS record at all** (#538) — the apex has never served anything
- the root Worker's staging deploy has never succeeded, so `frontend/` has never rendered from it either
- the still-live legacy site at `seichijunrei.zhenjia.dev` is a **separate, orphaned deployment** — its title is still `Seichijunrei`, from before the rename, so the repo cannot change it in any case. Owner's decision: treat it as dead; 301 and teardown are #545

## Durable Objects — checked before touching

`worker/entry.ts` re-exported `DOQueueHandler` and `DOShardedTagCache` from the OpenNext bundle. Deleting a Durable Object class is one of the few genuinely irreversible operations on Cloudflare, so this was verified rather than assumed:

Repo-wide grep finds those two names in exactly two places — `worker/entry.ts:6` and one June planning doc. `wrangler.toml` declares `durable_objects.bindings` only for `CONTAINER`→`RuntimeContainer` and `EDGE_GUARD`→`EdgeGuard`, and `migrations` only `new_sqlite_classes = ["RuntimeContainer"]` / `["EdgeGuard"]`.

**They are dead exports. No `deleted_classes` migration is needed and no DO state can be lost.** `RuntimeContainer` and `EdgeGuard` are untouched, and a test asserts the former is still exported.

## Unmatched paths now return a JSON 404

```json
{"error":{"code":"not_found","message":"No route matches this request."}}
```

- **404, not a friendly 200.** An "this is an API gateway" body is a soft-404: crawlers index it and clients get nothing to branch on. The path genuinely does not exist on this origin.
- **The same envelope as `unauthorized` / `rate_limited`**, so one client parser covers every edge rejection.
- **Via `app.notFound()`, not a new `app.all("*")`.** The existing `/catalog/public/*` explicit deny already calls `c.notFound()`, so both deny paths now share one shape and one code path instead of two.

## Two tests were stale, and they were not the same problem

Both lived in `worker/entry.test.ts` and both asserted the deleted fallthrough. They were handled differently on purpose:

**Kept and rewritten** — `"/catalog/* is NOT publicly routed"`. That name was **never true**: `/catalog/public/*` *is* routed (`app.ts`). It came from a comment in `app.ts` that this campaign already found to be false — contradicted seventeen lines below itself. The property worth keeping is narrower: **a real private catalog endpoint must be unreachable from the edge and must never touch the `CATALOG` binding.** `/catalog/search` is such an endpoint and it falls to `app.notFound()`; the sibling case in `gatewayFallback.test.ts` uses an invented `/catalog/public/*` path, which exercises the explicit-deny branch instead. Different code paths, different security weight. Renamed to state what it asserts.

**Deleted** — `"unknown path falls through to OpenNext"`. `gatewayFallback.test.ts` covers it twice (`/` for status, `/some/legacy/page` for the envelope). Confirmed the property survives *before* removing it, with a note in place saying so.

## Mutation verification

Each mutation confirmed to land on the target structure before running, and each read by **failing test name**, not count:

| mutation | result |
|---|---|
| `status: 404` → `200` | 8 fail, incl. `an unmatched path is 404, not a rendered page` |
| `code: "not_found"` → `"nope"` | 3 fail, incl. `the unmatched-path body is the shared JSON error envelope` |
| restore `app.all("*")` | many fail, incl. the whole anonymous branch — the catch-all intercepts ahead of the `/v1` handlers |
| re-add the `.open-next` import | **exactly 1** fail: `entry.ts imports nothing from the retired OpenNext bundle` |

A first pass ran all four back-to-back and reported an identical **17 failures for every one, including the restored baseline**. Identical counts across semantically different mutations was the tell: the host had exhausted its process slots and whole test *files* were failing to load. Those were false kills. Re-run one at a time, reading names instead of counts.

## Two CI configs were already broken, and are fixed rather than deleted

`.github/workflows/dependabot-agent.yml` used `cache-dependency-path: frontend/package-lock.json` and ran `npm ci` inside `frontend/` — but `frontend/` is a pnpm workspace member with **no lockfile of its own**. `.github/dependabot.yml`'s `directory: "/frontend"` npm ecosystem had the same problem and never resolved anything. Both were rewritten onto the root pnpm workspace, so the coverage is restored instead of silently dropped.

## Deliberately left broken, and why

- **`make local-login`.** It rewrote magic links to the legacy login page on `:3001`. Every local URL (Makefile, `supabase/config.toml` `site_url`, `scripts/e2e-setup.sh`, `scripts/local-login.sh`) now points at `:3000`/`apps/web`, but **`apps/web` has no login route until Neon Auth lands (#312)**. Pointing at a port where the app actually runs beats pointing at one where nothing listens; either way local magic-link login is unavailable until #312.
- **ESLint left the repo with `frontend/`** — it was the only package still using it. Pre-commit hooks and the PR template's suppression list are updated. No ESLint dependency audit was run.

Two e2e specs (`middleware-redirect.spec.ts`, `public-pages.spec.ts`) were **already dead** before this PR — `frontend/app` contains only `page.tsx` and `login/page.tsx`, and there is no `frontend/middleware.ts`. They would have failed today. Deleted.

## ⚠️ Ordering: this changes #541, and the change is dangerous if missed

`wrangler.toml`'s root Worker **still claims `routes = ["animichi.com/*"]`**. That is left to #541 deliberately (its stated scope), and warnings are in both `wrangler.toml` and `docs/ops/deployment.md`.

**With the page fallback gone, adding a DNS record before narrowing that route makes the apex answer a JSON 404.** #541 has been updated: narrowing the route now comes **before** DNS, not after. The original order was correct while `frontend/` still rendered; it became harmful with no edit to the list itself.

Separately: **#541 must not run until `apps/web` has the SEO layer from #544**, or the new domain goes live with no sitemap, robots or OG. Compounded by the note below.

## Known gap this opens

`apps/agent/agent/tests/unit/test_seo_static_files.py` was a **real** dependency on `frontend/` — it read `frontend/public/sitemap.xml`, `robots.txt`, `og-image.png`, `frontend/app/layout.tsx` and `frontend/lib/structured-data.ts` (`FRONTEND_DIR = parents[5] / "frontend"`). Deleting `frontend/` breaks `make test`, so the file is deleted with it.

**That was the only SEO-asset assertion in the repo.** #544 rebuilds the equivalent in `apps/web` and is in review; until it merges there is a window with no SEO guard. Flagging rather than hiding it.

The other `*frontend*`-named agent tests (`test_frontend_contracts.py`, `test_conversation_hydration_contract.py`, `test_runtime_journey_contract.py`, …) were each read and **kept** — the name refers to the frontend *contract*, not the package.

## Not swept

`frontend` still appears in `.claude/skills/frontend-tdd/SKILL.md`, `.claude/skills/e2e/SKILL.md`, `.claude/rules/css.md`, `.claude/agents/executor.md`, `.claude/agents/tester.md`, `.github/ISSUE_TEMPLATE/story.md`, `docs/testing-strategy.md`, `TODOS.md`. `docs/superpowers/**` and `docs/archive/**` are left alone on purpose — they are records, not live docs. `neon-test-base.yml:7`'s `feat/frontend-rebuild` is a **branch** name, not the package.

## Gates

`pnpm run test:worker` **221/221** · `pnpm install --frozen-lockfile` passes (it did not before this PR) · pre-commit gate clean · zero suppressions.

Closes #537.

## Summary by CodeRabbit

* **Web App**
  * The browser experience is now served through the new web application, with local development standardized on port 3000.
  * Legacy landing-page and email-login flows are no longer available.

* **Bug Fixes**
  * Unrecognized routes now return a consistent JSON 404 response instead of falling through to an HTML application.

* **Documentation**
  * Repository, deployment, and local setup guidance has been updated to reflect the current web and API architecture.


lifeodyssey:
## The SonarCloud failure is real. Two lines, and they matter more here than the rule usually does.

CI shows **"C Security Rating on New Code (required ≥ A)"**. The dashboard link is not readable from the terminal, but the check run carries two annotations:

```
.github/workflows/dependabot-agent.yml:64   Omitting "--ignore-scripts" allows lifecycle scripts to run during package installation.
.github/workflows/dependabot-agent.yml:158  (same)
```

Both are `pnpm install --frozen-lockfile` without `--ignore-scripts`.

### Why this one is not boilerplate

In most workflows this rule is noise. In **this** workflow it is the point: `dependabot-agent.yml` exists to process **dependency-update PRs** — it installs precisely the package versions nobody has reviewed yet. A compromised release's `postinstall` would execute on a runner holding the repository token. That is the textbook supply-chain entry point, and `--ignore-scripts` is the mitigation written for it.

### The repo already does this correctly elsewhere

```
.github/workflows/_deploy-component.yml:379   # `--ignore-scripts` disables package lifecycle scripts (pre/postinstall)
.github/workflows/_deploy-component.yml:392   pnpm install --ignore-workspace --frozen-lockfile --ignore-scripts
```

Same protection, same repo, with a comment explaining why. It is present on the deploy path and absent on the dependency-update path — which is the inverse of where it matters most.

### It arrived with the fix, not before it

Worth being precise, because it changes how to read the finding. This PR **repaired** `dependabot-agent.yml`: it previously ran `npm ci` inside `frontend/`, a pnpm workspace member with no lockfile of its own, so **it never worked**. There was no lifecycle-script exposure because nothing installed.

Making it work is what put the code path into service for the first time. **The risk shipped alongside the repair**, which is why the "new code" rating caught it and why it is genuinely new rather than pre-existing debt this PR inherited.

### Asked for

Add `--ignore-scripts` to both, and match `_deploy-component.yml`'s comment so the next person sees the reason rather than a flag. If either invocation genuinely needs lifecycle scripts to run, say which and why — that would be a real exception worth recording, not something to leave implicit.

Not blocking on my side beyond this: the rest of the PR verified clean earlier (221/221 worker tests, `--frozen-lockfile` fixed, four mutations killed by named assertions).

lifeodyssey:
## Codex review — four P1s claimed, all four verified independently, none blocks merge

Ran a Codex review seat in a detached read-only checkout of this PR's head. It returned four "definite breakages". I checked each against the tree rather than taking the verdict, because two of them turn out to mean something different from what the report says.

### ① apex `animichi.com/*` still routes here and hard-404s unmatched paths — **real, and already documented in this PR**

Confirmed: `wrangler.toml:53-54` still claims `animichi.com/*`, and the narrowed Worker 404s anything outside `/healthz`, `/img/*`, `/catalog/public/*`, `/v1/*`.

But this PR **says so itself**, at `wrangler.toml:10-11`:

> `routes` below still claims `animichi.com/*`. Narrowing it (and pointing the apex at `apps/web`) is issue #541's DNS + routes work; the apex has no DNS…

So it is a declared deferral, not an oversight, and the apex has no DNS record (#538) — nothing is reachable to break. It remains a hard ordering constraint: **route narrowing must land before DNS**, or the first visitor to the apex gets a JSON 404.

Also verified the reviewer's own caveat: `GET /healthz` survives at `worker/app.ts:374-376`, matching the route `infra/index.ts` declares.

### ② `make local-login` no longer works — **real, but not for the reason given**

The report attributes this to the helper "sending a Supabase OTP", implying this PR changed it. It did not: the only change to `scripts/local-login.sh` here is `localhost:3001` → `localhost:3000` and a comment. The Supabase OTP flow is pre-existing.

The finding is still correct in substance, via a different path: `apps/web` contains **zero** Supabase references (`grep -rli supabase apps/web/src/` returns nothing) and only Neon Auth (`lib/auth/neonAuth.ts`). So the script now rewrites a Supabase magic link to point at an app that cannot consume it. Deleting the app that *could* is what broke it.

Developer-workflow only, no production surface. Filing separately rather than fixing here — the real fix belongs to the SD-31 auth migration, not to a frontend retirement.

### ③ `scripts/e2e-setup.sh` advertises the `make e2e-public` target this PR deletes — **real, fixed in 28dc68a5**

Confirmed the target exists on the base commit, so this PR introduced the dangling reference. Removed the line. Also swept the other targets this PR deletes (`fe-lint`, `fe-typecheck`, `fe-test`, `fe-build`, `fe-check`, `check-all`) — no live references remain; the only hit is a dated May spec that records what was true then and should not be rewritten.

### ④ `robots.txt` / `sitemap.xml` / `og-image.png` deleted with no replacement — **real, but already solved in #544**

Confirmed all three are deleted here and that `apps/web/public/` currently holds only `fonts`, `images`, `sw.js`.

`gh pr diff 544 --name-only` shows #544 adds **all three** to `apps/web/public/`, plus the `test_seo_static_files.py` coverage this PR removes. So this is a merge-ordering dependency, not a gap: **merge #544 too, or accept a window with no `robots.txt`** — a window that is harmless while the apex has no DNS, and must not survive past the DNS cutover.

### Verdict

None of the four blocks merging *this* PR, because every one of them lands on a domain that has never had a DNS record. Two are genuine prerequisites for the launch sequence, and the order is now explicit:

**#548 (route narrowing) → #544 (SEO assets on apps/web) → #541 (DNS + apex route) → WAF gate (#559) → flip prod anonymous**

Non-blocking items from the same review, not actioned: stale topology descriptions in `docs/ops/deployment.md` and `docs/ops/cloudflare-hardening.md`, and obsolete `NEXT_PUBLIC_*` / Mapbox / Supabase build inputs still passed in `_deploy-component.yml` and `ci.yml`.

**What the review could not check**, stated so its coverage can be judged: it did not run tests, install dependencies, use `gh`, reach the network, or inspect live Cloudflare/Neon/DNS state. The 225 passing Worker tests were supplied to it, not re-derived. Its conclusions about #544 and about base-commit attribution are mine, not its — it had no network and could not see either.

PR #571, opened 2026-07-30
infra(#541): apex/staging topology via Custom Domains, with Pulumi unit tests

### LINKED ISSUE #538: P0: animichi.com has no DNS record at all — the wrangler route can never receive traffic

lifeodyssey:
`animichi.com` has no DNS record, and `wrangler.toml` claimed `animichi.com/*`
for a root Worker that — since #548 — answers JSON 404 to anything outside
`/v1/*`, `/img/*`, `/healthz`. Adding DNS to that would have greeted the first
visitor with JSON. This lays the topology that makes DNS safe, still switched
off.

Per stack, behind the existing `webRoutesEnabled` (default false):

  <apex>            WorkersCustomDomain -> web Worker   (origin for pages)
  <apex>/v1/*       WorkersRoute        -> edge Worker
  <apex>/img/*      WorkersRoute        -> edge Worker
  <apex>/healthz    WorkersRoute        -> edge Worker
  www.animichi.com  proxied A 192.0.2.0 + 301 ruleset -> apex   (prod only)

Custom Domain rather than a route because Cloudflare recommends it when the
Worker has no origin app server, and it provisions the DNS record and an
Advanced Certificate itself. Routes execute ahead of it, which is what lets one
hostname split API from pages. DNS and route narrowing share one flag on
purpose — separating them *is* the JSON-404 bug.

## Codex wrote the first version; two things changed in review

**Staging had no edge routes.** The brief (mine) described staging as a Custom
Domain and nothing else, and that is what got built. But `apps/web` calls
`/v1/chat`, `/v1/photo-search`, `/v1/conversations/...` on its own origin, so a
staging hostname pointed wholly at the web Worker has no chat API at all — and
nothing downstream would have noticed. Both stacks now get the same shape.

`edgeScript` was also hardcoded to `"animichi"`; staging needs
`animichi-staging`. Verified both names against `wrangler.toml` and
`apps/web/wrangler.jsonc` rather than assuming the suffix convention held.

## The tests are the point of this commit

`infra/` had no runner. It has one now — `node --test`, zero new dependencies,
the same idiom as `worker/`. `pulumi.runtime.setMocks` intercepts every
resource construction, so nothing touches Cloudflare.

14 tests over three files (one per config permutation, because `index.ts`
builds at import time and a process loads it once). They assert what `tsc`
cannot: which script each hostname resolves to, the exact route patterns, that
the placeholder is `192.0.2.0` **and proxied**, that the redirect points www at
the apex and not the reverse, that no apex `DnsRecord` competes with the
Cloudflare-owned one, and that flags-off builds nothing but the R2 bucket.

Mutation-verified, including three results worth recording:

- Inverting the redirect target, and un-proxying the placeholder, each fail one
  named test. Both are silent-wrong-behaviour bugs invisible until DNS is live.
- The "gate rule is secret" test survives removing `pulumi.secret(...)` alone,
  and survives downgrading `requireSecret` to `require` alone — each is
  sufficient by itself, since secretness is viral through `interpolate`. Only
  removing **both** fails it. That is the right sensitivity for a
  defence-in-depth invariant, and the test comment now says so; my first draft
  claimed it caught the single change, which was false.
- One mutation was a **false kill**: replacing `pulumi.secret(` with `(` left a
  trailing comma and failed to parse, so the "kill" was the parser, not an
  assertion. Re-run as `pulumi.output(` it survived — which is how the finding
  above surfaced at all.

Secretness is detected with Pulumi's own `runtime.isRpcSecret` /
`unwrapRpcSecret`. The first version inlined the two sentinel hashes as string
literals; gitleaks blocked the commit, correctly — a bare 32-hex constant is
indistinguishable from a key, and the literals duplicated protocol knowledge
upstream already exports.

Two tsconfigs, deliberately: `index.ts` stays CommonJS because that is how
Pulumi's nodejs runtime loads it, and the tests are ESM with top-level await.
Widening the program's config to fit the tests would have changed how the real
deploy loads the program in order to make a test pass.

`node --test` still warns MODULE_TYPELESS_PACKAGE_JSON. Adding
`"type": "module"` to `infra/package.json` would silence it and is exactly the
kind of change that breaks `pulumi up` in a way nothing here can catch, so it
is left alone.

Wrangler's two `animichi.com/*` route blocks are removed — routes belong to
Pulumi.

Stacked on #559 — retarget to `main` once that merges.

Does **not** publish anything: `webRoutesEnabled` stays false. Flipping it is the cutover, and needs `cloudflareZoneId` + `webDomain` + `wwwDomain` (prod) or `stagingDomain` (staging).

Closes #538 and #541 only once that flag is on and DNS is verified live — leaving both open.

## Summary by CodeRabbit

* **New Features**
  * Added Cloudflare Custom Domain support for production and staging web hostnames.
  * Added explicit routing for API, image, and health-check paths.
  * Added a production `www`-to-apex HTTPS redirect.
  * Added a staging access gate with configurable token setup and rotation.

* **Bug Fixes**
  * Invalid stack hostname configurations now fail instead of falling back to staging settings.

* **Tests**
  * Added coverage for disabled, production, staging, and unknown-stack deployment topologies.


lifeodyssey:
## Both remaining bot findings checked and rejected

Recording the reasoning rather than resolving silently, since "no change" and "not looked at" are indistinguishable from outside.

**1. "Non-stack-aware Pulumi resource names."** Pulumi *logical* names are scoped per stack — `animichi-web-domain` on prod and on staging are different resources in different state files and cannot collide. The suffixing convention in `infra/AGENTS.md` applies to **physical** names, and this code already follows it exactly as the pre-existing bucket does:

```ts
new cloudflare.R2Bucket("catalog-media", { name: mediaBucketName })
```

where only `mediaBucketName` carries the stack. The two kinds of name were conflated.

**2. "Infra tests not executable without `type: module`."** They are — `npm test` in `infra/` runs **15 tests green**. Node 24 detects module syntax and handles ESM plus top-level await without the field, emitting `MODULE_TYPELESS_PACKAGE_JSON`. That warning is deliberately left alone: setting `type: module` changes how Pulumi's nodejs runtime loads the program, and nothing available locally can test that consequence.

The third finding on this PR **was** real and is fixed in `d5fcbce9` — the gate instructions listed only `stagingGateEnabled` and the token, so an operator enabling the gate without the cutover flag would have hit a missing-config failure on `cloudflareZoneId` and `stagingDomain`. Those are now repeated in that block, with a note saying the duplication is deliberate.

Issue #517, opened 2026-07-29
Anonymous photo-search is unmetered: quota vars are unconfigurable and vision calls skip the budget breaker


lifeodyssey:
Found during the #515 review (fix for #502). Two independent gaps that together leave the most expensive anonymous path in the product without a ceiling.

## 1. The quota knobs do not exist anywhere they could be set

`PHOTO_SEARCH_QUOTA_*` appears in neither `worker/containerEnv.ts`'s `CONTAINER_ENV_KEYS` nor any `[vars]` block in `wrangler.toml`. Whatever default the code carries is the only value production can ever have — the limit is not configurable, in any environment, by any means short of a code change.

This is the same env-chain shape already recorded for `ZETA_API_KEY` and `OPENAI_COMPAT_API_KEY` in `docs/ops/secrets.md`, and for `APP_ENV` in #498. It is becoming the most common defect class in this repository.

## 2. Vision calls are outside the budget breaker

`daily_usage` is not written by the vision path, and the anonymous daily-cost breaker only gates `chat.py`. So photo-search spend is invisible to the mechanism built to cap anonymous spend.

Compounding it, an earlier review found the breaker's pricing constants (`MODEL_INPUT_COST_PER_MTOK_USD` / `MODEL_OUTPUT_COST_PER_MTOK_USD`) also have no injection path, so every turn is costed at $0 and the $5/day ceiling has never been reachable — see the `CONTAINER_ENV_KEYS` audit on #501.

## Why it matters now

photo-search is live and on the anonymous allowlist (#445 / #447), so it is reachable without an account. #515 adds a further path: a BYOK user with a broken key now silently falls back to the **platform** provider, which means a misconfigured key spends our money rather than theirs.

## Scope

Add the quota vars through all touchpoints (wrangler.toml × 3, `CONTAINER_ENV_KEYS`, `docs/ops/deployment.md`), route vision spend into `daily_usage`, and decide whether the breaker should gate the vision path as well as chat. The pricing-constant gap is tracked with #501 and should be fixed first — a breaker that costs everything at zero cannot protect anything regardless of what it gates.

lifeodyssey:
修复已随 PR #670 落 main(R1 移植 rebuild 上的 #585:匿名 vision 计量 + budget breaker;附加修复 budget 403 信封补齐契约必填 message)。staging/production 验证随部署链推进。

PR #577, opened 2026-07-30
fix: REAL review-bot findings from the comment backlog (batch 2)

lifeodyssey:
Fixes for the older half of the triaged backlog — bot comments on merged PRs
#182–#476. Written by Codex, verified and partly reverted here.

Touches `chat.py`, `history_compaction.py`, `selected_route.py`,
`catalog_errors.py`, `ChatPage.tsx`, `RouteCard.tsx`, `SearchResult.tsx`,
`useRecomputeTurn.ts`, `useSpotSelection.tsx`, `tool-steps.ts`,
`supersession.ts`, and their tests.

## One change reverted, and it was the dangerous kind

`packages/contract/openapi.json` was collapsed from **1860 lines to 1** — the
whole document minified onto a single line. Restored from `main`.

That file is a **committed generated artifact**: `packages/contract/AGENTS.md`
lists it under "committed generated wire artifacts", produced by
`pnpm run emit:openapi`. Its correct contents are whatever the generator emits,
so hand-editing it is wrong regardless of the formatting chosen — the next
regeneration would revert it and the diff would confuse whoever ran it.

Worth noting how nearly it passed: the JSON is semantically identical, so every
test stays green and no lint rule objects. Only the `1861 +----` line in
`--stat` gives it away. A reviewer skimming a green build would have merged it.

Second time today Codex has rewritten something the repo explicitly documents as
deliberate — the other was a TOML reader whose comment said it was intentionally
line-based. The pattern is worth naming: it reads the code, forms a reasonable
opinion, and does not weigh the note explaining why the current shape was chosen.

Gates after the revert: worker 225 pass · catalog 35 + 262 passed · apps/web
1572 passed · agent unit 1666 passed · oxlint clean.

### DIFF
--- a/apps/agent/agent/agents/agent_result.py
+++ b/apps/agent/agent/agents/agent_result.py
@@ -44,11 +44,16 @@ class ProducedRoute:
     route_ref: RouteRef
 
 
+RouteRejectionStatus: TypeAlias = Literal[
+    "empty", "stale_ref", "pending_sync", "upstream_unavailable", "contract_violation"
+]
+
+
 @dataclass(frozen=True)
 class RejectedRoute:
     """A current route outcome that did not produce a registry entry."""
 
-    status: Literal["empty", "stale_ref", "pending_sync", "upstream_unavailable"]
+    status: RouteRejectionStatus
 
 
 StepProvenance: TypeAlias = (
--- a/apps/agent/agent/agents/error_messages.py
+++ b/apps/agent/agent/agents/error_messages.py
@@ -21,6 +21,7 @@ from agent.clients.catalog_errors import (
 
 InputError = Literal["message_too_long", "non_text_message"]
 _DEFAULT_LOCALE = "en"
+CATALOG_ROUTE_UNAVAILABLE_MESSAGE = "Catalog route unavailable"
 
 _INPUT_MESSAGES: dict[tuple[InputError, str], str] = {
     (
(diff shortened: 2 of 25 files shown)


lifeodyssey:
## Review fixes pushed, each mutation-verified

Three fixes landed since the last review pass. Verifying that a fix *works* is not the same as verifying its test would *notice if it stopped working*, so each was broken deliberately and the suite re-run.

**Chat breaker ordering** — `_budget_rejection` now runs before the request body is read. The test proves this structurally rather than by inspection: it builds a `Request` whose `receive` callable raises `AssertionError("the body must not be read after budget rejection")`, so reading the body at all fails the test regardless of what the handler returns. Removing the moved-up check → **1 failed**.

Worth noting this is the third instance of the same shape today (also `#517`'s vision breaker, and a finding on this PR). "Prepare the input, then decide" is the default way this gets written, and it quietly defeats every cheap-rejection guard — the guard runs, it just runs after the work it exists to prevent.

**Retry input comparison** — replacing the structural compare with `left.input === right.input` → **2 failed**. Key-order and array-structure cases both caught.

**Selected-route contract errors** — covered by the added `test_selected_route_errors.py`.

## One lint failure fixed here, not by Codex

`lint:oxlint --deny-warnings` was failing: the supersession `describe` block had reached 69 lines against the 50-line ceiling. Split by theme — demotion rules stay put, retry-input matching moves to its own block. No test altered or dropped.

Flagging the process gap plainly: the unit suites had been run but the lint gate had not, so this arrived red.

## One thing I did not change

`equalValue` carries `WeakMap`-based cycle detection. Tool inputs arrive as JSON over the wire and cannot contain cycles, so that branch is unreachable for this data source. Left alone — it is tested and correct, and rewriting working code for a dead branch buys nothing. Noting it so the next reader does not assume cycles were an observed problem.

## Gates on the final tree

`apps/web` **1575 passed** (208 files) exit 0 · agent unit **1668 passed** exit 0 · web typecheck exit 0 · web lint:oxlint exit 0.

PR #586, opened 2026-07-30
feat(#526): authenticated session listing for the users Worker

### LINKED ISSUE #526: Migrated sessions beyond the one in the return target are unreachable — no list UI, no listing endpoints
The remaining half of #507, split out of the #514 review so it does not disappear into a closed PR.

#514 closes the headline journey end to end: save wall → login → migration → back to `/chat?session=<id>`. The review traced it through `useConversationHistory` → `conversations.py` → the edge's `app.all("/v1/*")` and confirmed the deep link works.

But a single `UPDATE conversations SET user_id` can move **several** conversations at once. Only the one named in the return target is reachable afterwards. Everything else the visitor built anonymously is now owned by their account and has no route to it:

- no session list UI and no route list UI anywhere in `apps/web`
- `GET /v1/conversations` and `GET /v1/routes` have no implementation

So the migration is correct at the data layer and partially realised at the product layer. A user who had three anonymous conversations gets one of them back.

Related dead surface found in the same sweep — the five `checkin`/`share` contracts have no implementation and no client, yet are emitted into the published `users-openapi.json` (`emit-openapi.ts:57`), so the public API document advertises five endpoints that must 404. Worth handling alongside, or splitting again if that is cleaner.

## Why this needs its own issue rather than a note

#507's ACs were all satisfied while the user could still reach nothing — 25 ACs, 18 unit, 7 integration, zero from the client's side. Recording "the rest is unreachable" only in a merged PR body reproduces exactly that failure mode: the next audit sees a closed issue and green ACs.

lifeodyssey:
Adds `listSessions` to the users Worker: an authenticated, user-scoped, newest-first page over `conversations`.

## Ownership is enforced in SQL, and that is mutation-proven

`userId` comes from `context.userId` — established by the authentication middleware — and never from `input`. A caller cannot ask for someone else's sessions because there is no parameter through which to ask.

The test that matters here is the isolation one, and it deserved scepticism: `fakeDb` **reimplements** the `WHERE user_id` filter in TypeScript. A fake that reimplements the clause under test will happily keep filtering after the real SQL stops doing so, and the test stays green while production leaks.

So it was mutation-checked directly. Dropping `WHERE user_id = ${userId}` from the real query:

- baseline `npm run test:worker` → exit **0**
- with the clause removed → exit **1**

It fails. The fake binds `values[0]` positionally, so removing the clause shifts the parameter list and the fake stops matching — the reimplementation does not paper over the deletion. Recording this because "we have a test for isolation" and "isolation is tested" are different claims, and only the second one was worth making.

## Pagination

`LIMIT input.limit + 1` fetches one row beyond the page purely to decide whether `next_offset` should be a number or `null` — the extra row is sliced off and never returned. Ordering is `updated_at DESC, session_id DESC`; the tiebreak on the primary key is what makes paging stable when two sessions share a timestamp, which would otherwise let a row appear on two consecutive pages or on neither.

`toSession` rejects rows that do not carry a `session_id` and `first_query` of the right type rather than coercing them, keeping the untyped `db.execute` result from widening into the typed contract.

## Gates

`workers/users` **37 passed** (4 files), exit 0 · `packages/contract` exit 0 · `tsc --noEmit` clean in both packages · `lint:oxlint` exit 0.

Contract, OpenAPI document and both mirrors updated together, per the checklist in `packages/contract/README.md`.

Closes #526.


lifeodyssey:
## Correction: the contract gate in the PR body was not actually run

The description claims `packages/contract` exit 0. That command did not execute. I ran `pnpm --filter @animichi/contract test`, and no package in this workspace is named `@animichi/contract` — the real name is `@seichijunrei/contract`. pnpm answered `No projects matched the filters` and exited **0**, which I read as a pass.

Re-run with the package's own command (`npm test` in `packages/contract`):

```
Test Files  5 passed (5)
     Tests  60 passed (60)
exit 0
```

The conclusion holds — but it was not evidence when I wrote it, and a green I did not actually observe is worth correcting rather than quietly re-verifying.

The other numbers in the description stand: `workers/users` 37 passed and the `WHERE user_id` mutation (baseline exit 0, mutated exit 1) were run with the package's own `npm run test:worker` and are real.

Noting the general shape because it bit three times today: a filter or script name that does not match exits **0** with no test output. An exit code of 0 accompanied by no pass count means the command did not run, not that it passed.

Issue #535, opened 2026-07-29
Five smaller findings from the three-week bot audit (P2/P3)


lifeodyssey:
The tail of the three-week bot-comment audit. Each verified against `origin/main`; grouped because they are small and unrelated, and separate issues would cost more attention than they are worth. Split any of them out if it gets picked up on its own.

**P2/P3 — invalid UTF-8 in a BYOK header escapes as a bare 500.** `apps/agent/agent/agents/byok_models.py`'s `_decode()` uses `errors="strict"`, and the dependency at `interfaces/routes/_deps.py:233` catches only `ByokError`. A malformed header byte therefore raises `UnicodeDecodeError` straight past the handler. Reachable by anyone who can send a request. (qodo, PR #477)

**P3 — `deferredSave.ts` doesn't guard the storage calls it makes.** `store()?.setItem/getItem/removeItem` are unwrapped; only the *acquisition* of `localStorage` is in a `try`. Safari private mode and quota-exhausted storage both throw on write, not on access — so the guard is on the wrong operation. Same class as the `byokStorage` fix that wrapped accessor evaluation. (qodo, PR #483)

**P3 — `post-deploy-assert.sh`'s `fetch()` nests three levels** (`for` → `case` → `if`), against the repo's ≤2 rule. (qodo, PR #523)

**P3 — `_write_step_summary` promises best-effort and doesn't deliver it.** `apps/agent/agent/scripts/purge_anon_quota_counts.py` does `Path(...).open("a", …).write(…)` with no `try`/`except` and no `with`. A failure writing the summary takes down the purge run it was summarising. (qodo, PR #513)

**P3 — bare `filterwarnings` with no approval note.** `apps/agent/agent/tests/unit/test_purge_cron_settings.py:27` (also present in `test_settings.py` / `test_cron_settings.py`). The repo forbids unapproved suppression; a warning filter is one, and these carry no rationale or tracking id. (qodo, PR #513)

**Also noted, not worth their own line**: `vision_supply_router.py`'s failure log records only `error_type` — adding `status_code` and provider kind would make #502's ERROR/WARNING split actually actionable; two tests in `test_photo_search_route.py` exceed the 10-line guideline; `docs/iterations/iter5/progress.md:34` has a dangling "see below" pointing at a #447 note that does not exist.

## Audit coverage, stated honestly

Over three weeks the bots produced **655 comments**: qodo 314, coderabbit 175, SonarCloud 164. Signal density differs enormously:

[...]

PR #613, opened 2026-07-31
fix(web): tolerate deferred save storage failures

lifeodyssey:
## Summary
- guard both localStorage acquisition and get/set/remove calls without changing non-destructive reads
- model deferred replay as absent, claimed, or failed so a live intent is sent only after successful removal
- report blocked storage as failed, never silently as none, and keep malformed/expired cleanup best-effort
- add mutation coverage for a throwing localStorage property getter and method-level DOMException failures

## Verification
- 46 focused unit/integration checks passed in final independent review
- apps/web type-aware oxlint and typecheck passed
- earlier single-worker full web suite: 1581/1581 tests, coverage above all ratchets, production build passed
- independent final review: PASS

Refs #535 (sub-item 2/5). The parent issue remains open until this final sub-item is merged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when saving deferred chat actions after login.
  * Prevented duplicate replay attempts by safely claiming saved actions before processing.
  * Restored deferred actions when replay requests fail.
  * Added safer handling for unavailable, expired, malformed, or inaccessible browser storage.

* **Tests**
  * Expanded coverage for storage failures, cleanup behavior, expiration, malformed entries, and failed replay requests.

### DIFF
--- a/apps/web/src/features/chat/save/createOnLogin.ts
+++ b/apps/web/src/features/chat/save/createOnLogin.ts
@@ -1,8 +1,7 @@
 import type { SaveRouteInput } from "@animichi/contract";
 import { saveRouteRequest } from "../../../api/hooks/use-save-route";
 import type { SaveRouteRequest } from "../../../api/hooks/use-save-route";
-import { clearDeferredSave, readDeferredSave, writeDeferredSave } from "./deferredSave";
-import type { DeferredSaveIntent } from "./deferredSave";
+import { takeDeferredSave, writeDeferredSave } from "./deferredSave";
 
 /**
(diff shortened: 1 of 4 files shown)


lifeodyssey:
Addressed the final automated review findings in 6828fae7: malformed/expired entries remain absent when best-effort cleanup is blocked; live claims still fail closed; SSR without a localStorage property is distinct from a blocked browser accessor; test cleanup is linear and restores the descriptor. Validation: 47 focused tests, 1,586 full Web tests, typecheck, type-aware oxlint, diff check, independent P0-P2 review, and targeted mutations for all three boundaries.

PR #614, opened 2026-07-31
fix(web): expose clarify option data states

lifeodyssey:
## Summary
- expose semantic data-state values for clarify candidates: available, selected, unselected, dismissed
- derive DOM state and existing fade styling from one phase-to-state function
- pin production-style keyed remount so a new clarification card resets to available

## Validation
- TDD red: 3 of 7 focused tests failed before the data-state implementation
- focused ClarifyCard tests: 7/7 passed
- TypeScript typecheck passed
- type-aware oxlint with deny-warnings passed
- production Vite/Nitro build passed
- independent P0-P2 review: PASS
- diff check clean; only ClarifyCard and its unit test changed

Refs #446. This resolves only the ClarifyCard data-state subitem; it does not close the broader photo-search issue.

### DIFF
--- a/apps/web/src/features/chat/components/ClarifyCard.tsx
+++ b/apps/web/src/features/chat/components/ClarifyCard.tsx
@@ -17,6 +17,7 @@ type Phase =
   | { readonly kind: "open" }
   | { readonly kind: "chosen"; readonly id: string }
   | { readonly kind: "rephrase" };
+type OptionState = "available" | "selected" | "unselected" | "dismissed";
 
 function reasonOf(part: ChatDataPart): string | undefined {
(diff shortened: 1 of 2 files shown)


lifeodyssey:
Reviewed the Qodo skill insight; no code change is appropriate. In this PR, data-state is the explicit supported DOM contract requested by #446, not hidden component state, so exact attribute assertions are the acceptance test. User-visible behavior remains covered separately through real clicks, disabled buttons, send calls, rephrase copy, and a production-style keyed remount. Qodo reports 0 bugs and 0 rule violations, and its own relevance note recognizes this as a supported contract.

PR #617, opened 2026-08-01
fix(ci): upgrade all setup-uv calls with cache pruning

lifeodyssey:
## Summary

**Replaces #616.**

- upgrade both tracked `astral-sh/setup-uv` call sites to v9.0.0
- verify the official tag SHA `c771a70e6277c0a99b617c7a806ffedaca235ff9`
- explicitly set `prune-cache: "true"` at every call site
- add a config-as-data test that scans tracked YAML and fails on missing pruning or a mismatched pin

## Verification

- `python3 .github/scripts/test_dependabot_config.py` (3 tests passed)
- `actionlint .github/workflows/dependabot-agent.yml` (passed)
- Ruby YAML parse for both action/workflow files (passed)
- `git diff --check` (passed)
- `make check` attempted; blocked before lint completion because uv could not download `tqdm==4.68.4` through the local proxy (`Operation not permitted`).

## Follow-up

Dependabot dynamic run 30703505153 separately reports `dependency_file_not_found` (`No files found in /`) in the root uv security job. That root-directory issue is intentionally not mixed into this replacement. Keep #616 open until this PR is reviewed and merged; close it only after linking this replacement and its evidence.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -17,4 +17,6 @@ runs:
     - run: pnpm install --frozen-lockfile
       shell: bash
     - if: inputs.python == 'true'
-      uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
+      uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+      with:
+        prune-cache: "true"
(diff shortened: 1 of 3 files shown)


lifeodyssey:
Review disposition: executable CI and security checks for commit `6db0169d1d8f2fafc21d610f0ea24e25d61d9e11` completed successfully; review submissions and inline review threads are empty. CodeRabbit/Qodo comments are top-level processing/status comments only, with no actionable thread. They are recorded as non-blocking informational comments under the current ruleset. `make check` remains environment-blocked by the documented uv download failure, not by this diff.

PR #882, opened 2026-08-07
fix(catalog): qualify ingest_jobs columns in ON CONFLICT WHERE (ambiguous status)

lifeodyssey:
Fixes the live Catalog spikes (Neon) lane failure (PostgreSQL 42702 'column reference status is ambiguous') blocking #869 and #874.

Root cause: in `INSERT ... ON CONFLICT DO UPDATE ... WHERE`, unqualified `status`/negative_cached_until resolve against both the target table and the VALUES row once the newer Postgres counts the VALUES relation as a name scope. The staging DB's engine version exposes it; the query is on main (not introduced by either blocked PR).

Fix: qualify with `ingest_jobs.` prefix. 44 worker test files / 317 tests pass locally (spikes need the live DB).


lifeodyssey:
findings triaged: hotfix PR — CodeRabbit/Qodo findings were adjudicated pre-merge (see PR review threads and the spike harness rework in #884); closing the ACK gap retroactively.

PR #929, opened 2026-08-09
fix(ci): neon-secrets adopt — authenticate the neon provider at import time

lifeodyssey:
## Problem

The deploy-neon-secrets-staging job fails on its first run (P5 wiring, #928): `pulumi import` of the first neon role dies with `authorization key must be provided`. The workflow comment claimed no NEON_API_KEY was needed because the key "rides inside the committed Pulumi.staging.yaml" — but **`pulumi import` never executes the stack program**, so the explicit `new neon.Provider("neon", {apiKey: config.requireSecret("neonApiKey")})` in index.ts is never constructed. The import-time provider only sees stack config in the `neon:` namespace (unset) and the `NEON_API_KEY` env var (unset in this job) → config-level auth failure.

Verified against the failing run [31310968618](https://github.com/lifeodyssey/animichi/actions/runs/31310968618).

## Fix

**`.github/scripts/neon-secrets-adopt.sh`** — import via `pulumi import --file` instead of CLI args:

1. **Neon roles**: the files `nameTable` + `providerInputs` bind the roles to the programs *named* provider URN (`urn:pulumi:<stack>::animichi-neon-secrets::pulumi:providers:neon::neon`), injecting the apiKey (secret-marked, from the `NEON_API_KEY` env var) plus the bridged providers name/version/parameterization read from the generated `sdks/neon/package.json`. This does two things the old script couldnt:
   - authenticates the import (the committed config secret cant);
   - binds the imported roles to the exact provider URN the next `pulumi up` constructs — otherwise the up would REPLACE them (create-before-delete → Neon rejects the duplicate role create → first CI deploy hard-fails, the exact failure adoption exists to prevent).
2. **Secrets Store secrets**: pin each entrys `version` to the `@pulumi/cloudflare` version the lockfile resolves (6.19.0), so they bind to the same default provider (`default_6_19_0`) the up builds; fixed their import IDs to the providers actual format `<accountId>/<storeId>/<uuid>` (bare UUIDs are rejected).
3. Guard: readable `::error::NEON_API_KEY not set` instead of a Pulumi stack trace; skip early (exit 0) once adoption is complete, keeping the idempotency contract.

**`.github/workflows/reusable-deploy-neon-secrets.yml`** — declare `NEON_API_KEY` (required: false) and export it in the deploy step env; corrected the header comment that claimed no key was needed.

**`.github/workflows/ci.yml`** — pass `NEON_API_KEY` through to the reusable workflow.

## Local validation

- `bash -n` + repo pre-commit hooks (shellcheck, actionlint, yaml) pass.
- File-backend simulation (Pulumi 3.247.0, project copy + generated SDK + lockfile): reproduced the exact CI error (`authorization key must be provided`), then confirmed the new --file flow passes provider config validation and plugin resolution — with a fake key the only remaining error is the expected network 401; with the real NEON_API_KEY in CI the import succeeds. Also confirmed the cloudflare ID-format fix (bare UUID → `invalid ID` → full ID reaches the correct API endpoint).

## Notes

- On the Pulumi version CI installs (3.253.0), the *CLI-arg* `--provider` route cannot load the parameterized (bridged) neon provider at all (`pulumi/pulumi-neon` 404) — only the `--file` `providerInputs` route carries the `__internal` name/version/parameterization the engine needs. This is the only path that both authenticates and binds to the programs provider URN.
- After adoption completes, `NEON_API_KEY` is unused (the up authenticates via the committed passphrase-encrypted config) — the script exits 0 before touching it.

## Summary by Sourcery

Ensure Neon secrets adoption imports authenticate correctly and bind to the same providers as subsequent Pulumi updates, and wire NEON_API_KEY through CI for first-run adoption.

Bug Fixes:
- Fix Neon role adoption imports failing due to missing NEON_API_KEY authentication in the Pulumi import-time provider.
- Correct Cloudflare Secrets Store secret import IDs and provider version binding so imports succeed and remain aligned with the default provider.

Enhancements:
- Generate Pulumi import files that bind Neon and Cloudflare resources to the exact provider URNs and versions the stack program constructs, ensuring idempotent first-run adoption.
- Improve the neon-secrets adoption script with stack auto-detection, required SDK presence checks, and clearer error messaging when NEON_API_KEY or dependencies are missing.

CI:
- Propagate NEON_API_KEY through reusable deploy-neon-secrets and ci workflows and document its first-run-only usage for adoption imports.

### DIFF
--- a/.github/scripts/neon-secrets-adopt.sh
+++ b/.github/scripts/neon-secrets-adopt.sh
@@ -15,16 +15,46 @@
 # state are skipped. Safe to run on every deploy; it exits 0 immediately once
 # the stack owns its neon roles (i.e. adoption is complete).
 #
-# URNs/IDs: roles' import IDs are `<projectId>/<branchId>/<roleName>`
-# (projectId/branchId are read from the committed stack config, never
-# hardcoded). The Secrets Store secrets' import IDs are the store-item UUIDs
-# recorded in the #926 file-backend stack export (staging.json, resource
-# `cloudflare:index/secretsStoreSecret:SecretsStoreSecret`) — they are stable
-# store item identifiers, not secrets themselves; if the store items are ever
-# deleted and recreated, these IDs must be refreshed from a `pulumi stack
-# export` of a working stack (or from `pulumi import`'s error output).
+# Why --file imports with an explicit provider (not `pulumi import` args):
+#   - `pulumi import` never runs the stack program, so the program's
+#     `new neon.Provider("neon", {apiKey: config.requireSecret("neonApiKey")})`
+#     is never constructed. The committed neonApiKey config secret therefore
+#     does NOT authenticate an import; the neon provider must be given its key
+#     explicitly. We inject it via the import file's `providerInputs` map,
+#     fed from the NEON_API_KEY env var (the same value the #926 run used).
+#   - The imported resources must bind to the EXACT provider URNs the program
+#     will construct on the next `up`, or the `up` would REPLACE them
+#     (create-before-delete -> Neon rejects the duplicate role create -> the
+#     first CI deploy hard-fails, the exact failure adoption exists to avoid).
+#     Neon roles bind to the program's NAMED provider
+#     (urn:pulumi:<stack>::animichi-neon-secrets::pulumi:providers:neon::neon)
+#     via the file's nameTable + providerInputs (which also carries the
+#     bridged provider's name/version/parameterization from the generated SDK
+#     so the engine can load it). The Secrets Store secrets bind to the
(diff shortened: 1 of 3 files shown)


lifeodyssey:
线程判定: 合并闸检查完毕,无未解决 review 线程(0 inline comments / 0 review threads);sourcery 为 COMMENTED 非阻塞建议(将内嵌 Python 提取到独立文件、为 pnpm-lock 版本解析加校验),已三审无需改动;coderabbit 因 rate-limit 未产生 findings;codecov patch 通过(改动行全部覆盖);SonarCloud Quality Gate 通过(2 new issues 属非阻塞提示,已确认不构成 Bugs/Rule violations);CodeQL 通过。合并判定:GO。

